PR #85, opened 2014-12-26
Added BD9G341EFJ Buck Converter

crackerizer:


keruseykaryu:
**BD9G341EFJ**: Not all pins endings are aligned to the 100mils grid. The LX pin should be passive not power output to avoid stupid ERC errors.
**KTD1624**: Use upper case letters in footprint's name.

crackerizer:
Fixed.

PR #87, opened 2014-12-30
Add AD82x series of OpAmps

Fat-Zer:


keruseykaryu:
Proposed symbol does not meet the KiCad Library Convention rules. Please fix these issues.

Fat-Zer:
I see only wrong (a bit small) text size in pin numbers and names... I don't want to reduce the pin length to keep the symbol common with other opAmp's, or should I?...
Any other remarks?

Fat-Zer:
I've post some ideas about library structure enhancements to the kicad maillist... 

IMHO the OpAmps looks crappy with 100mil pins (the nice-looking length is 150) and text of 50mils (text of input pins overlaps with power pins). Replacing power pins doesn't do the trick either because it doesn't fits the net.

I'm closing this pull request because I can't.find the ideal schematics for OpAmp... That should be done by somebody more skilled...

PR #89, opened 2015-01-02
A ref19x series of voltage regulators

Fat-Zer:


keruseykaryu:
Proposed symbol does not meet the KiCad Library Convention rules. Please fix these issues first.

Fat-Zer:
Can you point out what rules wasn't met exactly?

keruseykaryu:
Pin lenght, pin descriptions, designator and value filelds. Just read the Convention and check the rules for schematic symbols one be one.

Fat-Zer:
Ok, I can agree, that pins are too long (I've presumed that other components in the library already fits the requirements, so I haven't reduced it). I'll fix it. 
By the "pin descriptions" you mean the upper/lower case stuff right?
But what's wrong with designator/value fields?

Also IMO it's more suitable to place it in the reference.lib... isn't it?

Fat-Zer:
take a look again...
I've fixed some stuff, moved it to references.lib and made it a bit common with other symbols there.

keruseykaryu:
The pin endings of Vin, Sleep, Vout does not lead at 100mil grid.

Fat-Zer:
my bad ... fixed...

PR #90, opened 2015-01-02
Adding SST39SF0x0 flash chips

DavidGriffith:
Adding symbols for three flash chips from SST (Microchip): SST39SF010 (128x8), SST39SF020 (256x8), and SST39SF040 (512x8)


keruseykaryu:
Proposed symbol does not meet the KiCad Library Convention rules. Please fix these issues.

keruseykaryu:
I didn't talk about the documentation fields. Symbols have some glitches to fix.

DavidGriffith:
Oh... can you elaborate on that?  I just stretched an existing footprint.

keruseykaryu:
1. Open the [Kicad Library Convention](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention).
2. Go to the **General Rules for Symbols** chapter
3. Read it sequentially and check your symbol against given assumptions.

Do not look at existing similar symbols because they not meet this assumptions yet. We are trying to fix it but it took a time that we don't have it (a much). However, for the new symbols we require that they already meet these goals before merging.

DavidGriffith:
Would you please be more verbose about what, specifically, I need to do?

keruseykaryu:
Dave, please again read my previous comment.

DavidGriffith:
Okay, I now see that the field text, pins numbers and pin descriptions were not 50mil.  Is that it?

keruseykaryu:
Bravo! Only one issue left to find.

DavidGriffith:
What issue is that?

DavidGriffith:
Origin back to the center.  Was that the last one?

keruseykaryu:
The last edit worsened things even more.

DavidGriffith:
All I did was move the origin.  How could that possibly lengthen the pins?

reportingsjr:
Keruseykaryu, is there a reason to be so short with contributors? I understand that being a maintainer can be frustrating, but this is the same sort of behavior that Dick would present and he alienated a LOT of developers from KiCad for no reason.

keruseykaryu:
Being be very verbose and polite the situations where I had to repeat the same thing will continue. It is not my intention to bring somebody down, only to force someone to **act independently**. That's why I give you a hints not a solutions, and it can be perceived as rude of me.
Please do not compare me to Dick Hollenbeck. If Mr. Hollenbeck would be in my place you would have more trouble. However, obvious attempts to create turmoil and devastation I will block without question, just like him. And that situations already happens.

If anyone has doubts whether I am the right person for this place, please express it directly. I can always say goodbye.

CarlPoirier:
Hi David,

These look great now. Thanks a bunch for your contribution.

PR #92, opened 2015-01-02
PIC16F884 fixed and PIC16F887 added

kajusK:
Adding symbols for PIC16F884 with PIC16F887 as alias.
Also, there was a wrong symbol for pic16f884, as you can see in http://ww1.microchip.com/downloads/en/DeviceDoc/41291D.pdf, the pic16f884 has 40 pins, not 28, therefore the alias is incorrect, it should be pic16f883 to match the reality.


keruseykaryu:
Thank you for fixing this.
I accepted your commit unchanged because the `microchip_pic16mcu.lib` library is treated by me, so the KLC compilance is not important right now.

If you plan to add some other symbols somewhere else please follow the KiCad Library Convention document.

PR #94, opened 2015-01-03
Some new components

mistoll:
I've created some components.
Comment's appreciated.


mistoll:
MCP2551 is a dublicate. Please ignore that one.

keruseykaryu:
First. Please follow the [KiCad Library Convention](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) rules when creating new symbols. This is mandatory.
The `texas.lib` is not the ideal library to put there LM5051 component. I have no idea where to put it at the moment. The `special.lib` could be the right place if there wasn't a plan (at least I have it) to reorganize and finally remove this mishmash.
The `on.lib` is not a good name for library related to ON Semiconductor's products. I think this library should be renamed to `onsemi.lib`. Such name will be more transparent.

mistoll:
Ok, please give me some feedback on what's actually not matching the convention. Here's what I think:

AT90CANXXX
text size: fixed, reference: fixed
Symbol name: I omitted the suffix which specifies the footprint. Any advise?

LM5051
Could not find convention violations.
Moved to special.lib (But a plan sounds better ;-) )

TMP411
Reference: fixed
Symbol name: same as with AT90CAN: one symbol multiple footprints. OK?

NID60
Text size: fixed, changed reference to A (this is a pcb)

NCP45560
Renamed on to onsemi.lib (Maybe there's a even better library?)

keruseykaryu:
> AT90CANXXX
> text size: fixed, reference: fixed
> Symbol name: I omitted the suffix which specifies the footprint. Any advise?

Include suffixes, even if you need to create the twinning symbols. Entire `atmel.lib` library uses them.

> LM5051
> Could not find convention violations.
> Moved to special.lib (But a plan sounds better ;-) )

Look at the current pin alignment and **very strongly** recommended by KLC 100mil grid for pin endings. 
As for the second though. Ok. Put it there for a while (I can not define when I will be able to take care for it).

> TMP411
> Reference: fixed
> Symbol name: same as with AT90CAN: one symbol multiple footprints. OK?

This time I will not complain too much.

> NID60
> Text size: fixed, changed reference to A (this is a pcb)

I think the `U` reference was good and there was no need to change it. 

> NCP45560
> Renamed on to onsemi.lib (Maybe there's a even better library?)

I see no other better place.

mistoll:
Next try.

keruseykaryu:
The AT90CAN128-A and -M are reversed. Symbol with -A suffix should have TQFP housing, and -M suffix the QFN one (See datasheet packaging chapter). Also please restore aliases, of course with correct suffixes.

mistoll:
I removed the aliases, because the rest of atmel.lib doesn't use them. For example ATmega32U4 and ATmega16U4, which is the same part with 32k or 16k Flash.

keruseykaryu:
> I removed the aliases, because the rest of atmel.lib doesn't use them.

What are you talking about? I manage this library five years now and I know what is there.

mistoll:
Sorry, I got something wrong with the software and thus misinterpreted the library structure.

No, that I know that you can add a individual description to an alias, I added a one.

keruseykaryu:
Now I think I can finally merge this pull reqest.
Thanks a bunch

mistoll:
Well, next time I hope to get it right on the first try ;-)

PR #96, opened 2015-01-10
Fixed erroneous pin number (pin 50 on tqfp-44) for pic18F4431

fransschreuder:


keruseykaryu:
Thanks for finding and fixing this.

fransschreuder:
Wow, that's a fast merge! Thanks

PR #98, opened 2015-01-10
library/device.lib: Add symbol for common-anode RGB LEDs.

mgottschlag:
Copied from LED_RCBG, renamed LED_RABG and mirrored (except for the three
LED symbols).


mgottschlag:
Oops, looks like I messed up this one. Ignore it for now.

PR #99, opened 2015-01-10
library/device.lib: Add symbol for common-anode RGB LEDs.

mgottschlag:
Copied from LED_RCBG, renamed LED_RABG and mirrored (except for the three
LED symbols).

(second try, the first was missing the entry in device.dcm)


keruseykaryu:
I think I could merge this request.

PR #100, opened 2015-01-11
add lm1117 to dc-dc.lib

tracing1017:
add lm1117 to dc-dc.lib


keruseykaryu:
The `dc-dc.lib` is designed for switching regulators (buck, boost, etc.). Linear regulators has its own library `regul.lib`. Unfortunately the 1117 regulator already exists in this library.

Rejected. Closed.

PR #103, opened 2015-01-13
New library for motor drivers

kajusK:
I added several symbols for motor drivers.

I'm just wondering to which .pretty repo I should add the footprints - multiwatt 15, two custom ones for STK* and one for SLA7044M (I'm still trying to find out if it is a standardized package or a custom one).

BTW, this concept of separated repos for individual modules confuses me a little, is there any specific reason why the modules were separated in way like that? It just makes adding the new footprints harder, but it's just as my personal opinion.


keruseykaryu:
> I'm just wondering to which .pretty repo I should add the footprints - multiwatt 15

Pentawatt, Multiwatt are descendants of the TO-220 series housings but we have a separate library `Pentawatt.pretty` where we have some TO-220-5 footprints. I think we should create one more library called `Multiwatt.pretty` for this kind of housings.

> BTW, this concept of separated repos for individual modules confuses me a little, is there any specific reason why the modules were separated in way like that?

There is a reason. Imagine the one big footprint repo hosted on GitHub (or any else cloud). The Footprint Editor and Pcbnew list and fetch its content during edit or netlist read (first time, not cached). This takes some amount of time and this time depending on connection speed and lags. Ask yourself: How long you can wait for the data? How fast is your connection? How fast is the other people connection?

johnbeard:
This is true, but only because the .pretty repo is downloaded as one big binary blob from Github by KiCad - you could also pull down individual components one-by-one using the Github API.

The disadvantage to multiple repos is that there is no way to track components that go from one to another, and you have to keep an eye out for new ones if you maintain your own fp table.

kajusK:
I do realize it takes some time to download the footprint from github, but who wants to keep all the modules online only? It's ok if you're just sitting in the office all day, but any network problem or traveling outside the wifi range and you can't work... Or just slower network connection... And if the repos are separated, it should take the same time to download them, am I right?

Usually I want to have all the modules available out there, therefore the fp-lib-table contains all .pretty repos I downloaded from the KiCad github.

The pretty repos is a great concept for custom module libraries which are not in the official kicad repos, there it makes sense to me, easy to add, easy to remove. However, it seems to me much better to keep the copy of the pretty repos offline and just update it time to time (which is not very straightforward if you need to update tens of git repos instead of just one which contains libraries (which you usually keep offline anyway) and modules.

And another disadvantage, it makes submitting changes much more annoying, create several footprints and your github account is full of the forked repos... You want to create a schematic symbol and package for something? Sure, just fork this two repos and create two separated pull requests.

Sorry, but it still doesn't make much sense to me. 

Ok, so if you create a Multiwatt.pretty repo, I'll create a pull request with the multiwatt 15 package. What about the packages for [SLA7044M](http://sigma.octopart.com/22146899/image/Allegro-SLA7044M.jpg) and [STK672-04(8)0-E](http://www.cncroom.com/images/STK672-080.jpg)? Is there any repo for non standardized packages?

dandumit:
On 14/01/2015 10:05 AM, Jakub Kaderka wrote:
>
> much sense to me.
>
> Ok, so if you create a Multiwatt.prett
>
I would like to contribute as well to this library.  For simbol and 
footprint.
How should I do ?  Do I have to submit a change ?

kajusK:
By the way, is there any common space for such discussions/questions?

CarlPoirier:
Hi guys,

We have some instructions on contributing here. https://github.com/KiCad/kicad-library/wiki/How-To-Contribute

As for the repos, I suggest something more generic that can include the non-standardized packages, such as Multi-leaded_Power_Packages.pretty. http://en.wikipedia.org/wiki/Multi-leaded_power_package

keruseykaryu:
> As for the repos, I suggest something more generic that can include the non-standardized packages, such as Multi-leaded_Power_Packages.pretty.

Yes. But we probably need two. One for THT power packages (Pentawatts, Multiwatts, (e)SIP), and one for SMT power packages.

CarlPoirier:
Power_Packages_SMD and Power_Packages_ThroughHole is my suggestion!

keruseykaryu:
So, let's create it and do certain transfers between libraries.

CarlPoirier:
Both libraries have been created so the new footprints can be added.

CarlPoirier:
@kajusK and @dandumit, once we pull your footprints in Power_Packages_SMD.pretty, the appropriate changes to the fp-lib-tables (#105) will be pulled in.

Thanks!

PR #104, opened 2015-01-13
Rework of device.lib

### LINKED ISSUE #97: Swap CP and CP2 symbol in device.lib
capture d ecran 2015-01-10 a 17 55 11
Currently, default symbol for polarized capacitor is the first symbol on the picture.

- This symbol don't have a "plus" sign and can mislead users.
- IEC60617 standard use the last symbol.

If accepted, I can make the change

samdolt:
Remove old style version of CP
Add small version of common symbole
Translate some field to english


michal777:
Resistor and inductor are too small because it looks bad with IC's matched to kicad library convention. I'd prefer 300 mils. Diode and capacitor shouldn't be longer (maybe should be wider?). Before merging new size of R, L, C, D it would be good to use them in some simple schematic containing IC's adjusted to KLC and check the final effect.

CarlPoirier:
What about having the normal components 300mil long and the small versions at 200mil like they are now?

Fat-Zer:
By the word, I was gonna propose enhance footprint filters for resistor/capacitors/inductors but have finished only caps: 
C
```
C?
C_????_*
C_????
SMD*_c
Capacitor*
```

CP
```
CP*
Elko*
TantalC*
C*elec
c_elec*
SMD*_Po
```

samdolt:
Here a preview of small symbol in a real schematics. For comparison purpose, you can see current size of mosfet bottom right.
kicad-small-symbole

samdolt:
I agree with 300 mils long for normal size symbols.

@Fat-Zer : I added your footprints filters. I just changed SMD*_Po to SMD*_Pol.
I found no footprint corresponding to C*elec. Is it in a legacy module?

keruseykaryu:
> Here a preview of small symbol in a real schematics.

The resistors are too small. I think they a bit disappears in this schematic.

Here is an example of [our](https://github.com/keruseykaryu/kicad-kerusey-library) schematic symbols.
radar

samdolt:
@keruseykaryu  :+1:  You're version is more readable.

Maybe you can merge your library? Or share it? So we can review all symbols in device.lib, remove old "French" description, replace some symbols with yours and update the footprint filters to be more accurate.

keruseykaryu:
> Maybe you can merge your library? Or share it?

My KiCad library vision is already shared on GitHub. Back to my previous comment and click "our" word or use this **[link](https://github.com/keruseykaryu/kicad-kerusey-library)**.

michal777:
I don't like the long pins. Many people don't like connecting wires 90 [deg] to pins:
bez nazwy

That's why I prefer shorter pins.

CarlPoirier:
Are we all set on the small versions of symbols at 200mils?

For the big versions, I suggest the ones Kerusey shows, but with 50mil pins, which then totals to 300mils with the body. The KLC rule could be that when there are no pin numbers, then the pin length is 50mils.

michal777:
>For the big versions, I suggest the ones Kerusey shows, but with 50mil pins, which then totals to 300mils with the body
The body of the resistor is 300 mils so 300+50+50=400, not 300.

I think that resistors/inductors 300 mils (200+50+50) and capacitors/diodes 200 mils would look good with IC's and won't waste too much space. 400/300 or 400/200 mils would be also OK (maybe better).

I don't like the idea of two variants - we should try to find the best size that looks good with IC's. If user needs more space he can change A4 to A3 or divide schematic to more sheets.

example - 300 mils resistors, 200 mils capacitors and diodes:
zrzut ekranu z 2015-01-18 02 13 17

keruseykaryu:
> example - 300 mils resistors, 200 mils capacitors and diodes:

This example show us some issues that we should avoid during designing the new look. The body of the resistors should be slightly reduced, to avoid overlapping, like those connected to RC0, RC1. The capacitor "wings" shouldn't be too long otherwise they will look flattened.

samdolt:
A preview with 300 mils standard symbols and 200 mils small symbols
preview

For the footprints filter, it is ok to only add "pretty" modules and to remove old legacy entries?

Fat-Zer:
There are some thoughts that bothering me for last few days:
May be resistors/capacitors and most of other device symbols are normal-sized in their current state but nearly every IC symbol is too small?
Consider that: we have a 50mil text. If someone will actually plot the schematics to paper it will be 1.27mm height, so no one could ever read this... the minimum text height as a matter effect a smallest destination between pins should be about 5mm (nearly the same as 200mil).
The same argument for the resistors which are in current state are about 8x2mm. Actually they even shouldn't contain the text inside them because of the size issue. And from this point of view making them even smaller is just a crazy talk...

I won't insist on all this. I understand that it's unlikely that everyone here suddenly realize that they have done everything wrong before that moment and will work several weeks in a raw to rewrite each and every symbol on the library... but at least we could agree that it does make sense to keep the current generic device symbols on the library as they are now? May be just to move the text outside them.

CarlPoirier:
These look very great to me, @samdolt. I have only two minor questions. Are
the polarized capacitor CP1 "wings", as @keruseykaryu calls them, as thick
as those from C? They seem a little thinner to me. Also, why not use the
same "+" size for CP and CP1?

As for the footprint filters, do you mean the old library format? If so, I
say yes. It has been deprecated for quite a long time now. One who still
wants to use it will use a version from long ago and use the
fp-lib-table.for-legacy which has never been updated.

samdolt:
@CarlPoirier  I fixed your two issues. I'll spend some time this evening to check every symbols in device.lib and to update this pull request.

@Fat-Zer : I don't see the size problem when plotting into an A4 page. Text are quite readable.

CarlPoirier:
@Fat-Zer I'm also for 50mil text. As a matter of comparison, in *PCB Design Optimization Starts in the CAD Library*, they suggest a nominal height of 1.2mm for text on silkscreen. I know a PCB is not the same as a schematic, but it's just to show that this size is perfectly readable with a normal or corrected sight.

@samdolt I'd keep the pins at 50mils. We have to make everything fit with the KLC, else people are going to flag this as an issue. Right now it's 100mils, but we can specify 50mils for such components.

samdolt:
But it's hard to follow KLC for small variant. 
Bad version :
capture d ecran 2015-01-18 a 15 31 10

KLC version:
capture d ecran 2015-01-18 a 19 07 44

Maybe it would be simpler to remove small variant.

CarlPoirier:
I was talking about 50mils pins for the resistors. You a quite right however that not all symbols look good with 50mils. All in all, I think it will be better to simply not restrict the pin length when it comes to non black-box components, as I had suggested in the thread of issue #97. I vote for the symbols you have shown 11 hours ago.

Thanks a bunch for all the time you put into this.

michal777:
Is there any reason for having the same length of capacitors/diodes and resistors/inductors? Capacitor/diode symbols are naturally shorter.
I think it would be better to have one standard set of these symbols and avoid unnecessary inconsistency between schematics created with kicad.

CarlPoirier:
Well, I say either we have a rule for all of them, or we don't have any. We cannot begin to have a rule for every type of components. The more complicated the KLC gets, the harder it is for new contributors to follow.

samdolt:
This library needs a lot of work to comply with KLC.

This is the intermediate changelog:

Renamed:
	- CRYSTAL -> Crystal
	- DIODESCH -> D_Schottky
	- BILED -> Led_x2
	- BREAKER -> Jumper
	- BRIDGE -> Diode Bridge
	- CERAMIC_FILTER -> Ceramic_Filter
	- CODING_SWITCH -> Coded_Switch
	- DARL_N -> Darl_NPN
	- DPST -> Switch_DPST
	- DUAL_SWITCH_INV -> Switch_SPDT_x2
	- DOUBLE_SCHOTTKY -> D_Schottky_x2_Serial
	- DOUBLE_SCH_KCOM -> D_Schottky_x2_KCom
Moved:
	- BNC -> conn.lib/BNC

Removed:
	- 7Segment (Already in display.lib)
	- Connector (Already in conn.lib)
	- DISPLAY (Better to use display.lib variant)
	- DISPLAY_BL
	- DISPLAY-3LINE
	- DOUBLE_LED ( Duplicated content)

REFDES changed:
	- Crystal : X -> Y

Redesigned:
	- Jumper
	- Battery
	- CP
	- CP_Small

nickoe:
@samdolt Why not ust Darlington_NPN? The names is short enough and more concise.  What about adding a Darlington_PNP?

CarlPoirier:
@samdolt, can you keep this branch in sync with the master? I'll look into merging it. Sorry for the delay.

samdolt:
@CarlPoirier : I have merged master.

CarlPoirier:
Upon reviewing this, I'd like you to not delete any display but simply move them to the other library. Also, you can keep only one R_Small and that would be R_Small_2. I'll then merge it.

samdolt:
@CarlPoirier : Thank you four your review. It is better now?

CarlPoirier:
R_Small has been replaced by R_Small2, but the latter is still present!

CarlPoirier:
The KLC has been adapted to allow odd pin lengths for non black-box symbols.

samdolt:
@CarlPoirier : I hope I didn't make any mistakes this time.

CarlPoirier:
It's good to go now. Thanks a lot for your work.

PR #106, opened 2015-01-18
Dip switches added

kajusK:
I've created symbols for DIP switches. Could you please create a repository like Switches.pretty for the footprints?

Thank you


CarlPoirier:
Hi,

I created the repos Buttons_Switches_ThroughHole.pretty and Buttons_Switches_SMD.pretty. In the former, I put some DIP switches we had in Discret.pretty. Please review how they compare with yours; they might not be appropriate anymore now that we have the KLC.

I'll merge the 3D models once I have seen they match their footprint.

Thanks!

kajusK:
Hi,
thank you for the repos, I'll take a look. As there is a KLC with rule that the anchor must be set on pin 1, the footprints won't be placed to the correct position, I'll set the offset or move the 3d model itself to the correct position.

kajusK:
I updated footprint filter to match the footprint names (request waiting in KiCad/Buttons_Switches_ThroughHole.pretty#1). I'll try to update the 3d packages asap.

kajusK:
So here are the brand new 3d models of dip switches!

michal777:
It's probably not a big problem but emission in white plastic is 100% while in most of cases it should be 0% (plastic doesn't emmit light).

kajusK:
Thanks for the advice, I'm still not very comfortable with wings3d...

CarlPoirier:
Hi Jakub,

I'm not convinced about your switch symbols. It looks like you tried to replicated the actual part, whereas you should simply have used the usual SPST (on-off) symbol. Furthermore, I suggest you make one part with many units. Therefore we won't have 12 different switches in the library, but only one switch with 12 units. People use the number of units they need and set the right footprint. As an example, look at the 74xx series or the DG308 switches.

kajusK:
Hi Carl,
I used the Eagle before I discovered KiCad. The Eagle has dip symbols similar to the proposed once, that's one of the personal reasons I made them like that.

For me, it's similar to connectors. Why don't we use one part with multiple units for them? You would just use the number of units you need and set the right footprint... Same problem.

If I can guess, the most common way of using DIP switches is setting a configuration of something (e.g. microstepping for stepper driver, default behavior of MCU firmware...), in that case, one side is usually tied to some voltage level and the second one is connected to e.g. MCU IO port (like in the example below). In this case it makes sense to keep it in single unit as it makes the schematic more easier to read and it's usually simpler to draw. Could you imagine something similar with the 8 separated switches? It would take much more space on the sheet and it could confuse somebody else looking at it.

DIP switch example

Separation into units makes perfect sense for gates and similar parts you connect together in much complicated ways. But for DIP switches I see more disadvantages than advantages...

What do you think?

P.S. It would be great if there was a wiki of things to do for the contributors, I mean something like TODO with packages to fix, parts to add... And maybe some discussion page to talk about similar issues like this one. It could make the kicad library much better and bigger. The quality and size of the library is what matters when deciding which EDA to use.

CarlPoirier:
This is a quick gimp of what it could look like with the SPST symbols as units. I fail to see how it is not convenient. Resembling Eagle libraries is not a criteria. Furthermore, if you were to use a DIP-8 package for many different functions, you would be able to put each switch apart in your schematic.

image

I would like to hear other librarian's opinion on the matter.

As for the TODO list, there is this one! https://github.com/KiCad/kicad-library/wiki/Status-of-the-libraries

michal777:
I like dip switch in one piece more. Probably in most of cases it will be used as one device, not set of independent switches and it should be expressed on schematic.

CarlPoirier:
How about a box with a bunch of SPST symbols in it?

michal777:
Yes, SPST may be better, it's easier to see that they are switches (it may be not clear what the graphic represents).

kajusK:
You're right, the SPST solution is better, so here's the update.

I know about that todo list, but it only shows if the library is or isn't KLC compliant, there's nothing about which parts should be added (there's definitely still plenty of very common stuff not in libraries), nothing about missing 3D packages (table of missing/bad named 3D packages could be simply generated by script...), etc...

And what about some discussion places? There are definitely things that should be easy to find and discuss, do all the discussion in pull request isn't the best solution as it's hard to find something older and commenting old merged request isn't good idea...

keruseykaryu:
> You're right, the SPST solution is better, so here's the update.

IMO this isn't look as well. But this is only my opinion.

> I know about that todo list, but it only shows if the library is or isn't KLC compliant, there's nothing about which parts should be added (there's definitely still plenty of very common stuff not in libraries), 

We are not seers to know what others need. ;-) If I had to draw up a personal list of symbols to be added, it was a very looooong list.

> nothing about missing 3D packages (table of missing/bad named 3D packages could be simply generated by script...), etc...

We can create some other tables, but who will manage them.

> And what about some discussion places?

Join to Library-commiters mailng list.

kajusK:
Hi,
thanks for the answer. Is there anything else which prevents this pull request from being merged?

keruseykaryu:
Michał and Carl lead this request. Patience.

michal777:
They are good and ready to merge.
Sory for delay and thanks for your contribution.

PR #108, opened 2015-01-22
Footprint filter set, one incorrect pad fixed

kajusK:
Hi,
I noticed I forgot to add the footprint filter to recently pushed motor drivers, so here it is. Also, I fixed the missing pin number.

And also, I've got a question, is there any way how to set the correct number of pins (for CvPcb to recognize the correct DIP package) when there are some pins unused and therefore not defined?

Thanks!


nickoe:
For your second question, there is a bug report somewhere on the bug tracker on launchpad trying to address that issue. I don't think anyone is working on it IIRC, if it is not fixed already.

PR #111, opened 2015-01-30
Add support for STM32F030K6

MateuszKrawczuk:
Add STM32F030K6T6


keruseykaryu:
Please follow the [KLC rules](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) when designing new symbols. Your symbol violates a few of these rules. 
Check it, prepare a fix and update your request.

ricardocrudo:
Hello @MateuszKrawczuk I'd like to suggest use the following scripts to check KLC rules: https://github.com/ricardocrudo/kicad-utils

danger89:
@ricardocrudo 
Good work! It will be much easier for us all (especially new users) when these rules can be checked automatically, much more accessible!

MateuszKrawczuk:
I tested stm32.lib using @ricardocrudo  scripts and library without my patch returns a lot of errors. 
Kicad Library Convention doesn`t have any section 3.1 or 3.6 so I'm not sure what does it mean 3.1 and 3.6 in script.I 
3.1. Using a 100mil grid, pin ends and origin must lie on grid nodes (IEC-60617). and 
3.6. Field text uses a common size of 50mils. 

I used this library as reference to add new entry to it, I didn`t know whole library is violating Kicad Library Convention.
In my patch I should change:
LQFP32 to LQFP-32
How should I fix this error for example:
https://gist.github.com/MateuszKrawczuk/3c2f59910b7c85797598

ricardocrudo:
@MateuszKrawczuk the sections numbers references are in the text file (https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt). 
In the first case (3.1) the most part of the violations are related to position y of the components pins. According the rule the pins must lie in a 100mil grid point, i.e.: 100, 200, 300, not 50, 150, 250, ...
If you move your component 50mil up or down you problaly will get a lot of this issues solved.
The second case, just change the text size to 50 instead of 60.

keruseykaryu:
> I tested stm32.lib using @ricardocrudo scripts and library without my patch returns a lot of errors.

You don't need to correct whole library to meet KLC rules if you don't want to. Just make your new symbol(s) KLC compliant.
Of course we as librarians will be very happy if you decide to fully take care of it.

> In my patch I should change:
> LQFP32 to LQFP-32

Use hyphen notation. The best solution is to use a full name of matching footprint from `Housing_QFP.pretty`.

MateuszKrawczuk:
Can you check update version?

keruseykaryu:
The PA7 pin description is out of the component body.
The PA12 pin description overlap the prefilled footprint filed.
Designator and name inside the component body.

MateuszKrawczuk:
I created my own branches and I added some changes:
https://github.com/MateuszKrawczuk/kicad-library/tree/stm32-add-new
https://github.com/MateuszKrawczuk/kicad-library/tree/stm32-rework

PR #112, opened 2015-02-02
Common Microchip PIC16, Intel, Philips, Supertext, Video Libraries Update - Applying KLC Rules. Add LTC2983 and LTC2990.

keruseykaryu:
Updates:
- `microchip_pic16mcu.lib` - Pin endings aligned to 100mil grid. Reference and value text set to 50mil. Pins descriptions text and numbers set to 50mil.
- `intel.lib` - Pin endings aligned to 100mil grid. Reference and value text set to 50mil. Pins descriptions text and numbers set to 50mil. Few aliases has been added.
- `video.lib` - Pin endings aligned to 100mil grid. Reference and value text set to 50mil. Pins descriptions text and numbers set to 50mil. Fix AD9891 pinning issue. Update doc fields.
- `philips.lib` - Pin endings aligned to 100mil grid. Reference and value text set to 50mil. Pins descriptions text and numbers set to 50mil. Fix some pinning minor issues. Hudge doc fields update.
- `supertex.lib` - Pin endings aligned to 100mil grid. Pin length has been shortened to 100mil where applicable. Reference and value text set to 50mil. Pins descriptions text and numbers set to 50mil.
- `interface.lib` - LTC2983 (Temperature Multi-Sensor) and LTC2990 (Temp./Voltage/Current Monitor) symbols has been added.


CarlPoirier:
Hi Kerusey,

Thanks a bunch for this great work.

I glanced over it and I have seen some things I'd like to discuss. I have not checked since when they are present.

1. In the PIC16F library, some pin names are overlapped. This happens for the 677-690 and 882-886 microcontrollers.
2. In the Intel library, some symbols have hidden power pins whereas some don't. Actually this might be worthwhile to discuss what to do with this with regards to the KLC.
3.In the Philips library, there are some 100mil pins having double-digit numbers, whereas in other libraries, 150mil is used. I'd use the latter.

keruseykaryu:
> In the PIC16F library, some pin names are overlapped. This happens for the 677-690 and 882-886 microcontrollers.

In the original symbols size of text were set to 0.040" and therefore they fit. Currently, KLC force usage of 0.050" size so the text is bigger and longer, and another rule implicit that symbol should be symmetrical.
Thus, when adapting these symbols to the rules of KLC I had to go for a compromise that will not unduly oversize these symbols. I read somewhere once that the ATmega16 symbol is so big that it takes half a sheet. :)

Of course, in this case it can be simply repaired by moving the pins VCC and GND to the left.


> In the Intel library, some symbols have hidden power pins whereas some don't. Actually this might be worthwhile to discuss what to do with this with regards to the KLC. 

Hidden pins are retained where they were in the original symbols.

About hidden pins we already talked once. We can not prohibit their use, because then schemes using logic gates for example with visible power pins become horribly unreadable. Therefore, we have to create a rather complicated KLC rule where to apply, and where not to.

> In the Philips library, there are some 100mil pins having double-digit numbers, whereas in other libraries, 150mil is used. I'd use the latter.

Will be fixed.

keruseykaryu:
Previous comment has been updated.

CarlPoirier:
Hi Kerusey,

This is awesome. Thanks for your valuable contribution!

keruseykaryu:
Thanks Carl. The Wiki table has been updated.

PR #113, opened 2015-02-02
adding AP1117 regulator family

ricardocrudo:


keruseykaryu:
This description is too long:
> AP1117 is a low dropout positive adjustable or fixed-mode regulator with 1A output current capability.
It can be shorter and succinctly:
`AP1117, Fixed/Adjustable 1A Low Dropout regulator, Positive`

keruseykaryu:
The `AP1117` is a generic symbol or this is the adjustable variant?
If this is an adjustable regulator it could be renamed `AP1117ADJ` for better clarification and GND pin should be called ADJ.

ricardocrudo:
It's generic, the same part can be both, as datasheet says: "adjustable or fixed". The datasheet names the pin as "Adj (GND)", I'll rename it.

keruseykaryu:
I do not like this simple change in pin name. It causes also that fixed regulators (the other aliases of generic symbol) has the ADJ pin but should have GND only.

I think, however, you should do this to get rid of this confusion:
- Rename the symbol `AP1117` to `AP1117D15` and call this lead `GND`, because this is its function in these models. Leave the remaining aliases in this symbol to other fixed regulators.
- Clone `AP1117D15` symbol to `AP1117` and rename this lead `ADJ/GND`. Remove all aliases to the fixed version in this symbol (LibEditor will propose that anyway).

ricardocrudo:
Merge it as is, or close the pull request.

keruseykaryu:
> Merge it as is, or close the pull request.

In this case, I'll wait what other lib maintainers says.

michal777:
Adjustable has reference 1.25 V (to achieve wide output voltage range) and all other are fixed but can be used as adjustable (sometimes reasonable - may improve accurancy) so ADJ/GND is correct but I'm afraid that font size 0.25 will be nearly invisible on paper. I'm for merging.

keruseykaryu:
So. Two votes for merging as is, one vote for split.

CarlPoirier:
As Michal pointed out, the pin name is too small. It must be 0.5mil according to the KLC. I also think the symbol is too big for such a simple part. I suggest splitting and reducing the symbol size, such as this:
image

CarlPoirier:
I'm however ready to accept this:
image

ricardocrudo:
ok. I resized the pins, the body and text names.

keruseykaryu:
Since Carl accept such implementation there is nothing else like merge it.
Thank you for your contribution.

PR #117, opened 2015-02-12
Renamed 3D models of through hole capacitors to match the module files.

nealholl:
This should match the changes I've made in "Capacitors_ThroughHole.pretty"


CarlPoirier:
Will be merged after https://github.com/KiCad/Capacitors_ThroughHole.pretty/pull/4.

PR #119, opened 2015-02-16
FTDI and ADC-DAC libraries update to meet KLC rules

keruseykaryu:
Both libraries has been adjusted to meet the KLC rules.
ADC-DAC library has been recreated and expanded.


CarlPoirier:
Great work! Thanks.

PR #120, opened 2015-02-17
atmel.lib: Pin PC5 uses PCINT13.

arantius:
Previously the library labeled two pins incorrectly as both PCINT14.  This change just switches the incorrect PCINT14 values to the correct PCINT13.


keruseykaryu:
Good catch!
Your fixes has been committed.

Thanks a bunch!

PR #121, opened 2015-02-19
transistors: permutations and move to transistors

michal777:
Transistors (MOSFET, JFET, BJT) with different pins numbering in transistors.lib (no longer needed in device.lib).
MOSFETs and BJTs with two collectors (one is mounting plane).
What do you think about it?
There were discussions about mounting plane (separate pin 4 or the same as middle pin) and about rules 3.1 and 3.2. I'm not sure what are conclusions. If you allow to break 3.1 I'll make base/gate pins shorter.


keruseykaryu:
IMO, the general symbols should remain in `device.lib`. The `transistors.lib` library should contain specific types of transistors with correct pinning and correct housings.

michal777:
Ok but I'm afraid that it will make device.lib too big (only BJTs are 12 symbols, there are not IGBTs, diodes, bridges added yet). What about creating new library for semiconductors with different pins numbering?

keruseykaryu:
> What about creating new library for semiconductors with different pins numbering?
How, for example, this library would be called?

michal777:
Semiconductors.lib? I don't know. It may be harder to find name for library without semiconductors. Maybe divide devices.lib into active and passive components? It's no problem for me to put these transistors into device.lib but maybe it will be necessary to divide device.lib into smaller if it grows to much.

michal777:
I've added more commits related to my other commits: IGBT symbols and many changes in 3d models: new to-92, new resistor arrays, moved, renamed, improved, added 3d symbols of some smd packages.

keruseykaryu:
> Maybe divide devices.lib into active and passive components?

I do not support this division idea, but names: `device-active.lib` and `devices-passive.lib` might be applicable.

michal777:
Ok I will move these symbols to devices.lib.
This pull request contains changes in 3d models that depends of other.
References and comments:
KiCad/Transistors_TO-247.pretty#2 - merged, no changes in 3d
KiCad/Transistors_TO-220.pretty#2 - no changes in 3d
KiCad/Housings_SOT-23_SOT-143_TSOT-6.pretty#2 - merged, changes in 3d (this pull request)
KiCad/Transistors_SMD.pretty#5 - no changes in 3d
KiCad/Housings_TO-50.pretty#1 - changes in 3d (this pull request)
KiCad/Housings_TO-92.pretty#1 - changes in 3d (this pull request)
KiCad/Resistors_SMD.pretty#6 - changes in 3d (this pull request)

keruseykaryu:
I'm, we are not the only one librarian/s here, so maybe we shall wait what Carl would said.

CarlPoirier:
First, for the pin length, I'm leaning towards this, as suggested in #97 :

3.2. For black-box symbols, pins have a length of 100mil. Large pin numbers can be accomodated by incrementing the width in steps of 50mil.

CarlPoirier:
The size of the library is not that much relevant, at least not like the pretties because they are not fetched from the Interwebs on the fly. I say for now you put everything in devices.lib. We'll see later down the road if it needs splitting.

CarlPoirier:
I looked at your transistors and I like them very much. The 3d models matching the other pull requests are very great as well! Thank you very much, Michal.

PR #125, opened 2015-02-21
Mention library-repos-install scripts in README

cpavlina:
I thought it'd be a good idea to mention how to download the PCB footprint libraries locally in the README here, since we point them to the repos themselves as well.


CarlPoirier:
Sounds good to me. Thanks!

PR #130, opened 2015-02-24
adding new symbols

ricardocrudo:
adding MMBT3904/MMBT3906 transistors
adding switch SP3T
adding RJ45 with status leds


keruseykaryu:
`transistors.lib`: Your MMBT390x symbols should have footprint filters with an asterisk at the end. This way we can show also SOT23 housing in other variants (ex. SOT23_Handsolder, SOT23_Wavesoldering).

`device.lib': SP3T, pin 3 has a 60mil text height. Use height recommended by KLC. Pins might be shorter a half.

`conn.lib': Generic RJ45_LEDS symbol should have only pin numbers. Pin names instead of numbers are acceptable only when manufacturer uses this kind of enumeration. For example FINDER uses A1/A2 for coil and numbers for contacts in their PCB mounted relays (according to the electrical naming standards), but ZETTLER uses only numbers.

ricardocrudo:
The reference used to RJ45_LEDS part is https://www.amphenolcanada.com/ProductSearch/drawings/AC/RJHSE508X.pdf

PR #133, opened 2015-02-28
Add 3SK263 Dual gate MOSFET, 2SC1945 RF B-E-C transistors. Planned adaptation to the KLC rules.

keruseykaryu:
I'm planning to adapt `transistors.lib` to the KLC rules. There's not much to do. Most of the symbols has a pins in the 100mils grid.
The problem are MOSFETs symbols where the gate was placed in the grid of 50mils. For 3SK263 I managed to create a "little monster" and slightly spread gates apart, but for a single gate such treatment doesn't pass.
I do not want to enlarge the symbols and moving the gate to center line is not an option. Can we leave it as is?

The 2SA1016 and 2SC1945 are examples of the proposed changes: unified text height, fields position and visibility.


michal777:
If everybody agree that grid of 50 mils can be used I will change my commit with general symbols. The biggest problem that I have with 100 mils grid are JFETs. If gate is centered it's impossible to differ source and drain.
I see it's good to have pins on 2x grid in black box symbols but in case of transistors it's hard to apply and I can't see advantages.

keruseykaryu:
I also wonder on clearing 'Footprint' fields and leaving only simplest footprints filters. To be honest I do not like the name of footprints in the style of "SOT89-3_Housing" (_Everybody know that is a housing_). But I will not be insist to change this. Among other things, I would like to delete these entries in the official version of the library.

CarlPoirier:
The 100mils restriction indeed is not nice for JFETs, but the 3SK263 suggested is not that bad I think. If you put the pins straight, the pin names and numbers a superposed a little bit anyway. I suggest we make an exception for JFETs.

As for the footprint names, "Housing" can be removed as well. I believe it is a relic of the old relaxed nomenclature.

keruseykaryu:
Sorry guys. The latest GitHub for Win release completely smash my fork. I have to close this request and try to recover my work.

PR #134, opened 2015-03-02
digital-audio: update against KLC

ObKo:
1) All pins 100 mils long, text - 50 mils
2) Bodies filled with bg color (yellow by default)
3) Some pins renamed to match datasheets
4) Documentation and descriptions updated for Cirrus Logic devices
Result:
digital-audio-improve


keruseykaryu:
BTW: Please let me know if you want to make any library KLC compilant. I will add your name before you start to the status table: https://github.com/KiCad/kicad-library/wiki/Status-of-the-libraries

keruseykaryu:
Looks good. Therefore I think you should increase the pin length to 0,150" because the double digit pin numbers are too close to pin endings and symbol border. 
I have also some doubts for TPA5050. The power pins lies on the left and right. This doesn't look as well. The DATA_OUT(15) pin should be on the same level as DATA(2) pin. This way we help to make a short and straight cascade connections.

keruseykaryu:
Now looks much better. 
Thank you a bunch for taking care of this. This way we have got another library KLC compliant.

PR #136, opened 2015-03-05
adding +/-9VA to power lib

ricardocrudo:


keruseykaryu:
Please also create/update documentation fields for aliases.

ricardocrudo:
Can you help with this? what do you suggest to write in this fields? All the other power symbols doesn't have documentation field, so I don't have parameter.

keruseykaryu:
I was talking about the MCP1754 components from a second commit and his aliases.
Power symbols does't have any aliases.

ricardocrudo:
done

PR #138, opened 2015-03-06
microchip: USB3300 - Hi-speed USB OTG PHY with ULPI Interface.

ObKo:
Quite common and cheap USB HS PHY. Used in some STM32F4-based boards for Hi-Speed (480 Mbit/s) USB Host/Device/OTG interface.

http://www.microchip.com/wwwproducts/Devices.aspx?product=USB3300


keruseykaryu:
Looks nice. Thank you.

PR #146, opened 2015-03-14
Create microchip_dspic33dsc

El3ctr0nic:
microchip dsPIC33 Digtal Signal Controllers


keruseykaryu:
Things to improve:
Pin endings does not lie over 100mils grid as required in Kicad Library Convention.
Power pins should have Power Input atrribute not Input.

PR #148, opened 2015-03-15
add analog devices instrumentation amplifiers

jzrebiec:


keruseykaryu:
Things to improve:
- Pins could be shorten to 0,100". This can be done without breaking KLC rules about pin alignment.
- The skew could be increased to 20.
- Pin numbers must be visible in this case.
- Description and value text must have 0,05" height.

amp

jzrebiec:
By "The skew could be increased to 20" you mean this?
ad623an
I think the previous version looks better.

keruseykaryu:
The "Skew" is the one of parameters in symbol properities (first tab) in older LibEditor. This was wrong name, so in recent versions has been corrected. I didn't notice the change because I'm using localized GUI. Sorry for inconvenience.
Generally speaking, the distance between the pin and its name is too short. You have this distance set as 1 and pin names are too close of symbol border. If you change it to the 20 the realtion between border, name and pin will be look like I demonstrate before.

jzrebiec:
Thanks for your help! Now, it should be OK.

keruseykaryu:
Yes. Now it looks much better.
Commited. Thanks.

PR #149, opened 2015-03-16
Add all Spartan 6 symbols

alexforencich:
All Spartan 6 symbols, generated from Xilinx pinout files.


alexforencich:
I'm also working on the Virtex 6 and 7 series parts.  I could organize this in separate library files (spartan6, virtex6, kintex7, artix7, virtex7) if that would make more sense than putting it all in the existing xilinx library.

keruseykaryu:
Well. First of all you must read carefully the [KiCad Library Convention (KLC)](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt). This are sets of rules that you have to follow when creating new symbols, and the footprints if you wish to add them to separate Pretty libraries.
The symbols you created and proposed to merge violates some of rules described in _General Rules for Symbols_ chapter. Especially: **3.1**, **3.3**, **3.6**. Please correct these things in your proposed symbols. And IMO the pin length of 0.250" is a bit longer. Could be shorten to 0.200" or even 0.150" if there is a enough room for pin number at 0.050" height.

> I'm also working on the Virtex 6 and 7 series parts. I could organize this in separate library files (spartan6, virtex6, kintex7, artix7, virtex7) if that would make more sense than putting it all in the existing xilinx library.

Generally, I vote for that and support it. I don't know the exact opinion of other librarians here, but it sounds reasonable. Xilinx made a lot of FPGA devices, and the one collective library will be very huge and hard to manage in the future. 
We already have separate libraries for Microchip MCU/DSC. I think the Atmel's, Ti's MSP430 libraries should be divided in a similar way in the future, once we adapt them to KLC rules too.

Please made a corrections I've pointed above, and wait few days for other comments (especially from Carl and Michal) in this topic.

---

BTW: Would you like to take care of the existing Xilinx library: adaptation of other than yours symbols to KLC and sort a bit this mess?

alexforencich:
I can fix 3.1 and 3.6, but 3.3 is not possible unless there is a way to place the identifying fields in different positions on different symbol components.  Also 3.1 and 3.3 conflict unless the symbol height is fixed to an odd multiple of pin spacings.  As far as I can tell, the identifying fields are fixed to a specific offset relative to the part origin on all sub-units.  Is there a way to specify that on a per-sub-unit basis?

Also, I'm wondering if it would be better to split each I/O bank and each MGT bank into its own symbol and having all of the bank pins on the right-hand side of the symbols for everything except the shared power pins.  This may make the larger FPGA symbols easier to work with.  

Maybe a good way to split the libraries would be to do something like xilinx-spartan6, xilinx-virtex6, etc.  Or would that have to be Xilinx_Spartan6 as per the library rules?

alexforencich:
Unfortunately, I'm not particularly interested in the existing library.  Too many symbols and updating them is very tedious.  However, Xilinx does produce pinout files for many of their chips; I can look in to writing scripts to parse those and generate libraries for some of the older FPGAs as well.  It should be a relatively simple matter to adjust the scripts to read the varying pinout definition formats that Xilinx has released over the years.  That way all of the symbols will be consistent with each other from the get-go.

alexforencich:
Alright, I took another look at the library.  Looks like the issue with 3.1 is on the pins at the top.  I can fix that.  I can fix 3.6 as well.  However, I cannot fix 3.3 unless there is a way to place the name and value references in different positions on different symbol parts.  If this is possible, please let me know how.  Otherwise a non-centered origin appears to be the only workaround to place the text in reasonable positions.  Or I can just turn off references entirely, which is not such a good idea.  Also, I checked the pin lengths, and 0.250 is the minimum due to the long BGA pin names, even with 0.05 text height.

alexforencich:
Fixed issue with 3.1 and 3.6.  I will update the pull request with the new symbols in a new file once we can get the solution to 3.3 figured out.  

Also, it is not possible to give each I/O and MGT bank its own symbol.  This will require more than 26 sub-units, which according to http://en.wikibooks.org/wiki/Kicad/file_formats#Description_of_DEF is the limit.  Instead I will split off the MGT banks and keep everything paired up.

alexforencich:
Pull request updated.  I will fix the origin issue on all the symbols if you guys can provide a method for doing so.  I have libraries for virtex 6 and the complete 7 series standing by.

keruseykaryu:
Take a look at this: http://pastebin.com/736if6G4

alexforencich:
Updated as per your example.  Had to make the symbols quite a bit wider, but the origins are now in the center.

alexforencich:
Updated again to use BGA pin ordering for power and ground pins (A1, B1, AA1, etc. instead of A1, AA1, B1, etc.)

alexforencich:
Alright, all of the kinks should be worked out at this point.  Ready to merge if it looks good to everyone.

keruseykaryu:
I'm still seeing pin endings outside the recommended 100mils grid.

alexforencich:
Do they have to be on a 100 mil grid wrt. each other or the origin?  They should all be on the correct grid wrt. each other, but I may need to make a couple of changes (fix symbol size to increments of 200 mils) to get them on a grid wrt. the origin.

CarlPoirier:
Let's go for split libraries. If you can generate some of the older symbols that would be great as well! We'll remove them from the original library and the cleanup to do will be smaller.

As for the pin grid alignment, they have to be aligned with respect to the origin, and thus with respect to each other as well.

alexforencich:
Sounds like a plan.  I am working on Virtex 5, Virtex 4, and Spartan 3 at the moment.  Virtex 5 is pretty much done, the Virtex 4 transceivers need some work, and Spartan 3 requires a bit more work.  I'm also looking in to the Ultrascale devices.  

I will get the alignment issues sorted out and update the pull request shortly.  

Also, is it possible to get more than 26 sub-symbols per device?  I think there is one device right now that generates 27 sub-symbols, with the last one being [.  Might be advisable to adjust the symbol lettering code in Kicad to jump up to AA, AB AC, etc. when it runs out.

alexforencich:
Updated with corrected pin alignment.

CarlPoirier:
Which one has 27 symbols?

alexforencich:
XC5VTX240T-FF1759 in the virtex 5 library has exactly 27 symbols.  It goes A through Z and then [.  Perhaps I need to double up on the transceivers and put two on a side instead of just 1.

alexforencich:
Maybe I should just close this PR and create a new one for all of the current Xilinx libraries that I am working on?  I have Virtex 5, Virtex 6, Spartan 6, Virtex 7, Kintex 7, and Artix 7 at the moment.  Virtex 4 needs a bit more work, but it's almost done.  Spartan 3 needs a lot of work.  I am also looking at the UltraScale series both Kintex and Virtex.

CarlPoirier:
You can close it and start a new one in which you would reference to here, or simply rename that one.

I'll ask the developers for the 27th symbol.

alexforencich:
Well, I think there needs to be support for more than 27.  Simply adding a letter after 26 would be the most ideal solution.  So X, Y, Z, AA, AB, AC, AD, etc., just like for BGA pin numbering.

PR #151, opened 2015-03-16
Earth symbol

samdolt:
Earth symbols following  IEC 60617
IEC
Kicad

### DIFF
--- a/library/power.dcm
+++ b/library/power.dcm
@@ -148,6 +148,21 @@ $CMP -9VA
 K POWER, PWR
 $ENDCMP
 #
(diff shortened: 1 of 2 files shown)


keruseykaryu:
Could you set text height of default fields to 50mils? I know they are invisible, but last time Ricardo Crudo make these adjustments too in his commit #150 for remaining symbols from `power.lib`, so it will be nice to do not have different size.

samdolt:
It should be ok now.

keruseykaryu:
Commited. Thanks.

PR #152, opened 2015-03-17
adding new parts symbols

ricardocrudo:
Not sure if special is the right place to put RT9701.


keruseykaryu:
RT9701 - Good.
IM Relays - Looks terrible. Where did you find this symbol?
TRS Connector - Make the pins shorter. 0.100" is pretty enough.
CS4245 - Good, but `audio.lib` isn't a good place for that symbol. Move it to `digital-audio.lib`.

ricardocrudo:
I fork IM relay from device:RELAY_2RT

IM Relays - ok, I'll redo it.
TRS Connector - ok
CS4245 - ok

ricardocrudo:
About about this:
screenshot_2015-03-19_09-27-13

samdolt:
In relays.lib we already use this style (from IEC 60617):
Relay

For me, IEC style is more flexible and allow all relays to have a consistent style:
IEC1
IEC2

@keruseykaryu : Which style librarians recommend?

keruseykaryu:
> About about this

That's not the way that we should follow.

> Which style librarians recommend?

I should not speak on behalf of others. I like symbol which you demonstrate the first - because it's my drawing.

CarlPoirier:
Hi Ricardo,

The new relays are great. However, why have 48 of them when they are all the same?
Also, please configure your pin types properly for LM4811, especially the clock.
ISL83491: Please put all pins at 150mils.
Also, we're working towards eliminating the special library. I suggest you put the PRTR5V0U2X in device instead, since it consists of simple diodes.

Thanks for your contribution. I'm sorry I have not got to it sooner. I'm rather busy these days.

CarlPoirier:
Alright!

ricardocrudo:
The relay has a lot of alias, IM00~IM08, IM11~IM17,IM21~IM26 etc. All these relays have the same pinout. Maybe I can remove all those alias and keep only a generic name like "IM relays", what do you think?
About LM4811, what are the modifications? I guess change the clock graph style to clock is one of them. The VIN/VOUT pins are defined as input/output, should them be passive/passive?

keruseykaryu:
> @CarlPoirier: The new relays are great. However, why have 48 of them when they are all the same?

> @ricardocrudo: Maybe I can remove all those alias and keep only a generic name like "IM relays", what do you think?

Unfortunately they are not the same. The different coil voltages and sensitivity should be distinguished in the documentation fields. We can keep one symbol and set of aliases for 0x, 1x and 2x series. So it's better to update doc fileds than remove these aliases.
The 4x series has bistable operation and IMO a bit different symbol should be used for that series.

CarlPoirier:
Okay it makes sense that way.

@ricardocrudo I did spot the clock pin.

@keruseykaryu Any other comment before we merge?

CarlPoirier:
Hi Ricardo,

Thanks a bunch for all of this. They are very great!

PR #154, opened 2015-03-17
CC1200 added in texas.lib

samdolt:
CC1200 symbol, using this reference design http://www.ti.com/lit/df/tidr233/tidr233.pdf for inspiration


keruseykaryu:
Is better place for that symbol will be `rfcom.lib`?

samdolt:
Yes, it will be a lot better. 

It's done ;)

PR #156, opened 2015-03-19
Update microchip_dspic33dsc.lib

El3ctr0nic:
MC series added


El3ctr0nic:
Please let me know what minor fixes are needed to my lib ?
I fix it ASAP.
I have some nice lib for example STM32 Nucleo dev boards RFM radio modules etc. and i want to do evrything perfect before pull request ;]

Regards,
Marcin

keruseykaryu:
The designator should be 'U' instead of 'IC'. The designator field have 0.060" text size but should have 0.050" just like the other texts. The value field also should have 0.050" text size.

Take a look at PIC10 MCU library to see the difference.

CarlPoirier:
Kerusey,

I have suggested Marvin to create a library called "development_boards" and put his Nucleos in it. Arduinos and the likes could also be contributed to that same library.

El3ctr0nic:
I do it in that way ;]. I push pull request when i complete lib with all
nucleo board versions. I have footprint alredy done.
The most time consuming will be made 3d model ;/

I have made my own dev board to dsPic 33MC series. I want to publish it
maybe will be usefull for someone. It is possible to add it to KiCad ?? Are
You interrested??

Baord is compatybile with all TQFP44 dsPIC series (MC, and GP) include
MCP2200 USB UART bridge ICSP conn, USB or external 5V power. I plane made
bootloader.

2015-03-22 19:48 GMT+01:00 CarlPoirier <notifications@github.com>:

CarlPoirier:
Hi Marvin,

You will need to create the boards as templates in the folder of the same name. Please have a look at the examples there already. Sorry for the confusion.

CarlPoirier:
You wil find some more information in doing so in #163

Issue #153, opened 2015-03-17
How to resolve mess and hotchpotch in `special.lib`


keruseykaryu:
I think `special.lib` is a type of bag in which very strange things were thrown. Here are my proposals how to clean up this mess and finally remove this problematic baggage.

However, I still have a lot of doubts where to put some devices indicated by '???'.

| Device | Move to | Info
|---|---|---
| 20RS10 | Lattice.lib | New library
| 74LS629 | 74xx.lib | 
| 754C923 | 74xx.lib or Interface.lib **???** | Correct name is: MM74C923
| 75LBC176 | Interface.lib | 
| 9346 | Memory.lib | 
| AD725 | Video.lib | 
| _ADUC816_ | **REMOVE** | Already exist
| AV9173 | Video.lib | 
| CX7930 | Video.lib | 
| DG884 | Analog_Switches.lib | 
| DP83848I | Interface.lib | 
| DS1267 | Maxim.lib | 
| DS1302 | Maxim.lib | 
| DS1307+ | Maxim.lib | 
| DS1307N+  | Maxim.lib | 
| DS1307Z+ | Maxim.lib | 
| DS1602 | Maxim.lib | 
| DS1621 | Maxim.lib | 
| DS1804 | Maxim.lib | 
| DS7820 | Interface.lib | 
| DS7830 | Interface.lib | 
| DS8830 | Interface.lib | 
| DS90C124 | Interface.lib | 
| DS90C241 | Interface.lib | 
| EL2001 | Interface.lib | 
| EL7242C | Interface.lib | 
| EP1210 | Altera.lib | New library
| EP1810 | Altera.lib | New library
| EP300 | Altera.lib | New library
| EP310 | Altera.lib | New library
| EP320 | Altera.lib | New library
| EP600 | Altera.lib | New library
| EP910 | Altera.lib | New library
| FD1771 | Interface.lib | 
| FD2791 | Interface.lib | 
| FD2795 | Interface.lib | 
| HA2540 | Linear.lib | 
| HA5033 | Linear.lib | 
| HCLD0438 | Display.lib | 
| HCPL2731 | Opto.lib | 
| HD63484 | Video.lib | 
| HD63484_PLCC | Video.lib | 
| HI524 | Analog_Switches.lib | 
| HM53461 | Memory.lib | 
| HM53462 | Memory.lib | 
| ICL7660 | DC-DC.lib | 
| ICM7209 | Interface.lib | 
| IDT7132 | Memory.lib | 
[...]

PR #161, opened 2015-03-24
Added various OneWire symbols

samdolt:


keruseykaryu:
Why not create a simple `maxim.lib`? I have plan to add this library with symbols from `special.lib` #153.

samdolt:
Because Maxim has a lot of 1-Wire chips (~50 IC).

And if we follow strictly KLC rules 1.1 and 1.3, maxim.lib should be Maxim.lib.

CarlPoirier:
Indeed, all libraries should at one point be renamed!

It can be Maxim.lib. I'm not sure if 50 ICs needs their own library.

El3ctr0nic:
Please hurry up with Maxim lib ;]. I want add MAX31855 and some A/D converters

CarlPoirier:
Any of you can create the new library!

keruseykaryu:
MAX31855 - We have `sensors.lib` for this kind of devices.
A/D converters - We have `adc-dac.lib` where they fit perfectly.

So. Even if you don't want to create `maxim.lib` library, you don't need to waste your time waiting for us.

samdolt:
If we don't rename all library before the new library system, I think it is better to keep small caps for new library, to be consistent. So I have renamed Maxim_1-Wire to maxim.lib.

keruseykaryu:
OK. I will commit as is now. Carl as a major commander will give the signal whether rename or not.

CarlPoirier:
Thanks guys.

We definitely have to do the renaming before the KiCad stable release. I will warn people on the mailing lists and proceed to the renaming in one week from now, so on April 8th.

PR #165, opened 2015-03-30
Template for Arduino Uno rev3

pouic13:
Firts pull request for template based on arduino uno board.


michal777:
Looks really nice although I see two things to improve:
- pads are too big, spacing between pad and silkscreen line should be at least equal to line width (0.15 mm)
- holes shouldn't be on Edge.Cuts, they should be pads (footprints containing one pad with proper hole size).

pouic13:
i send a new "pull-request" with modifications ?

michal777:
The holes must be on netlist, corresponding symbols should be on schematic to avoid warnings.

pouic13:
i continu with this arrangements or i do modify other thing?? (in the last commit you have Arduino Mega board also).

michal777:
The last wrong thing that I see is that mounting holes has no numbers.
If pin on schematic has number 1 then corresponding footprint should have number 1 to avoid message while reading netlist:
*** Error: Component 'P5' pad '1' not found in footprint 'Connect:1pin' ***

I don't see more problems.

pouic13:
ok i correct this and i finish others boards.

pouic13:
If you want to check i finished to add the 7 templates.

michal777:
I found:

Micro:
- not connected pins are labeled NC (so they are connected by labels)
- bad paths to 3d (absolute instead of ${KIPRJMOD}/...)
- there is no information on silkscreen which pin is 1
Pro Mini:
- bad paths to 3d
- not all 3d models are in *.3dshapes/ (but I don't know if they should be)
- there is no information on silkscreen which pin is 1
Nano:
- hole is 0.2 mm from edge - it can be easily broken
- there is no information on silkscreen which pin is 1
Fio:
- there is no information on silkscreen which pin is 1
Uno:
- label NC on not connected pin
Mega:
- label NC on not connected pin
Mini:
- not all 3d models (but I don't know if they should be)
- there is no information on silkscreen which pin is 1

pouic13:
This error appear when you lunch an "ERC" test?

michal777:
Sorry, something wrong was with my comment, I updated it.
Connected NC causes DRC error.

pouic13:
What should i do for this error? "there is no information on silkscreen which pin is 1"
3D modules who are missing it's normal, And for board "Nano" holes are correct mesure, in original board holes are close to the edge. but should i change?

michal777:
Usually it's necessary to draw some marker or put "1" on silkscreen near to pin 1 to avoid mistakes. I think that it's important at least in Nano and Micro which looks symmetrical and can be put wrong by mistake (Dwgs.User isn't visible on manufactured board).

If some 3d models should exist please check the paths, some of them are wrong (broken links or absolute paths) and if some 3d models shouldn't be visible it's better to remove them from footprint.

Holes in Nano looks really fragile but according to http://ca.mouser.com/pdfdocs/Gravitech_Arduino_Nano3_0.pdf the hole sizes are: 0.07 inches = 1.778 millimeters (smaller than in the template so better change them).

pouic13:
I place num in layer Dwgs.User like a text it's ok?

michal777:
This won't be visible on board (only silkscreen layer is printed on board). Some information on silkscreen how to connect it to arduino and avoid burning anything would be useful.
It seems to be most important in Micro and Nano (they looks 100% symmetrical)
'1' on F.SilkS should be put near to pin 1 (this pin which is square on these boards).

pouic13:
Sorry i did not understand, so i not change layer just for board "Mini" et "Pro Mini". I indicate "1" on "Dwgs.User" but i don't know if necessary? because this boardd is very concentrate.

michal777:
I think that now it's ok. If you don't plan making more changes and other librarians have nothing against I will merge it.

pouic13:
I just rectify the "Mini" board shematic and so i retested it.
For me it's good.

pouic13:
ho sorry  wait, i found that i haven't add footprint "pin1" in board "Uno" and "Mega" in ".pretty" directory. 

I tested my template and it's very cool to have all footprint in a specific directory. For exemple when you draw your shematic and that inadvertently you erase all anotations. The anotation of template change whith. And when you start "CvPcb" footprint are mixture. If you have a directory who contain all footprint of template ( as i did ) You can found easely footprint in adding the library off your template.

I think it's important (if other persons want to contribut) to keep a directory who contain footprint specific off template.

CarlPoirier:
Hi,

Your templates are incredible yet very much needed by hobbyists!

Your text should read, for example:

"Shield for Arduino that uses 
the same pin disposition 
like "Uno" board Rev 3."

pouic13:
Sorry i'm french and i'hav got a bad english so if you want to correct me you can do it! :)

pouic13:
Hi, I'm writing to find out if is there a problem to merge?

michal777:
It's OK, Carl also looked at this so I will merge it. Sorry for the delay.
Thanks for your contribution.

PR #168, opened 2015-03-31
Added LT3652HV Battery Charger

ChlorideCull:
Chip from Linear Technologies. Datasheet referenced is http://www.farnell.com/datasheets/1692510.pdf.

Rebased onto master.


ChlorideCull:
Closed in favor of #171.

PR #170, opened 2015-04-01
Add Xilinx devices

alexforencich:
Expanded version of #149.  All Spartan 6, 7 series, Virtex 5 and 6 devices.


alexforencich:
I can rework the symbols to one bank per symbol once KiCad updates the numbering scheme.

CarlPoirier:
I asked on the developer mailing list about the units above 26.

alexforencich:
It also might be worth asking about per-symbol placement of name/value fields.  Fixing the placement for all symbols in the same device is not really a sensible implementation.  In this case, many of the pin names are very long so the symbols need to be much wider than necessary to keep the text in the center.  The alternative is to not put the origin in the center, but this is contrary to the library conventions.  Simple fix would be to allow placement of F0 and F1 fields inside the DRAW section and then hiding the F0 and F1 fields in the header.  The values specified in the F0 and F1 fields in the DRAW section would be ignored and the values from the header would be substituted.  This should not break backwards compatibility with existing libraries.

alexforencich:
Also, any opinions on the overall symbol size?  The power symbols can get very large.  Right now, if there are fewer than 200 ground pins, all of the power pins (except for MGT power and VCCO rails) go on the same symbol.  If there are more than 200 ground pins, then 150 ground pins go on the symbol with the other power rails, and additional ground symbols are created for each additional 300 ground pins.  Yes, there are chips that need 2 ground symbols - a couple of the Virtex 7 chips have over 600 ground pins.

CarlPoirier:
Hi Alex,

The number of units is extended to 52 in revision 5579.

As for the value placement, which symbol should I look at?

alexforencich:
Well, it would be better to extend to 26^2 (676) with the sequence A, B, C, ..., X, Y, Z, AA, AB, AC, etc.  You could even use the 'BGA sequence' that skips I, O, Q, S, and X.  This way, I can split out the symbols into 1 symbol per bank and get rid of the pins on the top edge.  Right now I cannot do that as the one with 27 symbols would require 54 symbols, which is above the new limit.  Then I would like to shift the text fields to the top and bottom edges of the symbol.

CarlPoirier:
Would 54 be the most you need?

alexforencich:
I'm not entirely sure.  54 is a rough lower bound.  Maybe I only need 50.  Maybe I need 58.  I won't know for sure until I regenerate everything with one bank per symbol, and I am waiting on kicad support for this.  I also have not tried to generate anything for the ultrascale devices, which may have even more pins than the largest 7 series devices.  I may also further break up the power and ground symbols on the really large devices as having 300 pins on one symbol is a bit much.  I presume that the current change just doubles up the letters after Z (X, Y, Z, AA, BB, CC, etc.) which provides a hard limit of 52.  This may not be quite enough, so changing it right now to something that provides quite a few more symbols is a good idea.  Either the simple 'base-26' counting system, or the 'base-21' system that is used for BGA pin numbers.  I think it would be a good idea to implement a scheme like this now so this is "never" going to be an issue in the long term.

CarlPoirier:
The hard limit is 26², but for having convenient drop-down menus,
Jean-Pierre set it to 52.

What is your suggestion?

CarlPoirier:
Hi @alexforencich, it would be great to merge this soon. Where are you at with this?

alexforencich:
Any update on the status for per-subsymbol positioning of the name and value fields?

alexforencich:
I finally got the spartan 3 (s3, s3a, s3e) device files parsing correctly, so I will include that when I next update the pull request.  The last thing I want to change is the location of the name and value fields.  I want to put these on the top and bottom edges so I can make the symbols smaller.  However, I cannot do this unless I can place the F0 and F1 fields independently on each sub-symbol.

alexforencich:
Also, for the ultrascale devices: should that be split into two libraries for kintex and virtex, or should I put all of those devices in one ultrascale library?  Xilinx provides one zip file for the pinouts for all of the ultrascale devices.  Not sure if anyone has an opinion on that.

CarlPoirier:
Placing the name and value fields independently won't happen soon I think. Just do what you can with the current placement. Also, I suggest splitting Kintex and Virtex.

alexforencich:
Alright, I will split those.  However, I'm going to hold off on any further development on this PR for at least two important kicad features: independent placement of name and value fields, and integrated footprint and symbol libraries.  Some of the FPGA packages are missing pins, so generating the packages and connecting them directly to the symbols in the same operation is very important.

CarlPoirier:
integrated footprint and symbol libraries won't happen anytime soon I fear.

alexforencich:
Well, it's on the roadmap apparently, so I'm going to wait for it.  If people want these libraries finished and merged, then let's make noise on the kicad development list so it gets fixed faster.  I think the weakest point in kicad right now is the fact that there is zero coupling between the schematic symbols and footprints.  The relationships between symbols and footprints are horribly convoluted due to this, and doubly so due to the existence of separate .pretty repos.  For example, in Eagle, the footprints and symbols exist in the same file, and there is a mapping between symbol pins and footprint pins that allows you to select whether you have an EBC, CBE, etc. transistor with the same symbol and footprint.  In kicad this is implemented by having 6 different symbols for each transistor, which is completely idiotic - you can't just select the package in cvpcb, you also have to delete and re-place the symbol.  I also just got burned on a design where the LED pins on the schematic and board were swapped with respect to each other due to a library version conflict.  I had made a copy of devices.lib some time ago to add a couple of symbols.  I was using the latest git version of the LEDs footprint library.  In the time between I copied the devices.lib file, somebody decided to swap the anode and cathode pin numbers.  No errors were produced by kicad at all, so I did not notice the issue until I was testing the boards.  Fortunately, I will only need 2 boards, so I was able to work around the issue by rebuilding two LED towers.  This sort of issue is only possible due to poor architectural design.  So I will wait until the architecture is fixed so this sort of issue is impossible before releasing such a complex library.

jhol:
Oh interesting, I'm thinking of adding the Lattice devices.

codemercenary:
Bump.

Would like this merged!  This looks great, I'm going to see what I can do to integrate them into my current project.

alexforencich:
Update pending on missing kicad features.  I need independent per-sub-symbol placement of name and value fields so the symbols can be made quite a bit smaller.

codemercenary:
I'm looking into that now, I thought it was possible to do this by turning on something like "edit pins per part".

codemercenary:
It would also be nice to be able to edit pins directly from the pin table.  And of course if the the KiCad development team would get with the program and move to github proper.

alexforencich:
Yeah, that would be nice.  The launchpad setup is rather strange.  And when they finally move it to github I will probably be opening about 10 issues.

codemercenary:
Same here.  And the build process needs work--I think part of the issue is that the project is maintained by CERN rather than a software engineering house, and this has resulted in some puzzling design choices.

But they are going quite fast on new things, eventually they will move (I hope).

codemercenary:
Hey I'd like to use the XC7A35T part from your library with the FGG484 packaging, did you finish that one before getting blocked?  Also, were you blocked due to frustration (understandable) or because of a critical missing feature?

alexforencich:
Mainly out of frustration.  It's not really a showstopper feature per se, it's just that the symbols have to be about 1.5x to 2x as wide as they really need to be.  The problem is that the kicad library convention requires the origin to be in the center of the symbol, kicad does not allow the name and value fields to be positioned independently on each sub-symbol, and the sub-symbols are all different sizes.  So the symbols must therefore be wide enough so that the pin labels do not overlap the part name that's hanging out in the center with whitespace all over the place.  I also don't want to commit an updated version later that moves all the pins around when they do finally get around to adding that particular feature as that will mean many nets will have to be moved if the symbol is updated in a schematic.

codemercenary:
I didn't realize they don't allow symbol units to be different sizes.  That seems crazy.  The grounding unit, for instance, can be super compact because all of the names are just `GND`, why would you make that 100mm wide when it's just a narrow strip?

alexforencich:
Exactly.  The problem is the long part value field.  I made the symbol wide enough so that the part name does not overlap the pin names.  What I want to do is put the part name off the top edge and the value off the bottom edge to eliminate this conflict, but this is not currently possible.  

There are a few things that Eagle does that should be ported over to kicad, this being one of them.

codemercenary:
I think I might try putting the anchor a little high of center on the part I'm going to be using.  I seem to be able to get asymmetric sizes working in the latest version, what is unsupported is placement of the part name on a per-unit basis.

rjordans:
Ping!  It would be really nice to see this make it into the library.  Does anyone know what the status is on the label placement issues mentioned above?

jkriege2:
Hi!

just had a closer look:
These seem to be generally in good order (although I'm not really sure how we should ensure that pins are correct...). Did you generate these by a script? Based on which data?

Else some additions would be required:
1. Please provide datasheet-links
2. Please provide correct values for the footprint-field (i.e. in the Form `Library:Footprint`, e.g. `Housings_QFP:TQFP-80_12x12mm_Pitch0.5mm`)
3. Fill FPFilter correctly. For the example above: `TQFP*12x12mm*Pitch0.5mm*`

See here for details on our rules on sybols/footprints: https://github.com/SchrodingersGat/kicad-library/wiki/KLC (Note: This is a link to a version we are currently working on ... oiginal is here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention ... and that new version should end up there soon)!

best,
JAN

jkriege2:
@alexforencich What's the status of you PR? Still on it?

alexforencich:
These are generated from the Xilinx provided pinout files with a python script that groups the pins into the various symbols.  I should probably write a separate sanity checker script that compares the generated library to the pinout files to ensure nothing was missed or mixed up just to be absolutely sure.

Datasheet links are a bit difficult because, for the most part, these devices are split up among several different PDFs.  I suppose I could link to the packaging and pinout datasheet or each series, but this would only be part of the story (no performance characteristics or anything else, only package drawings and pin listings).

Most of the footprints, as far as I can tell, do not currently exist in the library.  I am considering writing generator scripts for the footprints as well (in a number of cases there are several variants of more or less the same package, but with different sets of balls missing) but right now kicad is not really integrated enough to make that very clean.  I have heard some rumors about changing library formats....if that's the case, I might as well wait and rework my scripts to generate the new format at that time.  The main thing I am waiting on at this point is per-subsymbol placement of the labels so I can place the labels at the edge of the subsymbols instead of in the center.  It's insane that kicad doesn't support this, though I suppose it's not particularly surprising given the amount of duct tape and baling wire involved.  Kicad has some very nice features and a lot of potential, especially now that Eagle has gone to the subscription dark side, but there are some very serious architectural issues that need to be addressed, and the workflow and overall usability will continue to suffer until those architectural issues are addressed.

alexforencich:
Also, I need to add ultrascale and ultrascale+ devices.

jkriege2:
Hi!

it would be great if this moves forward ... Concerning the footprints, we have now several repositories for footprints that might help (although not yet really filled ;-)

- https://github.com/KiCad/Housings_BGA.pretty
- https://github.com/KiCad/Housings_LGA.pretty
- ...

Best,
JAN

thequbit:
Any chance this will be merged in soon?  Thanks.

SchrodingersGat:
@alexforencich wow this is a huge contribution, I haven't been following this PR but I've just pulled it down and the symbols are very nice.

Have you made any progress on footprints (I note that last discussion was quite a while ago). Having associated footprints would really improve this contribution ;)

The [footprint generator tool](https://github.com/pointhi/kicad-footprint-generator) is now really good. A high proportion of the KiCad footprints are generated from this tool. 

If you are willing to make this effort, that's great!

If you are not, then I am willing to consider merging this as-is :)

SchrodingersGat:
@alexforencich I am looking to merge this in, wondering if you would be willing to make some changes:

1. GND and power pins could be stacked on top of each to save a lot of space
1. Many symbols have a lot of empty space - they could be squashed together a fair bit

Thanks

dnotq:
@SchrodingersGat, please merge this!  alexforencich has been trying to get this committed since 2015 and has done a great job trying to make the lib acceptable.  So it is not perfect yet, but it is better than nothing and not written in stone (it can be changed and improved, which is the whole point of development and open source).

SchrodingersGat:
Good enough for me 👍

alexforencich:
Really?  There are still a number of things that I want to clean up on the symbols, namely the label placement.  However, I am blocked on Kicad features (independent label placement).

dnotq:
@SchrodingersGat, thank you!

@alexforencich, IMO you have done a great job and done the work no one else wanted to spend the time to do.  There are a lot of things about KiCAD that we have to work around every day, but if we wait for the software to be perfect then we would never have anything to use at all.  When independent label placement becomes available, then worry about the cleanup.  In the meantime, why not make this available so people can benefit from your excellent efforts?

PR #171, opened 2015-04-02
Move battery chargers to a new Power_Management.lib

ChlorideCull:
Also includes pull request #168.


keruseykaryu:
The old component have to be refactored applying KiCad Library Convention rules.
The LT3652HV also violates some of these rules, especially: 3.1, 3.2. 
Take a closer look at the pin types too - power pins can't be signal inputs! This is a frequent issue generated by the newbies.

PR #180, opened 2015-04-23
Add two new Intel parts

lkundrak:
8259, Programmable Interrupt Controller, PDIP-28
8088, 8-Bit HMOS Microprocessor, Minimum Mode, PDIP-40


keruseykaryu:
Please refine pin positions for I8088 symbol. Use 100mil grid as KLC says.

lkundrak:
I may need some help here.

> Please refine pin positions for I8088 symbol.

What changes are needed here? The symbol violates KLC's suggestion that output pins go on the right side. While INTA and HLDA pins are output pins they go in pairs with INTR and HOLD and putting them on the other side of the symbol would harm the scheme readability.

I don't know what else might be wrong.

> Use 100mil grid as KLC says.

Hmm, I can't seem to be able to choose anything bigger than 50 mils in eeschema?

The other modules in intel library seem to use the same grid & pin spacing as my symbols do (or I'm somehow misinterpreting it).

CarlPoirier:
Even though the grid is 50mils, nothing prevents you from putting them
100mils apart! i.e. having your own imaginary grid.

keruseykaryu:
> The other modules in intel library seem to use the same grid & pin spacing as my symbols do (or I'm somehow misinterpreting it).

Yes, you're right. But You doesn't look very carefully. All pins are aligned to 100mils grid. If You zoom out a bit the workspace to see standard grid with double space between dots You will see the difference.

lkundrak:
> Even though the grid is 50mils, nothing prevents you from putting them
> 100mils apart! i.e. having your own imaginary grid.

Ok

> Yes, you're right. But You doesn't look very carefully. All pins are aligned to 100mils grid. If You zoom out a bit the workspace to see default grid with double space between dots You will see the difference.

> That is the trick to make 100mils grid visible.

Understood. Pushed the updated version.

ricardocrudo:
@lkundrak you can use my [script](https://github.com/ricardocrudo/kicad-utils) to check the KLC issues.
Also check the pins graph style, I give a look in the 8088MIN and the pins 31 and 33 shouldn't be inverted. And the pin 23 has the inverted style along with the top line. I guess you should use just one of the sign.

danger89:
@ricardocrudo  Should this tool not be added to the main KiCad, simular to the KiCad Calulator. We need a KiCad tools or something. Which include your tool as well Richardo.

ricardocrudo:
@danger89 I don't know too much about the KiCad code, I know it's wrote in C++ and has some wrappers to python. Those scripts that I did were written in python, I've no idea about how hard would be to have them integrated to KiCad. However, Carl opened a discussion in #191 to suggest to have them as KiCad repository.
Yeah, I agree with you, in the future would be great have a DRC similar button in Schematic Library Editor and PCB footprint Editor to check the KLC rules.

danger89:
'We' can always port the Python script to C++ and then integrate it. I mean you got a working starting point.

CarlPoirier:
Indeed that could be great. For now however, I think it's better to have it
separate until we know the KLC will be in a final state.

danger89:
@CarlPoirier The longer you wait with porting, the more effort you need to make, to convert existing code to C++.

michal777:
I doubt it's good idea which developers will like. In future kicad scripting will be implemented in eeschema and probably then it will be possible to add these scripts as python plugins (I'm not sure, I don't know much about it).
I don't believe kicad as an application should suggest keeping our library convention because everybody can prefer his own library convention and the only reason we have KLC is that we need some consistency in THIS library.

keruseykaryu:
> I don't believe kicad as an application should suggest keeping our library convention because everybody can prefer his own library convention and the only reason we have KLC is that we need some consistency in THIS library.

I totally agree.

We can not force all users to go one way. We can establish our rules, but we *should* use them *only for those repositories*. Users who want to add something here *must* follow the established rules. If they want to create something for *his own use* then we should not even suggest to them how they do it.

I actually use my own rules in my libraries. And although it is difficult to manage two types of libraries in various styles, I do it. Often I draw and publish something using my own point of view, and then process it to adapt to the principles from the KLC and proposing here. Or vice versa when I want to include something that someone else draws for us.

If we force KLC check to Eeschema and Pcbnew library editor, then perhaps there will be more people who will want to have their own style checkers available from this level.
We can reconcile the two sides, if there was a plug-in system in both of programs. Then everyone - who will be able to - can write his own checker and run it. We could also create KLC plug-in and turned it on or off as needed.

This way we give this functionality without forcing it.

danger89:
@keruseykaryu  Thank you, for your explanation.

CarlPoirier:
That is a good point.

danger89:
@ricardocrudo You also like the idea of a plugin-system for KLC check scripts? Instead of repeating yourself to almost every commit, they need to execute your script manually, in order to help the developer improving the lib (component or module).

ricardocrudo:
@danger89 yes, I totally agree with Michal and Kerusey, them explanation was very valuable. It's good to have clear what are the aim of KiCad library repository.

> It can be similar to the DOM plug-in system indeed.

The DOM plug-in is the one used to generated the Bill of Materials? If so, I like this.
I guess the interface to manage this plugins would be the same used in the BOM generation, but, maybe, providing more wildcards (as %I, %O) and also an user feedback window would be necessary.

Even the scripts execution having to be manually accessed, I think is quite more straightforward to users have them with two or three clicks than with opening terminal windows and executing line commands.

danger89:
@ricardocrudo Yes, the DOM plugin-in is used for generating bill of materials (eg. in xls format).

Pending on the script needs, more wildcards are maybe needed. I think a user feedback window is indeed a must-have. But not that hard to make, it will display the same as the output on command-line (stdout) to begin with. 

I hope to see this KLC plugin-in mechanism in KiCad. And that your script will be part of this! :+1:

ricardocrudo:
@lkundrak are you planning to fix the issues?

lkundrak:
@ricardocrudo, yes. Pushed an updated branch, sorry for the delay.

ricardocrudo:
Hello @lkundrak. 8259 still having wrong pins electrical type definition. 8088MIN shouldn't be named as is, it should be simply 8088 and support both modes, renaming pins name as they are in datasheet seems enough.

lkundrak:
> Hello @lkundrak. 8259 still having wrong pins electrical type definition.

Whoops. Fixed.

> 8088MIN shouldn't be named as is, it should be simply 8088 and support both modes, renaming pins name as they are in datasheet seems enough.

Fixed.

CarlPoirier:
please test this

CarlPoirier:
Hi Lubomir,

Thanks for your contribution.

PR #192, opened 2015-05-02
Update `adc-dac.lib` and `Xicor.lib` against KLC

keruseykaryu:


CarlPoirier:
Hi Kerusey,

You have some 60mils text in LTC1257 and TDC1408. Besides that, it looks good.

BTW, what dictates the fill color? The latter and TDC1408 are not filled whereas all other parts in that library are.

keruseykaryu:
It seems that I've missed these symbols.

CarlPoirier:
Hi Kerusey,

Sorry for the slow reply. There was one pin left to fix but I took care of it. Thank you for your contribution.

PR #127, opened 2015-02-23
STM32: Updated against KiCad Library Convention

MateuszKrawczuk:
STM32F050C4(C6)
STM32F050K4(K6)
STM32F100C4(C6)
STM32F100C8(CB)
STM32F100R4(R6)

PR #193, opened 2015-05-02
STM32: Updated against KiCad Library Convention, second attempt

ObKo:
This request includes #127 

All symbols unificated and updated against KLC. Schematic file with all symbols:  https://gist.github.com/ObKo/48caedaf90137eec6004

There is also some further improvements in which I'm not sure:
1. Arrange ports counter-clockwise in stm32f1xx symbols according to rule 3.4 
2. Split ports into separate units for stm32f4xx symbols - their port names are really long.
3. Unificate descriptions and keywords - some common format required.

Any suggestions?


CarlPoirier:
Hi,

Thank you for your contribution. It's very great in the state it is currently already. As for your other improvements:

1. Rule 3.4 says for example that you would have PA, PB, PC and PD in a counter-clockwise ordering.
2. I'm not sure about splitting symbols as most probably it would still require the user to have all units on the schematic and hence not save much space. Would you have one unit per port?
3. As long as the information is there and separated by commas, it doesn't change much, but indeed it could still be great.

keruseykaryu:
Please look here: https://github.com/KiCad/kicad-library/pull/208

ricardocrudo:
Closing this request because it's obsolete, as the requester commented at #208.

PR #195, opened 2015-05-08
Fixed wrong IRF540N and IRF9540N pin assignment

Corna:
I've fixed the wrong IRF540N pin assignment
According to the [datasheet](http://www.irf.com/product-info/datasheets/data/irf540n.pdf) (page 8/9) the correct assignment is GDS, not DGS


keruseykaryu:
The IRF9540N has this issue too. Could you correct both?

Corna:
Done

keruseykaryu:
Thanks.

PR #201, opened 2015-05-14
Added C[CA]56-21* display schematics

Fat-Zer:
Schematic for 7-segment displays:
CA56-21*
CC56-21*

PS: I'm not quite sure about the rules on the vertical align of the components...


keruseykaryu:
IMO the common anode/cathode pins should be on the top not on the bottom.

Fat-Zer:
> IMO the common anode/cathode pins should be on the top not on the bottom.

I've started with common cathode and more straightforward position for it is on the bottom. I had an idea to put pins for CA to the top, but dropped it in favour of convince.

Also In that case DP label slides to the line and that means that the should be redone completely. 

IMO it's not so big deal where the pins will be placed, but if you insist on moving them top I'll do that.

keruseykaryu:
Many people uses this kind of schematic structure for LED displays:

displays

I think it is the best solution. Of course, not everyone may agree with such a statement.

Fat-Zer:
Take a look.

CarlPoirier:
I think it's great now!

keruseykaryu:
It would be great if the pin endings will lay on the 100mils grid in Y direction too.

Fat-Zer:
@keruseykaryu, thanks, missed that after the redraw...

ricardocrudo:
Thanks @Fat-Zer

PR #202, opened 2015-05-14
added 3d shape for smb connector

dfro:
Added 3d shape for smb connector. Footprint for this connector in KiCad/Connect.pretty/pull/20
Connector: [1-1337482-0](http://www.te.com/usa-en/product-1-1337482-0.html)


michal777:
I'm afraid it can't be added to open source library because it's property of TE Connectivity.

dfro:
@michal777  I have made this model form scratch and it does not fully repeats design of TE Connectivity.
Is it still violates TE Connectivity rights?

michal777:
If it is not the model available on their page it's ok.

dfro:
No, it is not.

michal777:
Could you rename it to SMB_Straight and scale the 3d model? Wayne asked to keep these 0.1inch/div units in official libraries.

michal777:
Sorry that I didn't write clear enough about the scaling. Scale in footprint properties should be 1,1,1 (I meant that in wings3d 0.1 inch is one default unit). Now it's 10x too big.

michal777:
Thanks. It's a bit shifted from the centre. Not a big problem, so if you won't correct this I will merge it tomorrow.

PR #203, opened 2015-05-14
Added IR2104 to ir library.

Bosvark:


keruseykaryu:
Please refine pins alignment. All pins are **out of 100mils grid** in X axis.

PR #205, opened 2015-05-17
Add DC Power Jack to DC Barrel Jack component

danger89:
Add also the word 'power', otherwise you won't find the component searching on: "DC Power Jack". Which is too bad.


ricardocrudo:
nice! thanks.

danger89:
You're welcome.

PR #207, opened 2015-05-18
Added NUP4202 and NUP2202

dfro:
Added NUP4202 and NUP2202.
In NCP45560 fixed pin position to fit KLC rule 3.1


keruseykaryu:
Because NUPx202 devices have simple diode network inside, all pins should be referenced as Passive.

PR #208, opened 2015-05-18
[STM32] Generate library from pin-files

DrLuke:
I've generated a library file from pin-files. The script I used to do this can be found here: https://github.com/DrLuke/stm32-librarygen

I couldn't review all generated files, but I randomly viewed through 20 or so, and found no errors.
Please let me know if you find errors or style issues I need to improve.


keruseykaryu:
First of all. The old library has links to datasheets included in documentation files. Your commit will destroy all these links. This wasn't good, but this is happen when creating library from a scratch. 
So, the pull request will be reviewed for merging **only** when you include these links for all proposed symbols.
Please also include a housing name in description for each component. This way people can search library if looking for particular housing.

To check symbols for issues against KiCad Library Convention use Ricardo's automated scripts written in Python: https://github.com/ricardocrudo/kicad-utils

DrLuke:
Thank you for your swift response. I was not looking forward to the manual labour of copying the datasheet URLs, so I just downloaded all datasheets from [st.com](http://www.st.com/stonline/stappl/resourceSelector/app?page=fullResourceSelector&doctype=datasheet&SubClassID=1169).

## Current state of Datasheets
Almost all datasheets are included. There were some uncaught special cases I will have to look at tomorrow and maybe manually add to the library. I only have checked a few devices for a correct datasheet, so there might be some wrong links!

## Help needed
Please, if you have the time, go through a few random devices and check if the datasheet is correct. Thank you!

ObKo:
Wow, I was thinking of writing such script when working on #193, good job!

However, there is some issues:
* Rule 3.1 violation (for example, STM32F100C(8-B)...., port C & D), in some cases (STM32F407I(E-G)) pins placed using 25 (or even 12.5) mil grid....
* Strange pin arragement - all ports are on right side, left side is almost empty (especially in F4 symbols). Also, they're in clockwise order...
* Ugly pin names on left side (should looks like PH0/OSC_IN not RCC_OSC_IN/PH0-OSC_IN)  Pin names on right side looks acceptable. IMHO, Port ID should be first for left side and last for right. I'm not sure if it's fixable using script...
* Do we need unsearchable symbol names (STM32F103Z(F-G)Tx)? Again, IMHO, symbol STM32F103ZFT with alias STM32F103ZGT looks better.
* Maybe, visually separate VDDA/VSSA from VDD/VSS?

Overall, it looks very good, and makes my pull request #193 obsolete :-(

DrLuke:
* I've checked the library with [kicad-utils](https://github.com/ricardocrudo/kicad-utils) and got no errors, and go none. I manually looked at the devices you mention, but got no problem there. Maybe you mean the F0 and F1 labels? I've offset them by 25 mils on the y-axis to visually seperate them from the box.
* I've tried to adhere to rule 3.5 by loosely defining outputs as gpio and everything else as inputs. Is this not acceptable? I realize it creates very empty left sides for very large devices, but it just made more sense to collect all GPIO-ports on the right side, as they're the most relevant pins on the device, similar to an output of a black box. 
    * By clockwise order do you mean that PA0 should be on the bottom and increase in pinnumber and port towards the top? Or should the pin order stay the same, but Port A be at the bottom?
* That's how the pinnames are defined in the pinfiles. I could just translate all '-' to '/' if that would be better?
* Is it sufficient to create aliases like this in order to make them searchable?: `STM32F103Z(F-G)Tx -> STM32F103ZFTx, STM32F103ZGTx` ?
* I thought that seperating VxxA was a bit unecessary, although I could alphabetically sort the pins so that VxxA won't be in the middle of Vxx pins

ObKo:
* I'm talking about pin positions. Take a look (beginning of stm32f407I(E-G)Hx):

> X ADC1_IN1/ADC2_IN1/ADC3_IN1/ETH_REF_CLK/ETH_RX_CLK/TIM2_CH2/TIM5_CH2/UART4_RX/USART2_RTS/PA1 N2 **5430** 7000 150 L 50 50 1 1 I

As the result:
30mil
I dunno why kicad-utils doesn't catch it. @ricardocrudo
* As I understand, CW order is applicable if we have ports on both sides. In our case it doesn't matter (all pins are on right by now). Anyway, we should address question about pin arragement to maintainers...
* At least, PH0-.... should be first, IMHO.
* I think, Yes. Although I have a doubt about 'x' at the end. It's just a temp grade, do we really need it?
* That will be better.

ricardocrudo:
@DrLuke I tested your pull request with the lib checker and I got a bunch of errors here (partially: http://pastebin.com/raw.php?i=STaTBXK0). How did you run the test?

I ran like that:

    [ricardo@crudo:kicad-library-KiCad|pr208 %]$ pwd
    /home/ricardo/devel/kicad-stuff/kicad-library-KiCad
    [ricardo@crudo:kicad-library-KiCad|pr208 %]$ ~/devel/kicad-utils/schlib/checklib.py library/stm32.lib

DrLuke:
I have found the mistake: I've plain simply pushed the wrong version of the library. How embarrassing! This commit successfully passes `checklib.py` for me. 

I will now get to work to implement the perviously discussed changes.

ricardocrudo:
Hi @DrLuke. I don't know if this is a problem but some components name are not matching the 'case sensitive'. For example STM32F038G6Ux is using a tiny x in the library (.lib file) and a capital X in the documentation file (.dcm file). If you have plans to fix this you can use the most recent version of checklib script, now it's checking the dcm file too. Thanks.

CarlPoirier:
Hi @DrLuke, if you can fix this it would be great to merge this before the stable release. Thanks!

DrLuke:
Whats the timeframe for the stable release? I currently have a rather busy schedule, but if the release is near I'll try to squeeze it in before that.

ObKo:
Any progress on this PR?

DrLuke:
* I've made all component names uppercase now. They are equally uppercase in both the component and documentation file.
* I've added aliases to all devices. For example, the device `STM32F071V(8-B)TX` will now have the aliases `STM32F071V8TX` `STM32F071VBTX`
* The library passes checklib.py without any errors

Is there anything you'd like to see added or changed?

PR #212, opened 2015-05-25
Fixed the transistors in library/device.lib to use P not E/C

ajeddeloh:
The generic transistor parts in device.lib use C/E instead of P for the electrical pin type for the collector (or drain) and emitter (or source). This pull req. fixes that and prevents the issues it causes with the ERC. The transistors.lib file also P for electrical type of its collectors/emitters.


michal777:
Thanks for fixing it.

PR #213, opened 2015-06-01
adding new symbols and fixing others

ricardocrudo:
Please, check the commits comments to details.


keruseykaryu:
Ricardo, 74AHC1G00 (and many other single logics) exists in `74xgxx.lib`.

ricardocrudo:
> Ricardo, 74AHC1G00 (and many others single logic) already exists in 74xgxx.lib.

ops, I hadn't seen it there. I'm removing it.

> Also, I'm not so sure that the PCA9685 would be suited as a power management device. Actually, I have not even an idea where to put it in the existing library set.

Yes, to be honest I don't like too. We can think about create the "led_drivers" library, maybe.

keruseykaryu:
> Yes, to be honest I don't like too. We can think about create the "led_drivers" library, maybe.

IMO, the `philips.lib` will be a good place for this symbol for now. If the group of led drivers grow up we can think about separate library.

ricardocrudo:
> MO, the philips.lib (I completely forgot about it) will be a good place for this symbol for now. If the group of led drivers will grow up we can think about separate library.

Done.

ricardocrudo:
I'm closing this PR and I'll create small pull requests.

PR #216, opened 2015-06-06
Added new diode library, NXP BAV99 included.

crackerizer:


keruseykaryu:
1. BAV99 is not an only NXP's product, so please remove the `NXP_` prefix from symbol name.
2. Please improve the A and K pins position. Current placement **still violates** KLC rules.
3. SOT-23 isn't only housing where these diodes were encapsulated, so $FPLIST should be improved.

Consider splitting duo-diodes to make them as a multi part component.

crackerizer:
I have updated the symbol as your guideline. The BAV99 and 99W has specifically SOT-23 footprint, so I only added SOT-23.

ricardocrudo:
What about instead of create this specific part create generic ones as was done for D_Schottky_x2***?

crackerizer:
Dual diode could have too many pin arrangement for a generic symbol, eg. AKA, KAK, AAKK, KAKA. IMO, creating separated component could be easier for user.

keruseykaryu:
At this point I agree with **crackerizer**. `diode.lib` library should include only specific components - like `transistors.lib`.
We must put an end to the excessive flexibility, because it does not lead to improve the quality of libraries.

ricardocrudo:
> Dual diode could have too many pin arrangement for a generic symbol, eg. AKA, KAK, AAKK, KAKA. IMO, creating separated component could be easier for user.

Ok.

> At this point I agree with crackerizer. diode.lib library should include only specific components - like transistors.lib.
> We must put an end to the excessive flexibility, because it does not lead to improve the quality of libraries.

Ok.

@crackerizer following the datasheet BAV99W uses SOT323, thus item 3) that Kerusey said before still unsolved.

crackerizer:
I have removed BAV99W from alias.

ricardocrudo:
merged with small changes, check: 91b62d8b183810fe088a62400de6f11559f7860b
Thanks.

crackerizer:
Thanks @ricardocrudo

PR #220, opened 2015-06-12
add MAX481 inside the Maxim librairy

penoud:
As mentionned I try to push my changes inside Maxim library by adding MAX481. 

Datasheet
http://datasheets.maximintegrated.com/en/ds/MAX1487-MAX491.pdf

It's my first push, I tried hard to follow the rules but if something wrong let me know.
Best regards


keruseykaryu:
First of all, please remove the backup files from pull request. We don't need them.

The symbol is too big. It can be reduced by half at least. Many users complained that the symbols available in KiCad are too big (Especially logic gates if sb need to know).
Pin functions: 1 - RX - isn't an input. 6, 7 - A/B - both should be passive.
Pin descriptions: 2 - nRE - use `~RE~` instead to mark it inverted.

penoud:
DearKerusey Karyu,
Thanks you for your comments. Concerning the size, the only way to get half
is to remove the inside drawing of the component. Did this approach ok with
your rules?
Best regards
Séb

2015-06-14 8:40 GMT+02:00 Kerusey Karyu <notifications@github.com>:

keruseykaryu:
> Concerning the size, the only way to get half is to remove the inside drawing of the component. Did this approach ok with your rules?

You can remove or use smaller grid for drawings.

penoud:
Change the component size and assign correctly the pin name and function

penoud:
Did I have to do a new pull request or comment in previous one is enough?

ricardocrudo:
Hello @penoud. Please check your symbol using [checklib script](https://github.com/KiCad/kicad-library-utils/tree/master/schlib) and also do not include bak, bck files to pull request. Would be better create a new pull request to make sure that git history will not contain the backup files. Thanks.

penoud:
Thanks for explanation. I will try to do so.

PR #223, opened 2015-06-18
makes regul library KLC compliant

ricardocrudo:
Merging this branch will make all parts in regul library KLC compliant. 

Some comments:
* All 3-pins regulators have the same shape
* All 4-pins or 5-pins regulators have the same shape
* AP1117E was splitted between AP1117E(adjustable) and AP1117EXX(fixed)


keruseykaryu:
Ricardo. Tell me why the designator field has been moved under the symbol? Footprint field interferes with the connections on the bottom? My old symbols have been deprived of wider border?

ricardocrudo:
Hi Kerusey.

> Ricardo. Tell me why the designator field has been moved under the symbol? Footprint field interferes with the connections on the bottom? My old symbols have been deprived of wider border?

I took 7805 as reference and I copied the pattern to others symbols. I moved the footprints down because when them are in the middle of the part and the labels are big is hard to read the pins names. Which one are your old symbols?

Surely this is just a initial purpose, I can make adjusts if necessary.

keruseykaryu:
> Which one are your old symbols?

It's better to ask which ones aren't mine.

> Surely this is just a initial purpose, I can make adjusts if necessary.

Use the 0,01'' line width for box outlines. I know this is not a KLC rule, but almost all of KLC adjusted libraries uses this style. In this way we keep one look.
Move designator, value and footprint fields to the top of the positive regulators, and to the bottom in negative regulators. Text should not collide with any pin and a wire connected to it.
Move some switching regulators to the `dc-dc.lib`. First from the shore: ADP1108, etc.
Move BQ2003 battery management device to the `Power_Management.lib`.

ricardocrudo:
> Use the 0,01'' line width for box outlines. I know this is not a KLC rule, but almost all of KLC adjusted libraries uses this style. In this way we keep one look.

OK

> Move designator, value and footprint fields to the top of the positive regulators, and to the bottom in negative regulators. Text should not collide with any pin and a wire connected to it.

I see that the negative regulators have the GND pin pointing up, I understand the idea behind it, but IMO is odd has GND pointing up. Are we going to keep this configuration? Also, I think would be nice have only one standard to all box outlines symbols.

It's OK to put designator and value in the top, but the footprint sometimes is not specified, and sometimes it won't be (at least in the symbol creation time), so if we keep it in the top, we'll have to 2 standards. Yet again, I'm in favor to have only one. What about to have the footprint in the bottom 50(or 100) mils right-side/left-aligned text related to the last pin, if there is one, or centered if not? In this way the text will never collide the pins.

The same approach can be used to designator and value, in the top left-side/right-aligned text related to first pin.

Examples:

example1

example2

> Move some switching regulators to the dc-dc.lib. First from the shore: ADP1108, etc.

I'm going to move this ones, is that right?
switch-regul

> Move BQ2003 battery management device to the Power_Management.lib.

OK

ricardocrudo:
closing PR, too old.

PR #227, opened 2015-06-21
makes Power Management library KLC compliant

ricardocrudo:


keruseykaryu:
Bad designator and value fields positions. Schematics might not be legible with current positions.

ricardocrudo:
Need to be reworked according conversation at #223.

ricardocrudo:
closing PR, too old.

PR #228, opened 2015-06-21
makes ESD Protection library KLC compliant

ricardocrudo:


keruseykaryu:
SRV05 and TPD2S017 should be completely redrawn. Pin names won't be necessary for these symbols.

ricardocrudo:
Need to be reworked according conversation at #223.

ricardocrudo:
closing PR, too old.

PR #231, opened 2015-06-26
Remove repeated component name

hzeller:
In many library descriptions, the component name is repeated as first word of the description.

For example this transistor BC237; the description is
BC237, Vce 50V, Ic 100mA, NPN Epitaxial Silicon Transistor, TO-92

There are others, that are near matches, but still the same repetition; for instance AT89S4051-P
89S4051, SO20, 4k Flash, 256B SRAM

This repetition of the component name is redundant (consider all the lost space in the component chooser that could show relevant information) and error prone to maintain.

A patch to cover up this in the component chooser was less preferred on the mailing list than fixing the problem at the root, so a change to the library was supported (Vesa, JP and Wayne)
  https://lists.launchpad.net/kicad-developers/msg18946.html

This is the patch to make it happen. Also, while at it, there were a few lines in the Zilog and Lattice *.dcm files that had DOS newlines, which is fixed here as well.

Commit message:
  - Remove redundant repetition of the component name as first word in the description.
  - Exact matches as well as slightly shorter names are skipped ("X123-M" as component name would match "X123" in description)
  - Fix non-canonical newlines in a couple of files.


hzeller:
(also, if this is accepted, it might be good to add this as a rule in KiCad_Library_Convention.txt to keep the description free of a repetition of the component name)

keruseykaryu:
> This repetition of the component name is redundant (consider all the lost space in the component chooser that could show relevant information) and error prone to maintain.

IMO it's not redundant. Over the past five years I typed this data for better searching purposes.

For my part, I strongly protest to remove these data!

hzeller:
Maybe in the past that was necessary. But isn't the better search capability in the component chooser that had been added last year more than sufficient ? It does not make any difference if the part name is added in the description or not.

keruseykaryu:
You know my opinion and it is immutable. Sorry.

nickoe:
@keruseykaryu Could you elloborate on exactly why it was better with the sort of redundant partname in the description previously? Was it because the old search dialog would give exact match on something unique in the description?  And hence some parts have a substring of the part in the description?  (I think that was how the old search dialog worked, I could very easily be wrong)

shaneburrell:
Component name should removed from descriptions. Leaving it will only perpetuate having less than ideal data to build better component search.

ricardocrudo:
I'm in favor to remove them. I don't know how bad was the searching in the past, but seems today this extra informations is not necessary anymore.

keruseykaryu:
Well. You (I mean all of you) made a decision to remove "redundant" content. OK. Do it.

I made a decision too - to leave... Recent moves, not only in this topic, assured me made such decision. My vision of how libraries should look has become increasingly divergent.

Now it's time to focus on applying KLC for other symbol libraries, because only 45% of them (from [my calculation](https://github.com/KiCad/kicad-library/wiki/Status-of-the-libraries)) has 100% KLC compilance status.

CarlPoirier:
Since this has been approved by the lead developers, let's merge this. @hzeller, did the checkbox discussed in the mailing list happen?

cpavlina:
@CarlPoirier: no, it didn't.

CarlPoirier:
Thanks for your contribution.

PR #232, opened 2015-06-26
Remove duplicate of the MC34063 regulator in the conn library.

hzeller:
The 'conn' library had an exact copy of the switching DC/DC converter which was out of place. It is already in the proper location in the 'regul' library.

This removes a component, so I guess this might be problematic if someone was using the conn:MC34063 symbol somewhere, but better a clean library than an annoying legacy. Are there ways in the library to 'hard redirect' eeschema to the canonical location ?

Suggested commit message:
  - Remove duplicate of the MC34063 regulator in the conn library
  - Add datasheet link to the MC34063 in regul library.

PR #233, opened 2015-06-26
Remove MC34063, which is a dc/dc converter, from conn library

hzeller:
This is a simplified pull request (see pull request #232) that avoids touching files that are actively being worked on in another pull request.
Does the main thing: remove the mc34063 duplicate in conn library (where it doesn't belong).


ricardocrudo:
Thanks

Issue #186, opened 2015-04-30
Do you accept 3d model files other than wings and wrl?


hyOzd:
Do you accept Freecad or Blender as source format for 3d models? I have a couple models that I would like to contribute.

Also do you accept X3D instead of wrl? Asking this, because blender cannot export to wrl format and it seems kicad works fine with X3D.

michal777:
Some .wrl exported from blender can crash KiCad (older builds are less robust) so I think it's safer to import 3ds or obj from blender to wings and export to wrl.

hyOzd:
What about freecad? Since freecad is a parametric modeler, it can be useful to have a freecad source file around, I think. 

Though, I must add; freecad's exporter features seems to be unstable. A couple days ago, I wasn't able to export a simple cylindrical shape to wrl. Maybe to play it safe, I should contribute Freecad+wings+wrl? But there is no clear path from freecad to wings. Only method I could find atm is this: Freecad export WRL -> Blender import WRL -> Blender export 3DS/OBJ -> Wings import 3DS/OBJ. Maybe I should forget about freecad :)

michal777:
Doesn't freecad export 3ds/obj that can be opened in wings? If it doesn't exporting freecad->blender->wing worked ok for me.

hyOzd:
Freecad can export OBJ. But color information is lost and some normals are wrong. Not really useful.

cbernardo:
Not to mention you may not have colors, and you definitely
won't have colors if you go from FreeCAD to wings then VRML. 
There are my old VRML modeling tools which can be scripted:

http://kicad3dmodels.sourceforge.net/
  


 For very complex models it can take a lot of time to make
a good script, but I think for most models it's much quicker
than Wings3D (and much better) and maybe about the same as
FreeCAD in the time it takes to make a model.

- Cirilo

hyOzd:
Here is another method I use to move from freecad to wings.

- freecad: Export different colored parts as different mesh files (STL)
- blender: import mesh files
- blender: apply colors to different parts
- blender: export to X3D/3DS
- wings: import from 3DS

I use this method when freecad wrl exporter messes up. I think blender can be skipped if you know how to use wings.

But my original question was if I should add freecad files to repository?

s-light:
Thanks for asking this question hyOzd!
i also do most of my part-3D modells in Blender and or FreeCad.
[...]

PR #235, opened 2015-06-29
Cx56-12: added a 3d image

Fat-Zer:
The model is exported directly from freecad. I don't see any glitches in it, but IMO it should be heavier tested on various systems to be sure.

See #186


ricardocrudo:
@Fat-Zer wow, looks very nice. I checked it here and the model is not matching with footprint. The offset I have here are: (0, 0, 0.039370) is it according yours?

Fat-Zer:
yes, it's already matches information in the footprint:
https://github.com/KiCad/Displays_7-Segment.pretty/blob/master/Cx56-12.kicad_mod
```
(model Displays_7-Segment.3dshapes/Cx56-12.wrl
(at (xyz 0 0 0.03937))
(scale (xyz 0.393701 0.393701 0.393701))
(rotate (xyz 0 0 0))
)
```

ricardocrudo:
@Fat-Zer I meant was, with this offset value (at (xyz 0 0 0.03937)) the display does not match. Should be something like (at (xyz 0.254000 0.300000 0.03937)) to match. can you check that?

Fat-Zer:
Oh, I see... I made the model before adjusted the origin to first pin... the right offsets are:
 (xyz 0.250000 0.300000 0.03937)
could you fix it in the footprint too?

easyw:
@ricardocrudo 
Hi, would be possible to add also the STEP model exported from FreeCAD file Cx56-12.fcstd?
https://github.com/easyw/kicad-3d-models-in-freecad/raw/master/misc/Cx56-12.step
the STEP model is useful to be used with kicad stepup script
This script is aimed to give kicad the ability to have an identical view of board and modules in kicad 3d-viewer and in FreeCAD or MCAD software
it can be found from http://www.kicad-pcb.org/display/KICAD/External+Tools, kicadstepup
https://sourceforge.net/projects/kicadstepup/
kicadstepup-3d-comparison

thank you
Maurice

ricardocrudo:
@easyw I don't know too much about 3D in KiCad. I'm bumping your comment here: https://github.com/KiCad/kicad-library/issues/186

PR #237, opened 2015-07-08
Added LM3670MF dc-dc converter to dc-dc.lib

SchrodingersGat:
Alright, let's try this again! I've added LM3670MF to dc-dc.lib, it passes all KLC checks except for "field text uses a common size of 50mils", but as far as I can see all field text is 50mils...


ricardocrudo:
@SchrodingersGat the pins have to be flat to the box outline. Don't make me create another check in the checklib script ;). I see you did the symbol as we see in the datasheet, but the better position to GND pin is in the bottom of the box, take a look at other symbols in this same library. The field that isn't with 50mil is the footprint one, you can't see it because there is no text assigned to it, but you can edit it by clicking in the  "T" icon of toolbar.

Thanks for contributing.

ricardocrudo:
Also fix the pins "electrical type". I suggest you to run checklib script again passing --enable-extra to see other hints.

SchrodingersGat:
Ok, I've made those fixes. It's now only failing EC03 - something to do with the position of the reference, name, footprint? I trawled through the source of EC03.py and tried to match it up - it would be really handy if the checks actually said where the failure is :)

SchrodingersGat:
I'm closing this, because I've made a proper branch in my repo for the changes. The updated pull request is now under my dc-dc branch.

PR #238, opened 2015-07-08
Add PowerSO-20 3D model by Walter Lain

Caerbannog:
This is the 3D model for a PowerSO-20 package.
Exemple reference: L6201PC (H-bridge).

PR #239, opened 2015-07-08
Add Molex KK156 and Molex KK100 3D models by Walter Lain

Caerbannog:
These models (.wrl and .wings) are for Molex KK-156 and Molex KK-100 connectors.

They are copied from Walter Lain's repository:
http://smisioto.no-ip.org/elettronica/kicad/kicad-en.htm
He explicitely grants a compatible license.

I renamed them and cleaned the UNIX permissions (660).


ricardocrudo:
Ok, I just arrived here, ignore my comment at #238

ricardocrudo:
Hmm, molex are divided between two footprint repos: Conectors_Molex and Sockets_MOLEX_KK-System. @michal777 what do you about make only one molex repo or then make the repos names follow an unique naming standard?

@Caerbannog the name of files should to be according with footprint files names, as requested in KLC 10.7 rule (https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt).

michal777:
I agree one repo for molex is better.
The problem is how to make required renaming. The method used by Carl (leaving old library and changing something in library table) is probably not good for users because they typically downloads all the outdated repos with new footprint wizard.

PR #242, opened 2015-07-11
Modify Maxim library DS1267 according rules and add MAX481

penoud:
Checked with the python script


Caerbannog:
The files named `library/maxim.bak` and `library/maxim.bck` are backups and probably shouldn't be included in this pull request.

PR #243, opened 2015-07-11
correct DS1267 and add MAX481 tested with python script

penoud:
without backup files 
checked with python script


ricardocrudo:
About DS1267:
* remove extra square (right side of U?)
* increase the box line width to 0.010
* fill background
* change CLK graph style to clock
* rename Reset to RST and change graph style to inverted or use ~RST~
* DS1267 has different pinout to different packages, so please provide full part number
* set package footprint

About MAX481:
* fill background
* increase the box line width to 0.010
* move anchor to center of the symbol
* improve description: should contain information specifically about MAX481 not the maxim series
* improve tags
* add footprint filters

diggit:
Before committing further changes, please merge upstream master to your branch.

PR #245, opened 2015-07-13
Added 3D models for 2mm Pin headers 

SchrodingersGat:
I have added multiple 3D files for 2mm pin headers

The 3D files were downloaded from - Downloaded from http://smisioto.no-ip.org/elettronica/kicad/kicad-en.htm (released under compatible licence)

All I have done  is batch renamed them according to the Kicad library convention.

Note that the appropriate footprints have been committed here - https://github.com/SchrodingersGat/Pin_Headers.pretty (pull request still pending).


Caerbannog:
There are some changes that don't correspond to your description:
- The first three commits, for instance "`Added LM3670MF dc-dc converter to dc-dc.lib`"
- Your last commit adds a `Molex_Panelmate_15.wrl`

Please clarify if this is intentional.

SchrodingersGat:
Ah, ok. I'm still getting the hang of branching and such. I didn't mean to add these to this pull request. Should I close this one, and re-branch just with the pin headers?

Caerbannog:
Yes, I dont know any other way.

SchrodingersGat:
Ok, will do. Thanks!

PR #254, opened 2015-07-21
Add D_x4_KCom

WillBickerstaff:
Added 4 x Kcom SOT-353 diode & Aliases
Rohm UMN1N
http://rohmfs.rohm.com/en/products/databook/datasheet/discrete/diode/switching/umn1n.pdf

Panasonic MA5J002E0L
http://pdf.datasheetcatalog.com/datasheet/panasonic/SKF00061BED.pdf

MMBD4448HCQW
http://www.diodes.com/datasheets/ds30153.pdf


ricardocrudo:
diode.lib should not contain generic symbols. You can pick one of this part numbers to name the symbol, give it a proper documentation (based on this specific part number) and create aliases to other part numbers that have same pinout. The aliases documentation should also to be according with each part.

ricardocrudo:
Thanks.

PR #258, opened 2015-07-26
added microfit headers straight through hole wrl models

keulYSMB:
Added Straight and Right microfit connectors.
footprint added in: keulYSMB/Connectors_Molex.pretty@4ff9c139e6e97d391f90372e740d8e7cf439f1cf
submitted in the PR: https://github.com/KiCad/Connectors_Molex.pretty/pull/5


ricardocrudo:
Should be better use other color to 3d models, the green used has the same color of the board making hard to see the connectors.

michal777:
If this is the copyrighted stuff from molex website it shouldn't be added to gpl library.

keulYSMB:
The rendering is green if you don't use the "render material property" option, otherwise it'll be shown white.
@michal777: you're totally right the models come from molex and thus shouldn't be redistributed...Thanks for noticing! 

please discard this pull request I do not have time to remodel every connector. I'll stick with footprint only for this time and maybe and homemade 3d models in the future

PR #259, opened 2015-07-27
Added ZXCT1009 high-side current shunt monitor, in SOT23-3 package. 

SchrodingersGat:
The datasheet specifies the pin numbering in a non-standard format, so this schematic symbol varies slightly to the datasheet in that it corresponds correctly to a standard SOT23-3 footprint


SchrodingersGat:
Also added the TI INA19x current shunt(s)
http://www.ti.com/product/ina193

SchrodingersGat:
Should these components be placed in 'sensors' or in 'linear' because they are effectively current sense amplifiers?

SchrodingersGat:
@ricardocrudo I have changed these symbols so that the pins are laid out by function rather than numerical order. 

Can you advise as to whether current shunt amplifier ICs should go in sensors.lib or linear.lib?

ricardocrudo:
@SchrodingersGat The INA* symbols are nice. Please, just fix out pin electrical type.

MCP9700-TT: what about change pins 2 and 3 positions?

what are the differences between MCP9700-TT, MCP9700A-TT, MCP9701-TT, MCP9701A-TT?

LIS331HH and ZXCT1009 seems ok to me.

> Can you advise as to whether current shunt amplifier ICs should go in sensors.lib or linear.lib?

Hmm, hard to classify. I think we can leave them resting in sensors.lib until we find out a best place.

Thanks for your contributions.

SchrodingersGat:
@ricardocrudo I have changed as suggested, and simplified the aliases for the MCP9700/9701 parts

ricardocrudo:
Perfect. Thank you and sorry for the delay.

PR #260, opened 2015-07-27
Added MCP73831/OT to regul.lib

SchrodingersGat:
Single cell lithium (ion, polymer) charge management IC
SOT23-5 package
http://www.microchip.com/wwwproducts/Devices.aspx?dDocName=en024903


SchrodingersGat:
What is the difference between regul.lib and dc-dc.lib? dc-dc seems much more KLC compliant, is regul.lib scheduled for replacement?

What's the best lib for battery charging components?

ricardocrudo:
> What is the difference between regul.lib and dc-dc.lib? dc-dc seems much more KLC compliant, is regul.lib scheduled for replacement?

I don't know. I newer here than this libraries. Probably @CarlPoirier knows.

keruseykaryu:
Because I've made up this partitioning let me answer. The regul.lib was designed for linear voltage/current regulator. The dc-dc.lib was designed for low voltage switching regulators (buck, boost, etc.). The ac-dc.lib was designed for SMPS controllers.

ricardocrudo:
Hi @SchrodingersGat 

Could you please do some adjustments?

**MCP73831/OT:**
* make pin name position offset shorter, like 20
* arrange inputs on left side and output on right side
* if possible, place vdd on top and vss on bottom
* remove package information from description
* I guess you can *charge* from keywords *charger* seems enough. also I would remove *management*, it's quite generic word that does not say too much about the device function

**bq76940:**
* even though datasheet use small letters, part number uses capital, so please, use capital letters to symbol name
* arrange inputs on left side and output on right side
* fix pins electrical type
* remove balance *monitor front end analog* from keywords (maybe add *AFE* )
* add i2c interface to description
* I would go to keep 21,22,23 pins placed along the box, even them are not visible

SchrodingersGat:
Adjustments have been made as suggested.

SchrodingersGat:
Should we make a new library for battery related devices?

SchrodingersGat:
@ricardocrudo All conflicts with the master branch have been fixed. This pull request also moves towards KLC compliance for a lot of the parts already in regul.lib (other than the new ones that I have added)

ricardocrudo:
test this please

SchrodingersGat:
@ricardocrudo I have moved all the battery-related components from regul.lib to batt.lib

See - https://github.com/KiCad/kicad-library/pull/465

SchrodingersGat:
I shall resubmit this as a new PR as it's getting too complicated.

PR #262, opened 2015-07-28
Added BAV99 dual diode in SOT23-3 package

SchrodingersGat:
Checked against .py checker script


SchrodingersGat:
Ah never mind, this already exists in diodes.lib. Closing.

PR #263, opened 2015-07-28
Added LPC11U00 series microcontroller symbol

SchrodingersGat:
(pinout corresponding to LQFP 48 package)

http://www.nxp.com/products/microcontrollers/product_series/lpc1100/series/LPC11U00.html


ricardocrudo:
* rename symbol and aliases adding /201 /301 /401 /XXX
* add documentation to aliases
* link documentation to devices datasheets (example: http://www.nxp.com/documents/data_sheet/LPC11U1X.pdf for 1X devices and so on).
* rename footprint field to LQFP48
* remove aliases with non compatible pinout  (LPC11U6XJBD48)
* I/Os should have electrical type = bidirectional

PR #264, opened 2015-07-29
Dc dc

SchrodingersGat:
Added the LM3670MF switching regulator
Added the ADP2108 switching regulator

Checked against KLC


SchrodingersGat:
@ricardocrudo should voltage regulators be going here (dc-dc.lib) or in regul.lib?

SchrodingersGat:
@ricardocrudo Added another dc-dc symbol (ISL8117) & checked against KLC.

CarlPoirier:
IMO both libraries will need to be checked for duplication and then possibly merged or their content better defined. Let's put them here for now.

PR #265, opened 2015-07-29
Added schematic symbol for the ILI9341 breakout board

SchrodingersGat:
This breakout board (commonly available on ebay for ~$5) has a 320x240 TFT LCD display, with an SPI interface
Also has an SD card slot.
This is the 9-pin interface module (there are other pinouts available e.g. with the addition of touch-screen interfaces.

Fully KLC compliant (--enable-extra)

e.g. http://www.ebay.com.au/itm/2-2-Inch-SPI-TFT-LCD-Serial-Port-Module-Display-ILI9341-5V-3-3V-New-/200953728196?hash=item2ec9c68cc4


ricardocrudo:
Hello @SchrodingersGat Maybe a link to display controller would be more appropriate in the documentation file name.

SchrodingersGat:
@ricardocrudo I think this is what you mean

PR #266, opened 2015-07-29
Added 3D models for 2.00mm pitch pin sockets

SchrodingersGat:
Downloaded from http://smisioto.no-ip.org/elettronica/kicad/kicad-en.htm

Renamed to match KLC rules


SchrodingersGat:
The issues in the pull request for the Pin_Sockets.pretty repo have been addressed

PR #274, opened 2015-08-06
Fixed misspells for Schottky diodes.

nosuz:
I found misspells on the Schottly diode descrioption lines and fixed them. Glad to be pulled this request.

Thank you.


CarlPoirier:
Thank you.

PR #280, opened 2015-08-12
New symbol for the 64-pin STM32F405R

zeroping:
This is the 64-pin version of the 100-pin STM32F405's already in the repo. Hopefully it looks OK to everyone.


ricardocrudo:
Hello @zeroping . Please apply [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) and check your symbol using [checklib](https://github.com/KiCad/kicad-library-utils/tree/master/schlib) script.

zeroping:
Well, I just got off my duff and used checklib on my version. (It's been over a month.) 
Just about all of the parts fail 3.1 and 3.6, even the ones I don't think I modified. Let me go clean this up...

CarlPoirier:
Hi Scott,

Indeed, not all existing symbols are compliant with the KLC because they
were made before it. We however enforce the KLC on new symbols. If you want
to give us a hand at making the whole library compliant, that would be very
much appreciated!

Well, I just got off my duff and used checklib on my version. (It's been
over a month.)
Just about all of the parts fail 3.1 and 3.6, even the ones I don't think I
modified. Let me go clean this up...

—
Reply to this email directly or view it on GitHub
<https://github.com/KiCad/kicad-library/pull/280#issuecomment-150363667>.

zeroping:
Ok, it's fixed now in my version.

I modified one of the library utils to do this:
https://github.com/KiCad/kicad-library-utils/pull/9

CarlPoirier:
Hi Scott,

That's awesome! I looked at it and although not perfect, I'll merge it in this state right now because it is much better than before and no rule is violated.

The only thing I think would be left to do is make the pin lengths uniform. There are 100, 150 and 300mils pins depending on the symbols. If they could all be 150mils that would be great.

Thank you for your contribution!

PR #289, opened 2015-08-31
Added TLC7528 dual 8 bit DAC to texas

IlGuru:
I added the TLC7528 to texas library.
I created it and inserted into the library by using Kicad 2015-07-09 Version.
If I do something wrong please tell me how to fix. I'd like to contribute with the project.


ricardocrudo:
Hello @IlGuru. welcome and thanks for your interest in contributing. The KiCad symbols and footprints follow a convention, please check some guidance how to contribute with KiCad libraries: http://kicad-pcb.org/contribute/librarians/.
Thanks.

---

P.S.:
@nickoe Do you know why KiCad change all lines of the library? Is there any relation with line ending, if so, was this changed recently in KiCad?

nickoe:
@ricardocrudo There were once an issue with the kicad_pcb files, where the track segment order were reversed when read and then saved, but that has been fixed. This could be a similar issue, but I suspect that we have to live with this until eeschema is reworked, if no one starts finding the root cause.

CarlPoirier:
test this please

CarlPoirier:
There are still pending issues with your symbols.

PR #294, opened 2015-09-05
Add box header 3D models by Walter Lain

marv:
Adding all the 3D models for straight and angled box headers created by Walter Lain

These can be used with the VASCH footprints, for which I will send another PR to rename
and rework them. When the will be approved I'll create footprints for the other sizes as well.

box-headers-3d


ricardocrudo:
hello @marv could you adjust those models to match with this PR https://github.com/KiCad/Connect.pretty/pull/33 ? Thanks.

marv:
Hey @ricardocrudo,
please excuse the long delay, I didn't find time to get to this until now.
I updated the naming scheme to "IDC_Header_{Straight,Angled}_[single digit # of rows]x[zero-prefixed double digit # of columns]" to be consistent with other packages like Pin_Header_ {Straight,Angled}_?x?? and Socket_Strip_{Straight,Angled}_?x??

SchrodingersGat:
@marv I'll jump in here as it appears that this conversation has gone stale.

Yes these models should be renamed following the format:

`ICD_Header_[R]x[CC]_{Straight|Angled}`

Note that the footprints themselves should be renamed accordingly (a PR will have to be made to https://github.com/KiCad/connectors.pretty)

marv:
Hey @SchrodingersGat! Thanks for jumping in!

Seeing KiCad/Connectors.pretty#41 I wonder whether the name should also reflect the pitch and technology?
The convention for including the pitch seems to be
`<name>_[R]x[CC]x[Pitch]_{Straight|Angled}` like e.g. `Harwin_LTek-Male_2x07x2.00mm_Straight`
but I couldn't find any footprints that have something like `{SMD|THT}` in their name, as that is usually part of the library name. What would be advise here?

SchrodingersGat:
Good point regarding pitch - I think that is a very good thing to include.

I would add `_THT` and `_SMD` as a final suffix depending on the part. The difference between these (generic) box headers and specific connectors e.g. Molex / JST / etc is that the *type* of connector is explicit in the part number for those connectors.

SchrodingersGat:
@easyw  @jkriege2 @Shackmeister have these 3D models now been scripted? If these models are already in the library then this PR should be closed.

Shackmeister:
Yes they have been scripted but hasnt been made "kicad ready" is there a footprint series?

Shackmeister:
Or rather all the straight connectors has been scripted, but I can easily make the angled ones for a cookie

pointhi:
I think 3d-models for https://github.com/KiCad/Connectors_Multicomp.pretty could be generated with the same script

SchrodingersGat:
fae2e00ca6aa6c61ae351e4a7dbb5ec6

If we can add these as scripted models, then that is preferable!

marv:
I fully agree, it's much preferable to have this generated by a script.
Closing this in favour of scripted models 

*steals a cookie*  Om nom nom

PR #298, opened 2015-09-16
Add 3D shape for TE 1746530-4

sclukey:
This is a 200 pin SODIMM DDR2 socket. The footprint is in KiCad/Sockets.pretty#4. The footprint was created by taking [TE's STP file](http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+View+Model%7FCVM_5-1746530-4%7FD1%7F3d_stp.zip%7FEnglish%7FENG_CVM_CVM_5-1746530-4_D1.3d_stp.zip%7F5-1746530-4) and converting it to a .obj with FreeCAD, then importing into Wings3D and coloring, cleaning, etc.


michal777:
Unfortunately it can't be added
http://www.te.com/usa-en/policies-agreements/terms-of-use-te-com.html
(see: Intellectual Property / Use Restrictions)

sclukey:
Aha, I didn't see that. Sorry about that.

PR #304, opened 2015-09-20
 [STM32] Generate library from pin-files #2

ObKo:
Since #208 is abandoned, I've decided to continue work on generator (https://github.com/ObKo/stm32-librarygen). 
What was changed from #208:
* Ports are generated on both sides in CCW order
* Better pin name handling (e.g. extracting alternate functions from pin names like "PB2/BOOT1")
* Alias generation (e.g. STM32F100R(4-6) is translated to STM32F100R4 with alias STM32F100R6)

Library was checked using checklib.py from https://github.com/ricardocrudo/kicad-utils


DrLuke:
#208 is not abandoned.

yaqwsx:
This patch quite old, will it ever be merged? Or will #208 be merged? Are there any issues I haven't found?

DrLuke:
I didn't get any feedback on it, so I kinda gave up... Feel free to improve it.

CarlPoirier:
Hi guys,

I'm sorry for the lack of feedback in your respective pull requests. If you can merge the kicad-library master and fix the conflicts, I'll have a look at it. I just assigned it to myself.

Regards,

Carl

pointhi:
Hy @ObKo, I have done a quick look onto your library and the first thing I spotted is that you set every pin to Input. I would suggest you to improve this.

ObKo:
Hello!

@pointhi, @CarlPoirier, I will fix issues, regen library and rebase this PR in next several days. I need some time to recall how that script working, because it's kinda write-only...

ObKo:
checklib.py from https://github.com/KiCad/kicad-library-utils running locally shows no violations, but jenkins complains about KLC 3.8 (= no doc) for every part...

DrLuke:
@ObKo: Please let me know if you need any help with the script. I indeed wrote is mostly as a write-only script, seeing how it would only have to be used once.

CarlPoirier:
Hello,

I have looked at the symbols. It's very great in general but there are some things that need to be fixed.

1. Symbols are quite big. In many cases, there is anough room to make them noticeably smaller. With the pin positions and text, you could calculate the remaining space in the middle and adjust the symbol width accordingly.
2. In some symbols you split some ports on both sides of the symbols, for example the STM32F031C4Tx with the PF pins. I understand you wanted to have the crystal pins separated from the rest, but this feature is optional since there is an internal oscillator. Thus they might not even be used. I'd simply put all PF pins together and same some space.
3. I suggest you order the ports in counter-clockwise position as per KLC rule 3.4. I however only suggest you do so if it doesn't make the symbol bigger. For example the STM32F048T6Yx, you could have port A on the left and then ports B, C and F on the right from bottom to top.

As for the documentation, it's also fin on my end. I'm not sure why Jenkins complains.

Regards,

Carl

ObKo:
I've rewritten script a bit and committed new library. Can you check it again?

Also, I'm squashing commits before push, because they're insanely big.

CarlPoirier:
Hi @ObKo,

Your changes are very nice, but I still see some things to fix.

1. Some symbols have extra space between the bottom pin texts and the side pin texts. STM32F038C6Tx is an example of this, whereas STM32F038G6Ux is perfect.

2. Your script doesn't seem to take into account the exposed pads and as such, some symbols lack the VSS pin. See the STM32F042K4Ux.

I found these using random spot-checks in the library. Once you have fixed these two issues, please look at EVERY symbol and make sure everything looks good. I'm waiting for your next commit and I will do the same before merging it.

Thanks for the effort you put into this, it's awesome!

Carl

ObKo:
@CarlPoirier

ObKo:
@CarlPoirier

> Some symbols have extra space between the bottom pin texts and the side pin texts. STM32F038C6Tx is an example of this, whereas STM32F038G6Ux is perfect.

That was imperfect 100-mil grid alignment code, fixed.

> Your script doesn't seem to take into account the exposed pads and as such, some symbols lack the VSS pin. See the STM32F042K4Ux.

My script can handle exposed pads. STM32Cube has a broken xml file for STM32F042K(4-6)Ux. I've checked datasheet and fixed xml by hand.

> Once you have fixed these two issues, please look at EVERY symbol and make sure everything looks good.

I've took a look at all parts, didn't noticed any overlapped or missing pins. In some parts, long-named VDD pins (like VDD_USB) causing rather big empty spaces on top. But I have no trivial fix for that...

CarlPoirier:
@ObKo I'm going through it tonight.

CarlPoirier:
Hi @ObKo,

There's one last thing I spotted that needs adjustment. On some microcontrollers, you indicate the pin function BOOT0 but not BOOT1. STM32F205RB is a good example in which PB2 should be BOOT1/PB2.

Let me know when it's fixed. I'll then proceed with the merge.

ObKo:
> There's one last thing I spotted that needs adjustment. On some microcontrollers, you indicate the pin function BOOT0 but not BOOT1. STM32F205RB is a good example in which PB2 should be BOOT1/PB2.
Again, it's not my fault, ST's xmls for that parts has no BOOT1 function for PB2. I'll check what parts has no BOOT1 and add it manually...

ObKo:
Ok, I've dived datasheets/reference manuals. Some parts (like F0 and L4 series) has no BOOT1 pin at all. For some parts (F1, F4) BOOT1 altfunc is missing in xml. I've wrote some workarounds for that in my script.

CarlPoirier:
@ObKo I'm very satisfied with the results now. You did an excellent job.

However, we have in the original library the microcontrollers STM32F050C4 and STM32F050K4 which are not featured in your library and we'll have to keep them. Can you find similar doc to generate them as well? If not, let's simply append them by hand to yours.

ObKo:
> However, we have in the original library the microcontrollers STM32F050C4 and STM32F050K4 which are not featured in your library and we'll have to keep them. Can you find similar doc to generate them as well? If not, let's simply append them by hand to yours.

That parts are marked as obsolete on ST's website, and aren't included in latest mcu database. But I've found old XMLs and datasheets for that parts.

CarlPoirier:
Awesome work, @ObKo!

Would you be interested in merging your script into Kicad-library-utils?

ObKo:
Hello again, @CarlPoirier 

I'd opened https://github.com/KiCad/kicad-library-utils/pull/15

PR #309, opened 2015-09-25
Added PIC18F442-I/L (PLCC-Housing) behind PIC18F442-I/PT

wolfgangbremen:
PIC18F442-I/L (in PLCC-44 Housing) has a different pinning against DIP40 or TQFP Housing.
The reference to documentation already contained in .dcm-File of is valid for all three variants of PIC18F442 or PIC18F452.


wolfgangbremen:
Noticed an error in Definition of component (DEF-part).
Documentation in .dcm-File is not complete.
Will start a new pull request.

PR #319, opened 2015-10-21
Added 'TEST_1P' (single-pin test-point symbol).

madworm:
I was missing a single-pin test point.


CarlPoirier:
Thanks!

CarlPoirier:
I'll merge it this way even though the text size is wrong. I'll work on fixing this on all symbols today.

CarlPoirier:
We'll be able to fix the field text size library-wide with this script:

https://github.com/KiCad/kicad-library-utils/pull/10

madworm:
I copied the test_point from the this library and just removed the 2nd pin + shifted the text around a bit. I assumed the settings were OK.

PR #331, opened 2015-11-17
Remove the deprecated pretty libraries for the KiCad 4.0.0 release

CarlPoirier:
WARNING: We still need verification about the content of the Capacitors_Elko_ThroughHole.pretty. Has it been moved?

Pretty libraries removed:
Capacitors_Elko_ThroughHole.pretty
Housings_SOT.pretty
Housings_SOT-23_SOT-143_TSOT-6.pretty
Housings_SOT-89.pretty
Housings_TO-12_TO-18_TO-42_TO-43.pretty
Housings_TO-50.pretty
Housings_TO-78.pretty
Housings_TO-92.pretty
Transistors_SMD.pretty
Transistors_TO-220.pretty
Transistors_TO-247.pretty
Power_Packages_SMD.pretty
Power_Packages_ThroughHole.pretty
Pentawatts.pretty
Housings_ROHM.pretty
Sockets_PGA.pretty
Muonde.pretty
Printtrafo_CHK.pretty
Crystals_Oscillators_SMD.pretty
Oscillator-Modules.pretty


michal777:
Capacitors_Elko_ThroughHole.pretty
moved to:
https://github.com/KiCad/Capacitors_ThroughHole.pretty/pull/4

- Only variant without copper + (btw it was reported that they caused errors not reported DRC).
- There are no universal variants in Capacitors_ThroughHole.pretty

CarlPoirier:
Thanks for the information. I can move the universal variants on Wednesday.

CarlPoirier:
https://github.com/KiCad/Capacitors_ThroughHole.pretty/pull/7

CarlPoirier:
Alright so it's ready for review and merge. Let's not forget to delete the Github repositories as well.

CarlPoirier:
All corresponding repositories have been deleted.

djBo:
This breaks existing older installations of KiCAD!
It is now pointing to empty and non-existent repo's with no fallback.
CvPcb is now completely useless, completely halting development on an existing board. Thanks! :((

michal777:
Don't use "online" libraries. They may change without notification. Please use your own local copy of the libraries. You can easily get them using library wizard.

CarlPoirier:
@djBo Just update your fp-lib-table with the one here: https://github.com/KiCad/kicad-library/tree/master/template

As Michal said, if you don't want to mess with this, just create local copies with the wizard. For the upcoming stable release, library repositories will be tagged and so KiCad will use the same version throughout time.

Regards,

Carl

stevenj:
Just a prediction but this "Issue" will cause a lot of grumbling when the new stable is released.  Its a pretty hostile change, with no graceful upgrade, and the error message makes things just look "broken".  And the comment by @michal777 about "Dont use online libraries" is also not helpful, I don't use online libraries but kicad installs them into your project "by default".  So whether you USE them or not kicad still looks them up.

michal777:
I don't like the default online fp-lib-table. Since it appeared people have problems with it (many people think it can't be changed to normal). I think it's pointless to fetch library each time instead of open library from your hdd because users are not interested in observing changes in newest and greatest online libraries. People need something more stable and reliable (independent of internet connection). Not everybody reads a manual before starting work, so the default options are used and may cause problems.
I don't know how the issue will be solved in stable release. @CarlPoirier - do you have more information?

stevenj:
@michal777 I agree with you 100%.  It is very nice, when looking for a new component for a design to be able to browse online libraries, if you want, and fetch what your after and copy it to your local library.  Having that ability is awesome.  But the default should be local, searching online should always be an option, so that upstream changes like this dont mysteriously break things.

PR #335, opened 2015-11-28
[fix] scale down diode bridge

pointhi:
The symbol for the diode bridge is currently very big, so I scaled it down to about the half size, which is more than enough size for this component.

I created a image to show the improvement in size:

diode_bridge_scaledown


CarlPoirier:
Thanks!

PR #336, opened 2015-12-03
Transistors: fixed PN2222A

BartDeWaal:
I noticed the pinout for this part was wrong, so I fixed the pinout to match datasheet

Changed keywords to words that are related to the part

Removed default non-existant footprint. The filter is set correctly, so
it should be easy to find a correct footprint.


CarlPoirier:
Indeed that was a weird NPN! Thanks.

PR #340, opened 2015-12-07
added 1V0 and 1V35 symbols. Fixed text size to 50mil to comply with KLC

keulYSMB:
All violations are fixed except the rule 3.8 violation (descrition keyword and datasheet) because it didn't seem to make much sense to add dummy information in order to comply with KLC.


ricardocrudo:
That's great. Thanks.

PR #344, opened 2015-12-14
added Texas isolators (Dual,Quad)

diggit:
I've added 2 and 4 channel isolators from TI. Checked using checklib.py with parameter --enable-extra passed without problem.
```
checklib.py -v --enable-extra kicad-library/library/texas.lib 
....
checking component: ISO7320C
  No violations found
checking component: ISO7321C
  No violations found
checking component: ISO7340C
  No violations found
checking component: ISO7341C
  No violations found
checking component: ISO7342C
  No violations found
....
```


ricardocrudo:
perfect. thanks.

PR #346, opened 2015-12-15
Added CYBL10x6x-56LQ PRoC, CY84xx7LQI-4xx PSoC BLE Low Energy Bluetooth Components

ashtonchase:
Added Cypress Semiconductor PRoC and PSoC 4 BLE Components. Created generic parts for multiple configurations of the part, with the same pin-out & footprint.

If I need to create a part for each specific configuration, I can. I just imagined that the specific configurations could be specified when entering component value.

First pull request, so comments/suggestions are encourages. I did reviewed the Library_Convention Guide prior to committing, but may have misinterpreted something. Thanks.


ricardocrudo:
@ashtonchase Hey, thanks. great work. Just a question. Is VCCD pin really a power output type? Also, I'd like to suggest to make the pins length 100mil shorter, to be according many other existing ones. Thanks for contributing.

ashtonchase:
Thanks for the response @ricardocrudo . 
With regards to VCCD, it's a pin used to connect a bypass capacitor to the core logic voltage regulator, internal to the device. In that sense, it's not meant to be used to supply power to devices, just for decoupling.

I also updated general descriptions on the alias parts. 
Also, I updated the original pull request comment to include the fact that I was pushing the PRoC (CYBL10x6x-56LQxx) and PSoC BLE (CY84xx7LQI-4xx) chips.

Here is a link to the PROC datasheet: http://www.cypress.com/file/139841/download
Here is a link to the PSoC BLE datasheet: http://www.cypress.com/file/137466/download

CarlPoirier:
Hi @ashtonchase,

I have one last comment. As per KLC rule 3.4, please order the ports counter-clockwise.

Regards,

Carl

ashtonchase:
Hello @CarlPoirier,
I interpreted your comment as updating the order or ports 3,4, and 5. Is that correct?

CarlPoirier:
It's great now. Thanks for your contribution!

PR #349, opened 2015-12-20
Atmel SAM4E 32bit schematics to atmel.lib

M-Bab:
Adds a 144 pin 32bit microcontoller to the Atmel database.


pointhi:
test this please

pointhi:
I suggest to change the pin type from passive to a fitting type (bidirectional, input,...) like it's already done for other microcontrollers.

M-Bab:
I tested it successfully against the fussy Kicad rulings before opening a pull request.
In my experience "passive" is the most appropriate connection type for the highly configuribly pins of a MC. Otherwise you have a high risk of getting a lot of ERC spam. Such a pin can be basically used as input, output, voltage sink or even voltage source ...

M-Bab:
Actually I don't want to sound angry at all but I am a bit confused about the reactions to PRs. There are still quite a lot open and a lot is carefully rejected first. This might really slow down the KiCAD community development and the enthusiasm - in contrast to other EDA software the KiCAD part database could explode with the community help.

Where is the limit at this point? 
1. Are you short on git maintainers to check and accept the PRs?
2. Do you want to optimize quality before quantity? Do you actually want to keep the database small for performance reasons?

If its 1. Maybe I would volunteer as a PR maintainer.
If its 2. I would suggest to open another git repo collecting the "dirty" parts with the goal of a maximum-sized part database. And people who want to use it can easily add it in their KiCAD config. If you don't want to go for that I will ;-)

pointhi:
I'm only a junior maintainer, but I try to answere your questions:

> 1. Are you short on git maintainers to check and accept the PRs?

From what I know, yes.

For example, currently only @CarlPoirier, @ricardocrudo and me are merging PR. But there are some other active contributors which also give feedback and help us a lot without beeing maintainers.

> 2. Do you want to optimize quality before quantity? Do you actually want to keep the database small for performance reasons?

Perfomance shouldn't be a problem.

Many PR get mostly rejected in their first version due to KLC issues. Beside of KLC issues there are other specified things like using the same dimensions and pinordering for similar components, input/output definitions, placing of pins, similar footprint naming, validating the footprint against the datasheet,...

Yes, Quality is important, otherwise there wouln't be a KLC. Also peer review is important to fix most issues before merging.

> If its 1. Maybe I would volunteer as a PR maintainer.

For that you have to contact @CarlPoirier. Normaly first you have to get some footprints/components merged to show you understand the KLC and get used to the process. But it would already help to check other PR for issues and point them out.

> In my experience "passive" is the most appropriate connection type for the highly configuribly pins of a MC.

From my knowledge, passiv describes something which cannot work as source (normally used for passiv components). Bidirectional on the other side is the correct usage, because GPIO pins are configurable as output as well as input.
Furthermore, possible ERC issues doesn't matter. Components should represent real world behaviour, and when ERC issues occour, either the ERC tool has to be fixed, or the incorrect mapped footprint.

hackscribble:
Hi @M-Bab 

Sorry your pull request got forgotten for so long. If you are still interested in contributing, please open a new pull request in synch with the current master branch.

Many thanks

PR #353, opened 2015-12-30
Fixed and added several Japanese BJT entries in "transistors.lib".

tahiro86j:
The so-named "2SA1016" appears to be "2SA1015" judging from its datasheet, for which a URL is provided with the entry. Therefore, "2SA1015" has been created as a renamed-copy of "2SA1016" and "2SA1016" has been deleted. Doing this makes better sense since "2SA1016" does not exist in the real world, and the existing "2SC1815" and the new "2SA1015" are complimentary pairs.

[Updated Jan.1 2016 at 12:53(UTC)]
This pull-request now consists of;
a.)Several improvements/fixes across two BJT entries already existing within the KiCad:master - 2SA1016 (renamed to 2SA1015) and 2SC1815
b.)Addition of six new entries for 2SA970, 2SC2240, 2SA1837, 2SC4793, 2SA1943 and 2SC5200. Note that these BJTs are widely used by hobbyists in Japan and abroad, and are also often found in use for relatively recent discrete audio amplifiers.


CarlPoirier:
Hi Tahiro,

For now I suggest you create one pull request for the corrections to already existing symbols. Then, once we are ready with your new set of BJT, you can open a second one.

Regards,

Carl

tahiro86j:
Okay, sounds good.

tahiro86j:
@CarlPoirier Done as instructed with #355 . Sorry for creating such a mess with #353 .

PR #355, opened 2016-01-02
Fixes for 2SA1016 and Japanese PNP-BJTs.

tahiro86j:
a.)The given URL for the symbol of "2SA1016" points to the datasheet of "2SA1015", and also judging from the given values for Vce and Ic being 50V (corrected to -50V for PNP) and 0.15A (corrected to -0.15A for PNP), it's quite assuring that the entry should be renamed to "2SA1015" from "2SA1016". This correction makes even better sense since the NPN-complimentary of "2SA1015" is the existing "2SC1815" and they are widely used in Japan (and by many audiophiles across the globe).

b.)For better technical correctness, values for Vce and Ic have also been corrected to reflect the that any "2SAxxxx" and "2SBxxxx" are PNP-BJTs. (This applies to the now-renamed "2SA1015" and "2SB631" and "2SB817".)


tahiro86j:
Further discussions may be required to determine the treatment-policy of Vce and Ic (Vds and Vgs for FETs) values for PNPs (Pch-FETs). Currently, most PNP and Pch entries appear to have these numbers written in positive values while they should be negative for technical correctness.
This may seem trivial to many, but with most datasheets indicating these values in positive AND THEN saying "negative for PNPs" on a side note, there is virtually no reason to indicate them in positive values, and/or excuse to leave them unfixed.

tahiro86j:
Now, there are upcoming commits consisting of some thirty Japanese BJTs newly added.
Symbols for TO-92 and TO-126 devices are simple duplication of 2SA1015(PNP) and 2SC1815(NPN) Still not sure of what to do with device-symbols for TO-220 and TO-247 since most of entries with these footprints use symbols with "spikes" sticking out. Which one should I use? Any advises?

CarlPoirier:
@tahiro86j It definitely makes sense to have the negative V_ce for PNP. Either that or we indicate V_ec instead!

CarlPoirier:
Do you have some pictures of the mentioned spikes?

tahiro86j:
@CarlPoirier 
a.)Vce issue;
I strongly support the "V_ce" instead of "V_ec". I suppose NPNs (and Nch-FETs) are probably more commonly used than their complimentary counterparts, mainly because of switching applications today. That said, it would be far less confusing to have clear indications of "negative" voltage in Vce, for PNPs (and Vds for Pch-FETs). I suppose +/- of before the numbers are harder to get wrong, than getting "V_ce" or "V_ec" right.
That way, people will know: "negative Vce ---> it's a PNP/Pch-FET".

b.)Spikes;
Not sure if I can paste images here, so I 'll just keep it brief;
For example, the symbol for "BUT11A" indicates that it is a TO-247 BJT and is spiked. I guess this means that it is a small Power-BJT. The same rule appears to be in place for TO-247 FETs (e.g.IRF540N). This makes great sense since they are screw-mountable to heatsinks, meaning they have relatively large max. heat-dissipation.
With these rules seeming apparent, I would even go further (if I were allowed to, of course) and say BJTs in TO-126 should also have spiked symbols since they are also screw-mountable to heatsinks.

Again, I am up for changes for the better, yet I also insist that these changes should be made gradually while minimizing the influences to existing users.

keruseykaryu:
Let me, as an inventor, to say some words about "spikes" and other graphical diversities in `transistors.lib`.

If you take the whole view you can find a four basic group:
  * small signal transistors like BC547 (Ic/Id below 1A), 
  * low power transistors like BD137 (Ic/Id about 1A or higher), 
  * power transistors like TIP120 (Ic/Id about 5A or higher),
  * and finally high power transistors like BD249 (Ic/Id more than 20A).

The first two group usually don't need any heat-sink to work. The last two group usually has a heat-sink. The "spikes" just indicates that this mechanical part can be used. Of course there are applications where we don't need heat dissipating. In these cases we - designers - have just information so device can take over more power "on his back".
So. When I draw a first time some symbols I just want to distinguish these memberships. As you can see the BC547 has a "normal" ring, BD137 has a quite wider ring, TIP120 has wider ring and "spikes", BD249 has wider ring and "spikes" over the secondary ring.

That's all folks. Now it's your turn to keep this graphical diversities or flatten them.

tahiro86j:
@keruseykaryu 
Much appreciation to details provided by your post. I will be looking over the library in coming weeks, especially while paying extra attention to the above details.

PR #358, opened 2016-01-03
Added 3 IC symbol/libraries

trougnouf:
Hello all,
I created three symbols:

antenova:
    M10478-A2 GPS module
diodesinc:
    AP3429/A 1.0MHz, 2A step-down DC-DC buck converter
melexis:
    MLX90393 Micropower Triaxis Magnetometer


pointhi:
I would suggest you to use existing libraries, like:
* ```rfcom.lib``` for the GPS module
* ```sensors.lib``` for the Magnetometer
* ```dc-dc.lib``` for the DC-DC buck converter

trougnouf:
Done, thank you pointhi!
I also added a yellow background to match the other libraries.

pointhi:
Ok, you have little KLC issues which have to be solved:

```
checking component: M10478-A2
  Violating Rule 3.6
checking component: AP3429/A
  Violating Rule 3.6
checking component: MLX90393
  Violating Rule 3.6
```

You can use https://github.com/KiCad/kicad-library-utils to check and fix those issues.

Furthermore I would recommend to look over the following things:

#### AP3429/A
* change the pin lenght to 100 mil
* change the reference to ```U``` like in the other symbols
* propably using left align instead of center allignment for the text fields
* probably moving the GND to bottom like in the other symbols

(simply look how other dc-dc symbols are made)

#### M10478-A2
* change the pin lenght to 100 mil
* the datasheet url return ```Page not found``` to me
* propably using left align instead of center allignment for the text fields

#### MLX90393
* change the pin lenght to 100 mil
* remove the useless ```N/C``` pins from the symbol
* propably using left align instead of center allignment for the text fields

@ricardocrudo or @CarlPoirier:
* is ```U``` or ```IC``` somehow prefered for symbols or are thy used equally
* what about the GND pins of M10478-A2. Should they be on one side, or aligned more space-saving

trougnouf:
Done. Thank you!

I ran into an issue where changing the font size did not satisfy checklib.py (even after creating an entirely new part). I had to manually modify the lib files as follow for all three parts:

>DEF AP3429/A U 0 40 Y Y 1 F N
>F0 "U" -200 300 50 H V L CNN
>F1 "AP3429/A" -200 200 50 H V L CNN
>F2 "" 0 -400 **60** H V C CNN
>F3 "" 0 -400 **60** H V C CNN
>[...]

into 

>DEF AP3429/A U 0 40 Y Y 1 F N
>F0 "U" -200 300 50 H V L CNN
>F1 "AP3429/A" -200 200 50 H V L CNN
>F2 "" 0 -400 **50** H V C CNN
>F3 "" 0 -400 **50** H V C CNN
>[...]

pointhi:
Hy @trougnouf, I would have some additional suggestions to improve your footprints:

## AP3429/A

* remove the spacing between the footprint body and the reference and value text fields
* when you look onto other dc-dc symbols, they have more or less a similar look. It would improve the compalibity using the same symbol size and pin positioning (like VIN is on the top left side) on your symbol.

## M10478-A2

* remove the spacing between the footprint body and the reference and value text fields
* change ```IC``` to ```U``` as mentioned in the [IEEE 315-1975 standard](https://en.wikipedia.org/wiki/Reference_designator)
* @ricardocrudo @CarlPoirier: what do you think about the GND positioning?

## MLX90393

* remove the spacing between the footprint body and the reference and value text fields
* change ```IC``` to ```U``` as mentioned in the [IEEE 315-1975 standard](https://en.wikipedia.org/wiki/Reference_designator)
* please change the positioning of the pins (especially grouping the SPI/I2C bus pins in a logical way together)

the rest looks good to me so far.

hackscribble:
Hi @trougnouf 

Many thanks

PR #361, opened 2016-01-05
Added STM32F030F4 to STM32 Library

nagelkl:


pointhi:
test this please

pointhi:
Thanks for your contribution.

yaqwsx:
Asking just from curiosity: Why is this PR merged preferably rather than #304? Is it because every symbol was manully reviewed?

pointhi:
Thats a good question. Shortly because I'm a new maintainer and doesn't know there was already a pull for this microcontroller pending.

I will look into #304 to find PR blockers, but merging such a big PR would be done by more experienced maintainiers.

nagelkl:
What is the benefit of auto generating the library as suggested in PR #304? In my opinion it is verry dificult to test if the parts are generated correctly

yaqwsx:
Yes, it is difficult, however in my opinion you can get a complete library with no missing symbols almost for free and the library can be easily updated, if new MCUs are released. For me it outweights the complexity of scripting and testing. However this is just my opinion.

PR #377, opened 2016-01-24
Add IGBT, MOSFET optocoupler gate drivers, and IRS2181(S) half bridge drivers, development_boards.lib added

El3ctr0nic:
Add IGBT, MOSFET optocoupler gate drivers and IRS2181(S) gate drivers


El3ctr0nic:
Preview of IRS drawing 
irs2181

El3ctr0nic:
On this draw IRS should be filipped ;)

ricardocrudo:
hello @El3ctr0nic. Please fix pin 5 electrical type of IRS2181. Also increase symbol body width to 0.010.

ricardocrudo:
Another thing, instead of use slash to  separate the footprint packages (DIP/SOIC8) use alias, because they have different part numbers. Thanks.

CarlPoirier:
Hi @El3ctr0nic,

For the LGA package, you have to open a separate pull request in a Pretty repository. I suggest you do so here: https://github.com/KiCad/SMD_Packages.pretty

For the Nucleo board, it needs to go in the kicad-library/templates directory. We need to have the connector schematics and the footprint. I suggest you take a look at the existing ones as an example.

So please remove the LGA package and the development_boards.lib from this pull request and I'll have a look at the rest.

Regards,

Carl

PR #378, opened 2016-02-01
SPDT switch added; MCP16301 DC-DC step down regulator added. 23LCV1024 Microchip SPI SRAM added.

mpkopec:


mpkopec:
Can someone help me with those violations, please? I have changed all the texts to 50 mils and jenkins still spits out that my components are wrong. Thanks in advance!

mpkopec:
Is anyone able to help?

reportingsjr:
Hi,
I really only help the library maintainers by trying to automate things, so I can only say so much about this.

Two of the errors you are getting are because the footprint and documentation field of the symbol are set to 60 mils rather than 50 (error 3.6). You can fix this by clicking "Add and remove fields and edit field properties" in the library editor and setting those fields to 50 mils. I know it is kind of dumb because they are blank, I'll add it to the list of things to fix.

The other (error 3.8) is because you didn't put a datasheet for the switch symbol which makes sense. So you can ignore these errors.

Don't expect this to get pulled in any time soon though. The few library maintainers are pretty overwhelmed. Sorry!

mpkopec:
Thank you Jon for your quick answer! I didn't know that something is 
still hiding with 50 mils. Although I did put online datasheets to 
components, but I will double-check.

Thanks again,
Maciek

PR #380, opened 2016-02-04
Create sanken.lib and add some motor drivers

trorornmn:


SchrodingersGat:
@trorornmn I would suggest adding these components to the "motor_drivers.lib" library, rather than creating a manufacturer-specific library for such a small number of components. This will help others to find and use your components :)

trorornmn:
@SchrodingersGat I moved symbols of "sanken.lib" to "motor_drivers.lib". Thanks.

SchrodingersGat:
Thanks @trorornmn .

What are the differences between the different aliases of the parts? It would be helpful if the description for each alias was slightly different to highlight the difference(s) between them. At the moment you cannot differentiate between the different versions!

trorornmn:
Thanks for your reply.

The difference between SLA7070~SLA7078 is motor current value.
The difference between SLA7075~SLA7078 is also the same.
The difference between SLA7070~3 and SLA7075~8 is presence or absence of micro step function. SLA7075~8 have micro step function.

SchrodingersGat:
@trorornmn can you please update the description for each variant to indicate these differences? :)

trorornmn:
@SchrodingersGat You mean that I should add the description to the line which starts from 'D' in motor_drivers.dcm, right?

SchrodingersGat:
Exactly. If you have a look at the parts A4950E and A4950K in motor-drivers.lib, you will see that they are differentiated in their description field.

trorornmn:
I updated the description in motor_drivers.dcm.
By the way, there is a datasheet and an aplication note for each motor drivers, so which URL should I write in dcm file?
Or, should I write a URL to the web page which includes both links to these files (Like this http://www.semicon.sanken-ele.co.jp/ctrl/en/product/detail/SLA7070MPRT/)?

SchrodingersGat:
These are coming along nicely! I think that the datasheet should be fine, but can you add the /en/ version not the /jp/ one?

Also, do you have a footprint for this? As far as I am aware there is not a ZIP23 footprint in the library to match against (the default footprint should explicitly point towards a footprint that actually exists)

trorornmn:
I replaced /jp/ version datasheet with /en/ one.
I don't have ZIP32 footprint. I want to make it, but I have no idea which .pretty repo should I use.

SchrodingersGat:
Ok, well for now, as there is no correct matching footprint, please remove the default footprint, and add "ZIP23" as a footprint filter.

Then we can merge, and footprints can come later. Cheeers

trorornmn:
Removed default foot print, and added ZIP23 in footprint filter section.

SchrodingersGat:
Awesome, thanks for the work, appreciated :)

trorornmn:
Thank you so much! I really appreciate your patience and help.

PR #384, opened 2016-02-09
Add OTA (CA3080 LM13700) to linear.lib

kinichiro:
- Add 2 Operational Transconductance Amplifier (OTA) components to linear.lib
- CA3080 is single OTA
- LM13700 is dual OTA
- KLC Compliance

Here is a component image.
ota

Thanks.


ricardocrudo:
They are really good. Just a small detail, documentation for CA3080A (alias) was missing.
Thanks.

PR #385, opened 2016-02-09
Add PT2399 to adc-dac.lib

kinichiro:
- Echo Processor IC
- CMOS ADC DAC 44K Digital processing VCO
- KLC Compliance

Here is a image of component.
pt2399

Thanks.


ricardocrudo:
Thanks.

PR #387, opened 2016-02-09
Add LSM303D, IRS2181*PBF, TLP250, HCPL3120

El3ctr0nic:
3D Acc and Mag MEMS


diggit:
Nice symbols.
Just please do some tweaks...
- align optocoupler LED with PD in line and flip arrows? All opto symbols are in this style.
- datasheet links should point to manufacturer's file and not someone else private site.
(http://web.itu.edu.tr/yildiri1/mylibrary/data/tlp250.pdf)
- It is not symbol of this PR, but could you adapt LSM303DLHC to your style? It is highly similar to yours LSM303SD and your design is more logical.

Nothing else prevents merging this...

El3ctr0nic:
Hi, I made fixes some times . Pls check it ;)

diggit:
Optocouplers look good.
Why is IRS2181PBF so high?
LSM303D
- put SETC lower, against CS (SETC and SETP will still have 300mil space, which is enough for cap.)
- SETC, SETP should be set to **passive** to prevent unnecessary warnings about not driven pins
- put footprint field outside of symbol, you can use (I suggest) `your_path/checklib.py -vv  --enable-extra  -c LSM303D --fix sensors.lib` to do it for you (pull latest version from git)

El3ctr0nic:
Hi again ;) i made fixes using .py script and Your suggestions.
IRS is so high becouse we can direct add bootstrap and Vcc caps to pins (clearer schematic draw i think). If i go down with dimensions i can't hold symbol symetry and 100mils pin grid and have enough space to cap. Its possible but connection wires will be very short(50mils) and it look ugly ;)
If it is need i change high.
Regards,
Marcin

irs2181

diggit:
What about this shape?
irs It is significantly smaller and still usable with capacitors placed your way.
Sorry for weird text, probably broken KiCad build...

diggit:
Please remove PBF suffix from IRS2181* as it is meaningless for symbol.
Everything else is ok, fix it and we are ready to go.

diggit:
Nice work you did, merging. :)

PR #392, opened 2016-02-12
rotate pins 90° to match correct pin order

BenediktSeidl:
otherwise the new footprint will not match, see [pull request #8 of TO_SOT_Packages_THT.pretty](https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/8)


ricardocrudo:
It's already fixed on master. Thanks.

PR #393, opened 2016-02-13
added W25QXX and DS3231

fromeijn:
My first git pull request ever, please check careful ;)


Fat-Zer:
See the KLC: https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt

For DS3231 at least next mistakes:
1. Pins are misaligned
2. Origin should be in the centre
3. Footprint filters are wrong.
4. For inverted pins add tilde (~) before name

And suggestions:
Pin numbers maybe will look better with increased length (150 mill)
SCL pin name may look slightly better with increased offset.
May be it's possible to make the whole symbol a bit smaller.
Not critical, But text fields are poorly aligned.

[Not sure, but hidden pins may be required to be attached to the main body rather than just float around; may conflict with last suggestion]

hackscribble:
Hi @fromeijn 

Many thanks

PR #397, opened 2016-02-17
add symbols similar to EN60617, light, horn, switches

jonasstein:


reportingsjr:
test this please

Fat-Zer:
[Most of it is just my opinion]
The leads are too long (at least for the bulb). IMHO no reasons to make them longer than 100mils.
According to KLC field text should be 50mils.
I don't see reasons why default value fields should include some standard-reference numbers... They may be in the description if you insist on including them...

Why do you made a separate library? IMHO all this will look good in the generic device.lib... also I like your switches more than that currently contains in the library... may be in may even replace the current one...

Why the switches got designator "S" rather than SW?
Shouldn't the circle in switch_NO_circle be clean?
Is it really correct that bulb and horn have designator "E"? (I don't know)

jonasstein:
@Fat-Zer thank you for your comment on my pull request.
The length of the leads was the KiCad standard setting. Is there an official rule for the lead length?
I agree, that they should be shorter.

The KiCad standard setting was different for the text field and I changed it manually. I will see where I missed to change it to 50 mil.

The default number makes sense for a resistor, but it gets complicate for most symbols in EN60617. I thought the user may find the right symbol easier with the number and will most likely have to change it anyway. 

The representation of a device depends strictly on the standard. A resistor in standard A may look different from standard B. Additionally the user does not want to mix standards in one schematic. 
I think it is important to separate the standards clearly. In my opinion this is solved by separate folders as of today. Everybody can contribute to this folder, if the symbol agrees with EN60617. 
This is different from "devices" folder, where a symbol may look like a "potato". EN60617 describes a very  comprehensive collection (thousands) of symbols and is widely used. So it is worth to have these symbols in a separate folder.

S is the designator for a switch in DIN40719-2. 
May be its worth a feature request to have a designator "SWITCH" in the library and the user may select a standard for the document. KiCad will lookup "SWITCH" in the selected standard and print "S?" or "SW?" or "Q?"... 
"BULB" -> "E" or "L" or ....

About the circle: You are right, it should be cleared.

Fat-Zer:
>The length of the leads was the KiCad standard setting. Is there an official rule for the lead length?
I agree, that they should be shorter.

See the KLC... in several words rules are: align on 100mil grid; length is (100+N*50)mil with preference to shortest unless it doesn't make appearance worse.

> The default number makes sense for a resistor, but it gets complicate for most symbols in EN60617. I thought the user may find the right symbol easier with the number and will most likely have to change it anyway. 

IMHO it looks at least crappy... Also note that for search means it will work as good if it will go to either documentation or keyword field. In general I doubt that there are people who remember those numbers and will try to search by them, or even who will be interested in them.

> I will see where I missed to change it to 50 mil.

Everywhere but in switch_NC...

> The representation of a device depends strictly on the standard. A resistor in standard A may look different from standard B.

Yes, it's a generic problem... IMHO there is no sane way to resolve it in the current situation. To create a separate library with thousands of components for each standard sounds just silly to me... I hope it will be resolved than kicad will migrate to s-exp schematic format. It shoukd be quite easy to have different alternative images for the same component.... Until than IMHO we will have to choose some compromise for the best symbol look...

>S is the designator for a switch in DIN40719-2.
May be its worth a feature request to have a designator "SWITCH" in the library and the user may select a standard for the document. KiCad will lookup "SWITCH" in the selected standard and print 

IMHO it's worth to completely replace the current symbol... I don't really feel strong about S vs SW...

>"S?" or "SW?" or "Q?"...
"BULB" -> "E" or "L" or ....

S and L respectively sounds better for me... but I'm not a standard guru...

jonasstein:
I have fixed the discussed problems. But I am not sure, what to write instead of the standard ID in the VALUE field. I guess the user would overwrite the "08-10-01_Light" with "100 W HIQ" or "6V/10A PAR 36"... Any suggestions?

reportingsjr:
test this please

reportingsjr:
test this please

reportingsjr:
test this please

reportingsjr:
test this please

reportingsjr:
test this please

reportingsjr:
Sorry for the noise, needed to test out a small fix for the script.

I'm not sure what the script wants to fix rule 3.8, will look in to it tomorrow.

Fat-Zer:
>I have fixed the discussed problems. 

Looks good for me... About 3.2 violation for switch with circle: I'm not sure; IMHO that may be considered as generic device, but that's up to library maintainers to decide...
>But I am not sure, what to write instead of the standard ID in the VALUE field. I guess the user would overwrite the "08-10-01_Light" with "100 W HIQ" or "6V/10A PAR 36"... Any suggestions?

I'd suggest just generic "Light", "Horn" and "Switch". If a user will decide not to replace it with something meaningful it would look ok anyway...

>I'm not sure what the script wants to fix rule 3.8, will look in to it tomorrow.

It's because footprints are missing Documentation field, which doesn't really make scene here... IMHO better to improve the python or jenkin's script rather fix it hear...

jonasstein:
The representation of the value field in KiCad requires that it contains the symbol ID.
There are variants of symbols with the same device, but representation and ID differ.
I suggest to write in **F1** fields in EN60617.lib  `ID_DescriptiveName`. The user can find the symbols sorted in the library tree and can read the ID and a descriptive name.
I agree, that this is no perfect solution, because `07-02-01_switch_NO` is no desired value like `100 k`, but it is the best compromise for now as the user will have to change this field anyway. The user will have to change it to a useful value as it is done in case of an resistor. This would be in perfect agreement to
https://en.wikibooks.org/wiki/Kicad/file_formats#Description_of_F0_and_F1

If we stick to this rule, we can change it later easily by a script for all symbols in EN60617.lib as soon as we have a better solution.
Would you agree with this suggestion, so that we can merge and the library and let it grow?

Fat-Zer:
>If we stick to this rule, we can change it later easily by a script for all symbols in EN60617.lib as soon as we have a better solution.
Would you agree with this suggestion, so that we can merge and the library and let it grow?

I still disagree with that; but yet another time: luckily, I'm not a maintainer of the library and I don't think that my word is decisive anyway... Although no hard feelings if it will be merged as-is.

>I agree, that this is no perfect solution, because 07-02-01_switch_NO is no desired value like 100 k, but it is the best compromise for now as the user will have to change this field anyway. The user will have to change it to a useful value as it is done in case of an resistor.

The problem here is that user will __have to__ rename the component, but unlike resistors IMHO it's likely that user will be satisfied with a default descriptive name unless it will include some bogus number. I personally hate than I "have to" do something... =)

>The user can find the symbols sorted in the library tree and can read the ID and a descriptive name.

In general I still don't really see any scenario why anybody may be interested in those numbers at all or either in the library ordered by them... just this one silly case:

E.g. somebody wants to add a bulb into his schematic... he opens his EN60617 pdf (I hardly believe he remembers those numbers) ; scrolls a dozen of pages; than he finally founds a number; then browses through the library and finds the component after which he has to also rename the component...
In this case most likely a user will enter a "bulb" into search field and if he is interested in a standard he will be happily surprise that the bulb conforms to it... Ok, if he interested in checking whether it really as in the standard he may look up the symbol by it's number in the sheet. But there is no need to include it into the symbols name... Just put it into the description or the keywords...

So I suppose for most users (and defiantly for me) those numbers in names will be more annoying than useful.

jonasstein:
@Fat-Zer I understood your argument, but KiCad uses this value to list the symbols in the library. Try to rename "01-02-03_switch", "01-02-04_switch", "01-02-05_switch" all to "switch" and the user is not able to distinguish them in the library anymore. Please try it out, and you will see, what I mean. If I misunderstood you, please show me your solution in a fork and I will try the symbols out.

Fat-Zer:
>Try to rename "01-02-03_switch", "01-02-04_switch", "01-02-05_switch" 

Sure, but they are "switch_NO" "switch_NC" and "switch_NO_circle" already. =)
I'm ok with those.names... Also if you can't think out of a meaningful distinguish name for one of futher components you can add a 1 like in CP/CP1 in device.lib...

btw: another tiny issue: all symbol names start with lowercase letter but Light... it's inconsistency; either all should be capital or small.

jonasstein:
I still think the 01-02-03 is rather equivalent to 1N4001 1N4002... then to "Speaker", "Light"...

reportingsjr:
test this please

reportingsjr:
The 3.8 errors are because there is no datasheet link. That makes sense for these symbols so I suppose you can just ignore them.

Fat-Zer:
>The 3.8 errors are because there is no datasheet link. That makes sense for these symbols so I suppose you can just ignore them.

@reportingsjr, you are the  jenkins' maintainer, aren't you? why don't you add "-v" to the checklib.py's args so it would be more obvious in such cases?

reportingsjr:
Hmmm, I will look in to that, but it will require some changes to the code. I'll try to get that enabled in the next few days.

ricardocrudo:
@jonasstein we already have most, if not all, of those symbols in the device library. If you are missing a specific symbol please PR it there, in the device library.

PR #404, opened 2016-02-22
device: fix polarized capacitor footprint filter

Fat-Zer:


pointhi:
looks good. Would you also fix ```CP_Small``` and ```CP1_Small``` or should I do it?

Fat-Zer:
thanks... Missed them.

pointhi:
thanks

Issue #200, opened 2015-05-14
Some common schematics for OpAmps


Fat-Zer:
![34](https://cloud.githubusercontent.com/assets/1060057/7625957/dfc27708-fa07-11e4-8cce-bf67aebc99f5.png)

Now the library contains a handful of different OpAmps images, some of them are not conform to current conventions. I understand that most of them exist by historical reasons, but IMHO now it looks like a huge mess... 

I don't know what will be the best solution, but the most straightforward way I can think of is the slightly update the library format, some my ideas on it was listed [on the maillist on january](https://lists.launchpad.net/kicad-developers/msg16325.html). On the other hand this is the most tough way and it can break compatibility of older versions of kicad with newer libraries.

So I at least propose a minimum solution: to choose a "best" OpAmp image that conforms to the current conventions (may be include it as some generic device) and recommend to use it for new devices and may be update some older devices to new image.

How does it sounds for you?

ricardocrudo:
I think would be great have, always that possible, the same shape. But, I like of the idea to make it at once, and to minimize the troubles that this could raise, make an announcement to the users, as was done to diodes A/K change, could be a way. In this, any symbol that someone pick to start a new OpAmp would be OK. In short, I'm in favor to make the KiCad library KLC compliant as fast as possible.

CarlPoirier:
I vote for uniforming this as well.

Fat-Zer:
![42](https://cloud.githubusercontent.com/assets/1060057/7753163/4b2568ae-ffec-11e4-81a0-869c111c919d.png)
Here is schematic I propose. can be vied here: https://github.com/Fat-Zer/kicad-opAmp-test
Please review it before I start convert the library (unless there will be some other volunteers).
 - OpAmp — a general OpAmp. (pin numbers are random)
 - Comp — a single power only OpAmp with V- as Gnd.
 - OpAmp-x2 — a dual opAmp
 - OpAmp-x2-v2 — a second variant of dual OpAmp, I like it more. if there will be no objections I'll use it.
 - LM158 — a preview of real schematic of a dual OpAmp (nothing special in that case)
 - CA3130 ­— a preview of another real opAmp with additional outputs.
 - CA3130-v2 ­— the second version of the same. pros: have place for output names. cons: looks a bit worser IMO.

So... I'm looking forward to hear your opinion at least on the complicated opAmps schematic...

ricardocrudo:
I like them, except CA3130-v2.

[...]

PR #405, opened 2016-02-22
Fixes to the OpAmp schematic

Fat-Zer:
This is proposed changes against issue #200.


ricardocrudo:
Nice!

Fat-Zer:
>can you please check the conflicts?

Done.

diggit:
Symbols look really nice. Can you fix documentation of their aliases? It should be peace of cake for you...

```
checking alias: LM101H
      missing datasheet, please provide a datasheet link if it isn't a generic component
checking alias: NE5532
checking alias: 4558
checking alias: ADA4075
      missing keywords
checking component: LM733CH
      missing datasheet, please provide a datasheet link if it isn't a generic component
```
*(unneeded info removed from script output)*
*alias documentation checking was added to kicad-library-utils lately, pull changes from its repo unless you have already done*

diggit:
Somehow, alias LM101H got back. As it is not valid alias to CA3130, please remove it. Issue #406

Fat-Zer:
> Symbols look really nice. Can you fix documentation of their aliases?

Done.
Note about LM733CH: it haven't found a datasheet link to some official manufacturer's site, so the link leads to some hobbiest's database.

> Somehow, alias LM101H got back. As it is not valid alias to CA3130, please remove it. Issue #406

I haven't delete it before due to I considered that I may be wrong about that...
And after quite a while it is now obvious that nobody knows about that and it's very likely it's just wrong...

diggit:
I can't even find datasheet for "LM101H". It is ok to delete it.
In closer look eg. AD813 has really close + and V+, same for negative. It is really uncomfortable to read. Any ideas how to solve it?
Maybe put pin names outside and place text fields with +/- inside for inputs?

Fat-Zer:
>In closer look eg. AD813 has really close + and V+, same for negative. It is really uncomfortable to read. Any ideas how to solve it?

It's the same for nearly all OpAmps... The only thing can be done is to reduce the field offset to 10 or even 5 mils, but I'd rather done it in a separate pull request, because this one is hanging around for quite a while already...

>Maybe put pin names outside and place text fields with +/- inside for inputs?

Hmm... I don't really like that... IMHO the pin names is an information that doesn't want to be loosed, especially +/- v+/v- pins due to that may be helpful for the upcoming migration to new schematic format (at least I'm one of guys who still hope it is still upcoming...)

timofonic:
Will it happen for 5.0? Is going to have the symbols embedded in the file too? What about including the PCB? Can it be done in a way friendly with version control like Git? 
It's nice to have hope abs there are reasons for it. Other projects are in a much worse situation, with lots of forks and not the great improvements KiCad is having these days. 

On March 22, 2016 9:30:01 PM CET, Golubev Alexander <notifications@github.com> wrote:
>>In closer look eg. AD813 has really close + and V+, same for negative.
>It is really uncomfortable to read. Any ideas how to solve it?
>
>It's the same for nearly all OpAmps... The only thing can be done is to
>reduce the field offset to 10 or even 5 mils, but I'd rather done it in
>a separate pull request, because this one is hanging around for quite a
>while already...
>
>>Maybe put pin names outside and place text fields with +/- inside for
>inputs?
>
>Hmm... I don't really like that... IMHO the pin names is an information
>that doesn't want to be loosed, especially +/- v+/v- pins due to that
>may be helpful for the upcoming migration to new schematic format (at
>least I'm one of guys who still hope it is still upcoming...)
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/KiCad/kicad-library/pull/405#issuecomment-200007313

-- 
Enviado desde mi dispositivo Android con K-9 Mail. Por favor disculpa mi brevedad.

Fat-Zer:
> Will it happen for 5.0? Is going to have the symbols embedded in the file too? What about including the PCB? Can it be done in a way friendly with version control like Git?

I don't know a lot as well, just some fragmentary (and may be particularly outdated) info from the maillist... The new format is on the road map for v5, so I hope it will be finished till it. But as I can see there haven't been done a lot since some drafts of the format was published... The format itself is S-Expression like the one used for pcbs, so pros and cons are mostly the same...

diggit:
Ok then.
Please solve conflicts, I'll merge your changes.
Sorry for late reply, PR got lost on queue.

Fat-Zer:
rebased on current master...

PR #407, opened 2016-02-23
Yet another footprint filter enhancement

### LINKED ISSUE #402: Footprint filter entries now broken for LEDs
Changing the "-" to "_" (dash to underscore) as the naming convention in  [ https://github.com/KiCad/LEDs.pretty/pull/10] (https://github.com/KiCad/LEDs.pretty/pull/10)
without changing the Footprint filter entries in the Symbol Editor properties makes it that Cvpcb will not not find SMD footprints when Filtering on Keywords. It has, for example, Filter keywords like LED-0805 and LED-1206, while the components are now LED_0805 and LED_1206.

I'm a real newbie (i found this when going thru the Getting to Blinky video tutorial series from ContextualElectronics) - so while I was able to change the Footprint Filter locally on my computer using Symbol Editor (an answer i got on the kicad.info forum), I'm not at all confident enough on how to do this to submit back to the Github libraries.

Fat-Zer:
This also resolves #402


pointhi:
please don't forget https://github.com/KiCad/Air_Coils_SML_NEOSID.pretty. So adding the Keyword ```Coil``` would be a good idea.

The rest looks fine

Fat-Zer:
Thanks for the notice... done

pointhi:
thanks for your contribution

Issue #403, opened 2016-02-22
3D model for a 7mm resistor


Fat-Zer:
Currently there is no model for a 7mm resistor in the library. IMHO this is a quite major loss...

If nobody is going to fix it in a couple of days, I'm going to pull-request one, but I will make it with FreeCAD rather than wings.

Any comments / any good ready-to-use models / other volunteers?

Fat-Zer:
Ok, I've done a model... I was a bit confused what resistor is suitable for the "Resistors_ThroughHole:Resistor_Horizontal_RM7mm" footprint. I've decided to use dimensions for SFR16S from this datasheet: http://www.vishay.com/docs/28722/sfr16s25.pdf.

![resistor-7mm](https://cloud.githubusercontent.com/assets/1060057/13261898/53a78770-da73-11e5-855b-438393f1de3e.png)

It differs from other resistors, by obvious reasons, but IMHO It looks a bit better.

Shackmeister:
@Fat-Zer hi, im case this is not relevant any more could you please consider closing this issue? :)

Fat-Zer:
Closing due to outdated as mentioned on https://github.com/KiCad/kicad-library/pull/408.

PR #408, opened 2016-02-23
3d: added a model for the Resistor_Horizontal_RM7mm

Fat-Zer:
see also https://github.com/KiCad/kicad-library/issues/403


ricardocrudo:
Hello @Fat-Zer can you please adjust the 3d params to match with the footprint? thanks.

Fat-Zer:
Hi, what do you mean?
The footprint's 3d parameters are adjusted in the PR mentioned above.

diggit:
There are 2 problems:
- Kicad expects, that 3D model dimension units are in **0.1 inches** insteal of **mm** -> you have to scale it by 1/2.54 (approx 0.394) in all axis in Footprint 3D model assignmnet. (many models are scaled this way)
- THT footprints have anchor on pin 1 ([KLC 6.3](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt#L66)), but it seems, that your 3D model has origin in center

**Default**


**Scaled and shifted**
Shifting should be fixed in 3D model itself.

Fat-Zer:
@diggit, everything is done the neighbouring PR: https://github.com/KiCad/Resistors_ThroughHole.pretty/pull/8

>Shifting should be fixed in 3D model itself.

hm... why?

diggit:
Because shift is specific for each device. Eg if sb. creates generator for these 3D models, nobody wants to edit each part. It is easier to rely on fact, that pin 1 is on origin.

hackscribble:
Hi @Fat-Zer 

Sorry for the long time to get back to you about this PR.

The components in `Resistors_THT.pretty` have been renamed and matching 3D models provided.  So it looks like this contribution has been superseded and, therefore, I am closing the PR.

PR #410, opened 2016-02-25
new element PIC32MX270F-SPDIP-28

PiotrCzCc:


PiotrCzCc:
OK, few pins was uncorrected.

PR #414, opened 2016-02-28
Fix MSP430G2230 symbol

welterde:
Had the wrong pin number mapping and footprint.


Fat-Zer:
Just wondering, why you changed the datasheet link?
Note that it redirects to the same file as before...

welterde:
Indeed. So no real reason other than perhaps it looks a teeny bit nicer.
But I can remove that change again if wanted.

Fat-Zer:
I've only noticed you replaced "www" with "focus" subdomain... I supposed that was intentionally so I wondered if there goes some magic out-there on ti.com's side... nevermind...

ricardocrudo:
Thank you.

PR #430, opened 2016-03-02
add TUSB2036 2/3 port USB1.1 hub controller

lorem-ipsum:


diggit:
Layout looks great, but there are differences between pin types you set and datasheet.

Datasheet...
- EECLK is **tristate**
- ~EXTMEM is **input**
- all ~OVRCUR pins should be **input**s according to description in datasheet (but in I/O type column is output, typo in datasheet?)

You have swapped pins NPINT0 and NPINT1
Please check all pins again.

It is recommended to use 10mil line thickness for symbol body.

Which version of KiCad do you use? There is missing footprint field in library (even empty one would be ok).

lorem-ipsum:
Thank you for the feedback. You were right, the pins were wrong. I went mostly by the table and the given IO type in that, as it turns out that is wrong/misleading in some places. I now checked every pin again and I think now everything should be correct.

I also changed the symbol thikness and added the missing footprint fields. I removed those manually, because they caused the KLC check to fail with a font size of 60 mils and couldn't find them in the editor. Is there any way to tell the editor to use a 50 mils font size for everything by default?

diggit:
You can edit all fields properties by clicking **T** icon right next to *Edit component properties*

Unfortunately, I don't know how to change default field font size (and I want to). You can [report it as bug](https://bugs.launchpad.net/kicad) as defaults should not collide with rules.

diggit:
There are some EC03 violations, nothing critical, it just helps to keep symbol appearance unified. It looks ok in this case.
Nice work you did..., merging.

PR #431, opened 2016-03-03
rebuilt shematic and fix librairie (Arduino template)

pouic13:
Hi all, 
I rebuilt all templates schematics with news components of new libraries schematic include in Kicad.
All Components are in harmony with template and grid 50.00mils.  
I remove the old library "special" and i clean up files. 
Net list are regenerate and pcb view are updated.


ricardocrudo:
Great!

PR #438, opened 2016-03-09
TI SN74F163A 4-Bit Binary Counter

lorem-ipsum:
This adds the SN74F163A to the texas library.

I have some questions regarding that schematic:
- I've added all the different packaging and shipping versions as aliases. Is that intended or is the main part enough?
- I've added every possible footprint to the component. However the full names listed in the aliases specify the footprint exactly. Should change them so that every set of footprints has it's own part or is that the intended way?

For the SSOP package no footprint exists yet because it uses a larger pitch


diggit:
Generic 74xx components does not suit you?

- package types are filtered just by type like SO\*, MSOP\*, and so. Pin count filter then finds matching package. It is handy for \*_handsoldering type of footprints and similar cases. Choosing narrow or wide package i up to user.

You can get clue, how it is done, by looking onto other components.

lorem-ipsum:
Generic 74XX does the job, I'm just blind. Thanks for the explanation anyway.

PR #439, opened 2016-03-09
Power management

SchrodingersGat:
I have started to port some of my power-management components (the ones that meet KLC at least) over to power_management.lib

I have also created power_management.dcm as it seems to be missing

I'll try to get all components in power_management to KLC but for now I'm just adding my own


SchrodingersGat:
@ricardocrudo  @CarlPoirier  any input on this? :)

PR #445, opened 2016-03-15
add 28F400 to memory.lib

pavelkokorin:
Intel PA28F400BX Flash EEProm 4-MBIT (256K X 16, 512K X 8) (PSOP 44 pack.) 5V, 12V Prog


diggit:
Please comply with rules.
```
checking component: 28F400
  Violating Rule 3.1
  Violating Rule 3.6
```
Look [here](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt)

You can check it locally with more info about violations using [https://github.com/KiCad/kicad-library-utils](kicad-library-utils) with -vv parameter

diggit:
In addition
- use 0.010 outline thickness
- set DU pin as invisible as it it DNC
- swap order of pin functions for pin 31

pavelkokorin:
all the rule vioaltions are corrected.

I'm not sure i got your suggestion correct
> swap order of pin functions for pin 31

I'v changed pin name that way "D15/A-1" -> "A-1/D15"

diggit:
Good, few other things
- your symbol is matching 28F400 in PSOP 44 only ->
 - remove wrong alias
 - set footptint filter to PSOP\* only
 - fill footprint field, PSOP-44 is ok
- fill rectangle with background color

I am afraid PSOP package does not exist in KiCad's libraries. Do you plan to draw it?

pavelkokorin:
Done.

I've already created PR to add PSOP-44 package https://github.com/KiCad/Housings_SSOP.pretty/pull/13

diggit:
Ok.
- Whenever possible, use manufacturer's datasheet. This one looks good. http://download.intel.com/design/archives/flash/docs/29045101.pdf

 (google - 28f400 filetype.pdf)

- Don't place footprint field inside of symbol. You can try `your_path/checklib.py -vv  --enable-extra  -c 28F400 memory.lib` from latest kicad-library-utils to solve it for you. It should find sane position.

pavelkokorin:
> checklib.py -vv --enable-extra

good to know, thanks

diggit:
I forgot to mention, there is --fix parameter to do this job for you, hope you found it.
There is [KLC 3.5](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt#L36). CE, OE,... are all inputs, so put them on left. 
This will involve symbols height raise (don't forget to center symbol afterwards, watch that pin 100mil grid, use checklib.py to test it).

pavelkokorin:
Ok, it's done.

diggit:
Nice work you did, merging.

PR #339, opened 2015-12-05
Added LPC812M101FD20 to nxp_armmcu

michalkona:

PR #448, opened 2016-03-16
Added LPC812M101JD20 to nxp_armmcu.

michalkona:
This pull request based on #339.


diggit:
This symbol is for SOIC-20 package only. 
- put SOIC-20 into description instead of "SOIC package".
- fill footprint field with SOIC-20 (don't leave this field inside of symbol, you can use `some_path/checklib.py -vv  --enable-extra  -c LPC812M101JD20 --fix nxp_armmcu.lib` for recommended fields position, but you'll have to set pin 18 back to bidir.)
- footprint filter SOIC\*

michalkona:
Done

diggit:
Nice, just change VDDCMP type to BiDir, VDDCMP is reference for comparator and it is covered by input type well. So BiDir is correct, it can be configured as output too. Fix this and we are ready for merge.

michalkona:
I think all is done.

diggit:
Nice work you did, merging.

PR #450, opened 2016-03-22
IRF7404 improved.

mpkopec:
KLC compliance corrected. Layout changed to be more mosfet-like. Added
docs and footprint filter.

Large number of +- in .lib is probably just git diff misunderstanding as I have seen in gitk.


diggit:
Nice improvement.
nothing is perfect on first try, even second...
- add body diode
- correct symbol, note the difference between enhancement and depletion FET

You can get inspiration from symbol CSD17578Q5A for transistor symbol part and for description and keywords. (beware, CSD... is N-MOS)

@ricardocrudo Is it necessary to have individually accessible Drain and Source pins? They could be placed to overlap each other and have only 1 connection point. Like some regulators with multiple grounds...

These transistor need some unification.
IDK which way to go:
- like CSD* transistors - package matching layout
- like this PR
- simpler symbol with 3 connection points, D and S pins overlapping to connect them at once

hackscribble:
Hi @mpkopec 

Many thanks

mpkopec:
I will do this at some, point but probably not for this component ;)

PR #451, opened 2016-03-24
Add MAX13432E

SchrodingersGat:
https://datasheets.maximintegrated.com/en/ds/MAX13430E-MAX13433E.pdf
Alias MAX132433E


diggit:
You have some typos here. MAX13433E and not MAX13**2**433E.
Nice drawing, just:
- move top edge of rectangle 50mils higher to improve readability of top pin names 

FP aliases...
- DFN - you must match specific pitch and dimensions, so use **DFN-14-1EP_3x3mm_Pitch0.4mm***
- SO\* is ok

**Unfortunately none of those FPs exist in kicad yet.**
(this does not prevent merging, but usage...)

SchrodingersGat:
Hey @diggit  I have made these changes - please let me know if there's anything else I can do.

diggit:
Please save library with kicad build 6331 or higher, there were changes in library formatting. It will simplify diff.

SchrodingersGat:
@diggit I've re-saved it with the newer version

diggit:
Thanks!
After second look into datasheet, SO14 version does not have exposed pad. You created symbol which is specific for DFN package:
- fill footprint field with "DFN-14-1EP" (there is nice place for it at bottom side next to EP pin)
- remove FP filter SOIC\*

SchrodingersGat:
@diggit thanks, good spot :) I have now created two separate parts for the DFN and SOIC version (with the correct suffix as per the datasheet). Also assigned footprints and such.

LMK if there's anything else!

diggit:
Symbols are great!
Please fix these:
- clear Datasheet field - it is common to component and all its aliases, thus colliding
- for DFN version use `DFN-14-1EP_3x4mm_Pitch0.5mm*` FP filter, because you have to match proper dimensions and pitch

Now I realized, I made mistake. FP fields should contain library name and footprint name. At least KiCAD says. So use whole name for SO package (Housings_SOIC:SOIC-14_3.9x8.7mm_Pitch1.27mm) and leave blank for DFN.
In this case, FP filter for SO version is useless, but none complains, leave ti there...
*I think, I told you not to put library name there earlier. My fault, sorry!*

diggit:
Oh, all aliases have same datasheet. Still, put link in documentation and leave Datasheet field blank. Otherwise, it may conflict in future.

diggit:
@ricardocrudo @CarlPoirier Is there any guideline when use FootPrint field and when FootPrint filter for single package component?
- When you fill FP field, kicad automatically assigns footprint. You have to fill it with exact library name and FP name, otherwise you get warning in kiCAD.
- When you set FP filter, KiCAD lets you choose from all matching FPs, so _Handsoldering, _Largepads and similar suffixes works too.

Deducing from this, it makes sense to use FP filter only. What do you think?

SchrodingersGat:
@diggit  i have removed the explicit datasheet and footprint fields

diggit:
I realized, that it comes from its purpose.
FP filter - is used to filter all allowable footprint types (not defining pin count)
FP field - carries assigned FP. You set some sane default which user mostly don't need to change. (for single FP components). When you assign it different FP, FP field changes.
 - the only correct format is Library:FootPrint.

 In such format, it takes a lot of space in schematic, so setting it invisible makes sense.

So here it is:
-FP field: *Housings_SOIC:SOIC-14_3.9x8.7mm_Pitch1.27mm*
-FP filter: *SOIC*\*

SchrodingersGat:
Yup, getting a handle on it now :) fixed.

diggit:
FP filter for DFN is matching assigned FP
*Housings_DFN_QFN:DFN-14-1EP_**3**x4mm_Pitch0.5mm*
,but wrong FP
*Housings_DFN_QFN:DFN-14-1EP_***4***x4mm_Pitch0.5mm* too.

Extend FP filter to sth like this:
DFN-\*_3x4mm_Pitch0.5mm\*

SchrodingersGat:
Understood - fixed

diggit:
even that trailing \*

SchrodingersGat:
Added *

diggit:
And more adventure awaits! But not in this PR.
Nice work you did, merging.

PR #453, opened 2016-03-27
Added parts of the cyclone V chip familty of Altera

Groonie:
Some of the Altera Cyclone V ic's have been added. Created using the pin-out files.


Groonie:
Updated the text sizes and documentation, hopefully it will pass now.

CarlPoirier:
Hi @Groonie,

That's nice work but I think there is room for improvement on their size. I suggest you separate them into units using the groups you used already.

Regards,

Carl

Groonie:
The individual chips are now split into units, some are still quite large. Bet let me know what you think.
kicad_cyclonev_parts

Groonie:
It looks like this now
altera_banks_cv

diggit:
Please put all non PIN fields outside of symbols. Some symbols have overlapping texts due to it. Then you can shrink symbols even more.

Groonie:
What exactly do you mean by putting the PIN fields outside of the symbols? Normally
the pin descriptions (green text) is always inside the symbol. And the pin number outside (red text).
The problem is maybe there is a label in the center, But all banks in a package always have the same label position.
And since the banks all have different sizes, it's difficult to find a correct spot. Therefore I have put it in the middle. Take a look, I hope it's not that bad.

hackscribble:

hackscribble:
Hi @Groonie 

Many thanks

PR #463, opened 2016-04-03
Add Microchip TC1262-33VDB voltage regulator

swalense:
Add Microchip TC1262 500mA Fixed Output CMOS LDO voltage regulator with SOT-223 package.
Datasheet at http://ww1.microchip.com/downloads/en/DeviceDoc/21373C.pdf


diggit:
Symbols need some tweaks:
- move GND pins on bottom edge and place them to overlap each other (user will see 1 GND pin, but he will connect them all by 1 wire)
- move FP field outside of symbol
 - you can try checklib.py from kicad-library-utils to fix if for you
 ` your_path_to/checklib.py -vv  --enable-extra  -c TC1262-33VDB --fix regul.lib`
- change FP filter to SOT-223\* , reasons:
 - library name may change
 - \* se want match something like SOT-223_Handsoldering too
- this regulator has more variants in SOT-223 with different output voltages, could you alias them and fill their descriptions too? (use "Copy Document from parent" and just modify output voltage)

BTW Some other SOT-223 regulators have missing pin 4 GND. Your symbol may be used as correct template in future.

swalense:
Thanks for the comments; they should all be addressed.

diggit:
Congratulations to your first PR!
Nice work you did, merging.

PR #471, opened 2016-04-11
KLC and description for LM35

SchrodingersGat:


SchrodingersGat:
Ignore this. I will re-issue this PR with the new temperature-sensor symbol artwork

PR #475, opened 2016-04-12
Added A110xxLHL series of hall-effect switches

SchrodingersGat:
Ref: http://www.allegromicro.com/en/Products/Part_Numbers/1101/1101.pdf


diggit:
Symbol **A1301KLHLT-T** looks same. Use existing symbol instead of duplicating it.

Keyword *sensor* is redundant for sensor library, please remove it.

SchrodingersGat:
@diggit  the A1301 (and its ALIAS parts) are in a different footprint, should I add them to the A110xx symbol as ALIASes?

If I do that, I'll have to add a new fp-filter to the symbol, even though the A110xx parts don't match the new fp-filter. Is that ok?

diggit:
Ok, I didn't check FP filters. Leave it as it is. (I hope, KiCAD's library format will improve.)
- fill FP field with full format name (Library:FootPrint)
- solve conflicts

SchrodingersGat:
@diggit  I have solved the conflicts (they arose from recent merges in KiCAD master branch)
I have also added FP association as requested.

diggit:
Thanks!
Nice work you did, merging.

PR #476, opened 2016-04-12
Added TPS62177

SchrodingersGat:
ALIAS TPS62175

Reference: http://www.ti.com/lit/ds/symlink/tps62177.pdf


diggit:
- keywords are space separated, use sth. like: Step-Down DC-DC Converter
- do not use Datasheet field to store datasheet link, put it in documentation
- use direct links with pdf suffix to minimize redirections (http://www.ti.com/lit/ds/symlink/tps62177.pdf)
- change SW pin to power output
- change NC pin to passive type, it can be left floating
- use upper case letters for Vin, En, ~Sleep as in datasheet
- use same symbol for both components to make it easily replaceable

SchrodingersGat:
@diggit  I think I got them all

diggit:
Stay consistent, use keyword *step-down* instead of *stepdown*
Fix it and we are ready to merge.

SchrodingersGat:
Done!

diggit:
Umm, I should never promise immediate merge after fix.
Move "with Sleep Mode" at the end of description to share description format with others. (devices
 tps62175 and tps62177 )

SchrodingersGat:
Doneski

diggit:
Great, thanks! Bored huh?
Nice work you did, merging.

PR #479, opened 2016-04-12
Added LM1117 series of linear regulators

SchrodingersGat:
http://www.ti.com/lit/ds/symlink/lm1117.pdf


diggit:
- add suffix \* to FP filters

**TO-263, SOT-223**
- set TAB pin invisible an place it over VOUT pin
- unify your design with some already existing *1117* LDO, there are multiple, choose some (I hope, at least some of them comply with KLC)

**TO-220, TO-252**
- these are 3 pin packages, you can't use same symbol for them.

*A lot of existing symbols does not comply with KLC, they have incorrectly formatted: FP filters, FP field, keywords, etc... Avoid copying these errors*

SchrodingersGat:
@diggit  is this what you mean regarding putting the TAB pin over the Vout pin?

I have unified my design with AP1117

diggit:
Good job with unification.
- remove TAB pin name (for AP117 too), it is not necessary and decreases readability in symbol placer
- nice catch with that AP1117, fix few other things as you already touched it
 - set filter to SOT-223\* only (single FP component) ->
 - set FP field accordingly

SchrodingersGat:
@diggit I have made these changes and also improved keywords for the AP1117x

diggit:
Thanks!
Nice work you did, merging.

PR #480, opened 2016-04-12
Added LM3480 series of ldo

SchrodingersGat:
Ref http://www.ti.com/lit/ds/symlink/lm3480.pdf


diggit:
Great, another SOT-23 LDO pinout :D

Description
- *Quasi* Low Drop....
- add *100mA* *positive* *fixed* into keywords
- FP field should be left empty of filled in full name format Library:FP (so long, that it should be set invisible)

SchrodingersGat:
Done

diggit:
\+ *fixed* *positive* to keywords

SchrodingersGat:
Yup

diggit:
Nice work you did, merging.

PR #481, opened 2016-04-13
STM32: Fixed missing power pads in some parts.

ObKo:
Some parts (with -QFN- packages) in auto-generated STM32 library has no power pads. That happened because pin files from ST not always declares power pad properly.

Fix for script: https://github.com/KiCad/kicad-library-utils/pull/26


SchrodingersGat:
thanks @ObKo :) great work!

diggit:
@ObKo  Thanks for fixing
I took a look on those MCUs. FP fields are in incorrect format (should be Library:FootPrint, invisible) and FP filters are not set at all. Can you improve mentioned script to set it correctly? (in separate PR)

Nice work you did, merging.

PR #482, opened 2016-04-13
Added AMS magnetic rotation sensors

SchrodingersGat:
AMS5048A
AMS5048B
AMS5050A
AMS5055A

Ref: http://ams.com/eng/Products/Magnetic-Position-Sensors


diggit:
- pin length 150mil is enough
- set all TEST pins to passive type
- use same name and case for pin names as in datasheet: Test -> TEST, SCK -> CLK
- I/O interface on left, PWM out on right

AS5055 seem unfinished to me. There is completely missing I/O interface. Missing pad 17 for exposed pad (yes, it will be NC and invisible, but it is necessary for correct FP assignment).

Please check you symbols before opening PR. You'll help to quick merging.

**Quantity ≠ Quality**

SchrodingersGat:
@diggit that's such a weird thing, I must have accidentally deleted those pins? Whoops, sorry about that. I have fixed all the pin names too. Please LMK if there's anything else I can do to improve this - I think that it should be good to go, though.

diggit:
Just add I2C/SPI keywords accordingly and we should be ready to merge.

SchrodingersGat:
^

diggit:
Thanks!
Nice work you did, merging.

PR #484, opened 2016-04-13
Added AD7171 16-bit ADC

SchrodingersGat:
http://www.analog.com/media/en/technical-documentation/data-sheets/AD7171.pdf

Application: kicad
Version: (2016-04-10 BZR 6684, Git 4d77388)-product, release build


diggit:
Any special reason to setting pin type *passive* instead of *input* for Vref* and Ain* pins?

If you shift VDD more left, you can put whole component name above symbol or are you matching layout with another symbol?

Please match symbol names and case with datasheet. (SCK -> SCLK, ~RST -> ~PDRST (if you shift VDD, there will be enough room for longer pin name).

*Keep in mind, that everything is opened for discussion. These are not orders. (maybe except KLC)*

SchrodingersGat:
Orders complete ;)

keruseykaryu:
@diggit: "Any special reason to setting pin type passive instead of input for Vref* and Ain* pins?"

IMO. Because they are analog inputs not digital inputs and this setting is safer to ERC.

diggit:
They are analog, but they need to be driven. On the other hand, eg. voltage divider is common source for Vref and ERC might not be happy for input type, hmmm.
It makes sense to change them to passive type. Several ADCs have set them as inputs. Open for fixing... 
@keruseykaryu your opinion is welcome.

@SchrodingersGat please set Ains and Vrefs back to passive. (revert 9e34593c384a35f127223d8be7b4e95615856e9a)

diggit:
Don't hesitate to ping me by mentioning me when you do changes. Notification are send when new message appears or sb. mentions me.
Nice work you did, merging.

PR #485, opened 2016-04-13
Added Texas Instruments TS5A3159 and TS5A3160 switches

swalense:
The TA5A3160 contains the symbol, the TA5A3159 is an alias. Both are SPDT switches with similar characteristics, only differing on the break-before-make / make-before-break ability.

http://www.ti.com/lit/ds/symlink/ts5a3159.pdf
http://www.ti.com/lit/ds/symlink/ts5a3160.pdf

Note,
- The symbol in the pull request only shows the pins. I have a version showing the switch as presented in the datasheet (see picture); however it does not appear particularly useful in a schematics... Please let me know if you believe it's better to use that version.
- The symbol showing the switch is not compliant with the rules; however I would update it before resubmission if needed.

screen-capture-3


diggit:
Minimal design is ok. What do you think about putting GND/V+ on bottom/top edge? If you don't think it would be useful, put GND/V+ on bottom/top line of pins at least.

change NC, NO, COM to passive - it should not raise warning if you connect some of them to power output pin type.

diggit:
and remove *Texas Instruments* from keywords:
 1. these are 2 separate keywords
 2. obvious fact for texas library (filter in "place component" is considering library names too - try it...)

swalense:
Thanks for the review, I have included all the changes. The pins on top/bottom make the symbol slightly bigger but I don't think it's an issue.

diggit:
Ok then.
Nice work you did, merging.

PR #486, opened 2016-04-13
Added ON Semiconductor CM1213A-01SO

swalense:
The symbol is for the single channel, SOT-23 variant of the ON Semiconductor CM1213A ESD protection diode.

http://www.onsemi.com/pub_link/Collateral/CM1213A-D.PDF


diggit:
- use SOT-23\* FP filter
- remove redundant circle(dot) at pin 1 (there are 2 overlapping)
- set fill to FG for junction circles to be dots
- set FP to full name (Library:FP)

diggit:
Thanks!
Next time, write some message when you are done to notify us.
Nice work you did, merging.

PR #487, opened 2016-04-13
Added Maxim Integrated DS3231 RTC

swalense:
Note that the footprint on the symbol (field F2) is set to SOIC-16-W to indicate that it's the wide version. There is no such label in the footprint library, which causes a warning when Kicad attempts to do the association automatically. However the footprint filter is set to the value found in the library (SOIC-16_7.5x10.3mm_Pitch1.27mm).

Please let me know if the F2 footprint field should be changed to SOIC-16_7.5x10.3mm_Pitch1.27mm.

https://datasheets.maximintegrated.com/en/ds/DS3231.pdf


diggit:
- mention pkg (SO-16) in description
- add *I2C* keyword
- FP field, should be empty or in format Library:FP (invisible)
- FP filter - add \* suffix to match names like SOIC-16_7.5x10.3mm_Pitch1.27mm**_Handsoldering**
- *SDA* is bidir pin type
- *~RST* is open collector pin type (it is input/output and output is open collector)
- change shape to height>width and center it, sth like this:

swalense:
All suggested changes have been applied.
Note,
- I've added SOIC-16 instead of SO-16 for consistency with some other parts in the library,
- The RST pin is open-drain but there is an internal pull-up... Don't if that should be reflected in some way in the symbol?

diggit:
This component is single FP, so FP field should be filled.

Looking into ERC matrix now, ~RST pin should be set to bidir.

In future, all dedicated SDA pins should be changed to open collector type. There are several thing to change...

swalense:
Footprint field added. I've also replaced the number of pins in the filter by a * following the discussion on the TLC555.

Type for RST and SDA changed. Also looking at the ERC matrix the latter indeed makes more sense.

diggit:
Sorry, I confused you. Leave SDA as bidir for now.
I was thinking aloud about open collector for SDA. It has to be discussed and then changed at once.

swalense:
Done. I thought you were suggesting to start with this part. ;-)

Note, the change could apply to SCL too for devices that can act as bus master...

diggit:
You are right about SCL, but I can't remember any device with dedicated i<sup>2</sup>c *master* pins. Most of i<sup>2</sup>c master devices are multipurpose pins on MCUs a thus bidir pin types.

Nice work you did, merging.

PR #261, opened 2015-07-27
Added ADM101E (miniature RS232 transceiver)

SchrodingersGat:
uSOIC-10 package
http://www.analog.com/en/products/interface-isolation/rs-232-rs-422-rs-485/interface-rs-232/adm101e.html#product-overview
http://www.analog.com/media/en/technical-documentation/data-sheets/ADM101E.pdf

PR #488, opened 2016-04-14
Can chips

SchrodingersGat:
Added multiple CAN transceivers to interface.lib

Removed from https://github.com/KiCad/kicad-library/pull/261 as it was getting too big.


diggit:
ADM3053
- Vref is *output*
- reorganize TxD, RxD, CANL, CANH pins to match other symbols you did (with respect to origin)

others/all
- FP filter - use filters without lead count if not necessary, it is filtered by pin count filtering (# filter icon in KiCAD)
- place SPLIT on right side (below CANL?), it is mid of terminators
- FP field should be empty or in Library:FP format (most of these cases). Invisible, as it is quite long text.

I like the way, you did these symbols. :)

SchrodingersGat:
@diggit I have made all these fixes (pretty sure). I see what you mean now regarding the way to format the FP filtering. I'll do it this way for all future PRs. Cheers!

diggit:
I am probably nit-picker, but ADM3056 has RxD and TxD pins differently than rest of CAN drivers.
(Just move RxD above Txd and we should be ready to go.)

BTW filters in KiCAD are quite powerful, regexps FTW! You can enter *adum?160* to match adum3160 and adum4160, or *mcp25*\**-h* to match all those CAN drivers with high temperature grade sign and even more...

SchrodingersGat:
@diggit I have changed all the CAN chips to match the ADM3056 as I think it is a better layout. I have now also added default footprints for all chips that only have one package.

Regarding the regex comment, I am not quite sure what you mean here. I know that when you are searching for a part to add, you can type e.g. "MCP25?1" and it will return all the MCP2551x and MCP2561x parts. But this is in the "add component" dialog in EESchema. Do you mean that you can construct the name of the actual symbols using regex? I'm just a tad confused as to how this applies here :)

diggit:
Thanks.

I mentioned regexes, because I didn't know about this functionality in KiCAD and it really helps to filter symbols from PRs. Nothing more, sorry. :)

**ADM3053**
- use two dashed lines for isolation barrier

**all DFN**
- extend DFN\* filter, [explained here.](https://github.com/KiCad/kicad-library/pull/501#issuecomment-213938524)

**MAX3051**
- change pin name VSS -> GND

**MCP2551-I/P, MCP2551-I/SN**
- Vref pin is power output

__MCP2562\*__
- Vio pin is power input


and solve conflicts. :)

SchrodingersGat:
I have extended the FP filter as requested and done other FP improvements for all the CAN chips.
Other fixes have been made too as requested.

diggit:
Nice,
Please place FP field below symbols (MCP and LTC drivers). It is still visible in symbol placer.

SchrodingersGat:
Ok, I have moved the FP fields as requested.

SchrodingersGat:
@diggit any further requirements for this PR? There's another PR outstanding for CAN chips so I'd like to get this one merged for consistency.

SchrodingersGat:
Location of SPLIT pin moved to right-hand side, between CANH and CANL pins (where appropriate)

CarlPoirier:
Hi @SchrodingersGat and @diggit,

Thanks for the wonderful work here. I'm merging this so Oliver you can proceed with the other PR.

Carl

SchrodingersGat:
Cheers, thanks @CarlPoirier

PR #489, opened 2016-04-14
Added ADUM3160 and ADUM4160 USB isolator chips

SchrodingersGat:
e.g. http://www.analog.com/media/en/technical-documentation/data-sheets/ADuM3160.pdf

Application: kicad
Version: (2016-04-10 BZR 6684, Git 4d77388)-product, release build


diggit:
Symbol look good.

What do you think about 2 separation lines allowing some data direction drawing etc.?
Eg. look on: ISO7342C
Line does not have to be dashed.

Please remove ADuM4160 from analog_devices.lib, because you replace it.

SchrodingersGat:
@diggit I have added better separation lines but I'm not sure that the direction arrows make sense in this case because the D+/D- pins are not 'directional' line in the case of the other isolator ICs. In the ADUM?160 chips, the isolated lines are bidirectional.

diggit:
Arrows do not apply here. 
Extend FP filter to match wide version only:
- SOIC-\*_7.5x10.3mm_Pitch1.27mm\*

SchrodingersGat:
@diggit i have improved the fp filter

diggit:
Thanks!
Nice work you did, merging.

Issue #467, opened 2016-04-10
Broken library ir


diggit:
Library of international Rectifier is real mess. Wrong descriptions, missing keywords... PbF suffix does not belong here.
Symbols need some rework. IR211* symbols have outputs on left and inputs on right - exact opposite of KLC 3.5.

El3ctr0nic:
I can fix IR lib. Remake mosfet driver using pull request 387 style drawing.
Pls check my pull request and give me some feedback.
Regards

diggit:
PR #387 merged. It will be great if you can take care of it. BTW I don't think those IR21* are similar to your IRS2181\*.

El3ctr0nic:
I find some way to fit it ;) Give me 2-3 days

PR #455, opened 2016-03-30
Modified LT1761 LDO regulators family.

mpkopec:
More compact design now. Changed names of particular devices to match
pinouts and part numbers from the datasheet (page 2, section Pin
configuration).

PR #491, opened 2016-04-14
ir.lib klc fix, LSM303DLHC redraw and minor fixes

El3ctr0nic:
Try to fit KLC
Some symbols redraw


El3ctr0nic:
I something mess up pls wait 4 update ;/

diggit:
Thanks that you took care of #467
there are few more things to fix:

**ALL**
- remove [ ] aroung A,V,mA,....
- FP filters - use \* suffix and without explicit pincount, -> *TO-263*\*, *SOIC*\*, *DIP*\*, *LGA*\* etc. (pin count matching is separate feature in KiCAD)
- FP fields - when component has only 1 FP, use full format (Library:FootPrint) -> *TO_SOT_Packages_SMD:TO-263-5Lead* etc. and set invisible, otherwise leave empty

**IR**
- make symbol of same shape and size for IR\*21\* chips in library, try to place their common pins to share same position, it helps easy replacement [explained here](https://github.com/KiCad/kicad-library/pull/455#issuecomment-207978528) - same thing you did for LSM303DLHC

**LSM303DLHC**
- C1 pin type passive
- SDA pin type bidir

El3ctr0nic:
I try to fit it best I can ;) IRS2181 can be 1:1 repalce with IR2104. IR2110 have more pins (separate suply for outputs and input logic) so it hard to make replacable 1:1 ;/.  Supply pins  Vss and Com are usulay connected by a trace so if i change their position it make corss wire drawing ;/ If someone use optoisolation to drive inputs pins, Vcc and Vdd also con be connected ;/ 
if i change size it broke same design style ;(
 If U have somme ideas pls help ;)

Replacing to IR2110 need minor changes... I think....
Thanks for all suggestion and issues advices in my pull request :) 

Sorry for my english it's not my primary language ;)

irlib

diggit:
I am aware of difference between these drivers. You made good compromise.

Please remove pin count from FP filters. (SOIC-8\* -> SOIC-\*). KiCAD has separate filtering option for matching pin count.
For IR2104, SD pin in datasheet is inverted.
Set pins Vb and Vs as passive.
Then push commit to let us do the final check.

English is not my native language too. ;)

diggit:
Thanks :)
Next time, please put some basic info what are you fixing to commit message, *Fix ir lib pin types and FP filters*.
When you push commits to your branch, we don't receive notification. When you are ready, write some message to PR to notify us.
Nice work you did, merging.

PR #492, opened 2016-04-14
Added series of Silicon Labs Si473x-D60 radio receivers

swalense:
This pull request includes a symbol for the Si4735-D60 with SSOP package and aliases for the Si4734, Si4731 and Si4730 (that only differs by band and RDS capabilities).

https://www.silabs.com/Support%20Documents/TechnicalDocs/Si4730-31-34-35-D60.pdf

~~The two ground pins and the two non connected pins that must be tied to ground are placed in the same spot. To avoid overlapping text and confusion with pin numbering all pin numbers are hidden and only one of the grounds is visible.~~

~~Please let me know if this is a good practice (obviously it prevents connection errors but the user is forced to make the connections based on the pin names only).~~

Edit: the two ground pins are now shown; a non connected pin (to tie to ground) overlaps each of them.


diggit:
Check pin types:
- DOUT output
- DFS input
- SDIO bidir
- maybe more...

**NC**
IDK about one proper solution. It would be easier to do it invisible+overlap way, but someone may want to do some kid of dark magic around these pins...
1. invisible and overlapping with GND pinc
2. visible NC pins
3. or sth. like this...


open for discussion for all...

swalense:
Thanks for reviewing the type of pins. LOUT/DFS should also be bidirectional as the functionality of the pin depends on the software settings.

diggit:
thanks for fixes.
I am ok with current state.
Nice work you did, merging.

PR #493, opened 2016-04-15
Added Texas Instruments TLC555 timer

swalense:
The TLC555 is a CMOS timer similar to the NE555.
http://www.ti.com/lit/ds/symlink/tlc555.pdf

Note, the variant TLC555CPSR comes in a package referred to as "SO" that looks a lot like the SOIJ-8. However some parameters are outside the specs given for that package (in Microchip packaging specification). Therefore I prefer not to put it in the footprint filter (DIP-8 and SOIC-8 are included).


diggit:
Thanks for FP explanation, right decision.

Use FP filters without pin count. KiCAD has separate filter option for this, which does not rely on FP name. Mach only necessary data.

*SOIC-8_3.9x4.9mm_Pitch1.27mm* -> SOIC-\*_3.9x4.9mm_Pitch1.27mm
- Pitch info is probably not necessary (I've never seen different pitch SOIC), but leave it there
*DIP*\*

DISCHarge pin is open collector type.

Where does your pin arrangement come from? I browsed some 555 schematics and DISCHarcge pin is often used together with TRIGger and THREShold pins, what leads to place them on same side...

swalense:
Indeed I put the discharge pin on the other side without thinking further (my project uses the 555 with a 50% duty cycle that discharges through OUT instead of DISCH).

I makes much more sense to keep it on the other side. So I also moved CONT to the right (I believe that in the vast majority of cases only a decoupling capacitor is attached to that pin).

screen-capture-4

diggit:
Looking more on usage (555 is damn multipurpose ic), 95% cases, RESET pin is tied to VCC and cont through cap to GND. To reduce wire crossings, I think pin order on right side (top to bottom) should be RESET, ,OUT,CONT.

swalense:
You're right, it seems that the reset is rarely used (and thus tied to VCC).
Changed in the latest commit.

diggit:
Thanks!
Nice work you did, merging.

PR #495, opened 2016-04-19
Added Atmel AT42QT1010 and AT42QT1011 SOT-23-6 touch sensors 

swalense:
This pull request adds a symbol for the SOT-23-6 variant of the Atmel AT42QT1010 touch sensor with an alias for the AT42QT1011 (which only differs by the absence of recalibration feature).

http://www.atmel.com/Images/Atmel-9541-AT42-QTouch-BSW-AT42QT1010_Datasheet.pdf
http://www.atmel.com/Images/Atmel-9542-AT42-QTouch-BSW-AT42QT1011_Datasheet.pdf

Notes,
- SNSK and SNS are placed on one side as a capacitor must be connected between these pins;
- there is probably no good place for SYNC as it can be tied to VDD (for fast detection), ground (low power mode) or controlled externally.


diggit:
Can you swap left and right pins? Control interface on left and "controlled" pins on right. Everything else look good.

swalense:
Done.

diggit:
Thanks!
Nice work you did, merging.

PR #496, opened 2016-04-19
Added 2N7002 N-MOSFET

swalense:
I know that adding a transistor is like a drop in the ocean... But since it is a relatively common model and since it provides the right pin assignment effortlessly, why not submitting the one I have in my repo for review...

Notes,
- The symbol is a direct copy of the BSS138 and fails on the 100mil grid check because the gate pin is not centred.
- There are variations of the maximum current depending on the manufacturer. The 115mA in the description seems to be a common value and corresponds to the maximum continuous current.

https://www.fairchildsemi.com/datasheets/2N/2N7002.pdf


diggit:
Everything counts.

Why did you create separate symbol? BSS138 has same FP filter. Just fix its FP field and you can link your transistor as alias of it.

100 mil grid is problem for such symbols, ignore it. (for now)

swalense:
Indeed it was not very smart to duplicate the symbol. ;-)
I thus aliased the BSS138 and added the library to its FP field.

diggit:
Thanks. Can you convert 115mA to 0.115A ? Most of transistor use basic units for Ic(Id).

diggit:
Perfect!
Nice work you did, merging.

PR #497, opened 2016-04-19
Added Texas Instruments LM4990 audio amplifier with MSOP-8 package

swalense:
This pull request is for the 8-pin version of the LM4990.

http://www.ti.com/lit/ds/symlink/lm4990.pdf

Notes,
- BYPS is placed next to +IN as the two will most likely be tied together to bias +IN.
- ~SHTD is placed next to VDD to ease the always-on configuration.
- VO1 will normally be connected to -IN (via a resistor) to set the gain.

The datasheet refers to a VSSOP package but it is not the VSSOP-8, which is a 0.5mm pitch package. The LM4990 has a pitch of 0.65mm and exactly matches the footprint of the MSOP-8, hence the FP field and filter.


diggit:
There is MSOP with power pad which has 0.65mm pitch, but it has 10 pins. Are you sure, you are not referring to this package?

swalense:
I compared the MSOP-8 footprint in KiCad with the pad and stencil description for the DGK (VSSOP) package in the datasheet; both the pad position and size x/y match. Another TI datasheet also indicates "8-Pin MSOP (DGK) Package"... So I believe this DKG package is really a MSOP-8.

I also have a LM4990MM (without power pad) soldered on a MSOP-8 (0.65mm) adapter and it fits correctly.

diggit:
I got confused by wrong page in datasheet. VSSOP is really matching MSOP-8 FP.
Nice work you did, merging.

PR #499, opened 2016-04-22
Added AD5593R

penoud:
Check with python script checklib.py with onyl 3.8 error


diggit:
You should use latest KiCAD, build 6331 or newer. (this solves white-space changes)
It looks like, your local repository is out of date. There were some changes, which your bfacfe902bb59c1a33dfe82bd3f824cb884fa396 reverts.
Pull from upstream, make changes, check diffs, if ok -> commit. Then create new PR.

Symbol:
- place i2c interface on left and IO pins on right (completely swap left and right side pins)
- set *Vref* pin type to passive, otherwise connecting to power output pin generates warning (what do you think @keruseykaryu ? pin is external ref. or connected to internal ref.,  [datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/AD5593R.pdf))
- keywords: *... Reference Single Supply Temperature Sensor ...* are considered as standalone keywords, remove all these keywords but *temperature*
- remove FP info from description
- set FP filters (you can follow recently merged PRs have them correct)

penoud:
THank you for the comment and explanation. Let's work ;-)

PR #502, opened 2016-04-23
Language fixes and description improvements for the conn library

SchrodingersGat:


diggit:
Good catch.

Connector **<**Europe**>** 64 contacts ranges A **et** B
- remove < >
- replace:
 - *A et B* -> *A,B*
 - *A et C* -> *A,B,C*
 - ...

INV3: Symbol general de Connector
- *simple 3pin connector* (imo quite useless symbol)

CNT-RJ45:
- sth like: *RJ-45 with LEDs and shield*

SchrodingersGat:
I have made the requested changes,

also I have improved keywords generally for a lot of the symbols in this library

diggit:
Thanks for improvements!
I don't think, that *connector* keyword is necessary in conn library, but ok.
Nice work you did, merging.

PR #504, opened 2016-04-24
Added MCP2515

fredbar:
MCP2515 is a CAN controller from Microchip.
http://ww1.microchip.com/downloads/en/DeviceDoc/21801e.pdf


diggit:
**[ > Please read through and follow symbol KLC < ](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt#L29)**

Use checklib.py from [kicad-library-utils](https://github.com/KiCad/kicad-library-utils) to check your symbol.

- Your symbol has no logical pin arrangement.
- Do not use Inverted graphical style, pin name overline is preferred.
- Fill description, keywords and document link.
- Hide NC pins.
- 150 mil pin length is enough
- Fill FP filter

fredbar:
Thanks for pointing me to the tools needed to check myself.
I updated the design, and tool did not comply anymore. Please tell me if everything is fine for you too.

diggit:
There is useful switch `--enable-extra` which enables some extra checking (used together with`-vv`). These checks are optional, but help keeping library consistent.

- EC-04 is preferred to follow. Set outline thickness to 0.01 and BG fill.
- ~TX?RTS are all inputs
- fix datasheet link, it is pointing to MCP2551 instead of MCP2515
- fix FP field to full format Library:FootPrint (use *Assign Footprint* button in **T**ext field editor)

Logical pin grouping has higher priority than input/output on left/right. I does not make sense to split SPI pins. Same for TXCAN and RXCAN. Place control pins on left and controlled pins on right, in groups.

fredbar:
Ok, thanks for your very precious advice. I made changes and enabled the extra checks.
Hopefully, it should now please your eyes and mind.

fredbar:
Hi. Any news? If this component is OK, i have other ones to submit that hopefully ought to be fine at first throw. (now that i have a good idea of what you request)

diggit:
Busy week...
Please swap left/right side pins. SPI is control interface together with TX?RTS and RX?BF and should be on left.
150mil pins may improve pin number readability.
Set FP field invisible.

fredbar:
Hi. No problems, i just wanted to up a bit, in case the request became too low in the stack.

If i change the pin positioning to 150mils, rule 3.1 breaks. I understand that pins are a bit more related to their texts when going 150mils but i fail at getting the priorities between rules and your advice.

I am also wondering the differences between control and other stuff.
SPI is not for control. It is data that the chip sends to CAN bus. Should it be seen as a control? Should not imho, and should be seen as the output from the bus, while CAN ought to be the input. 

In the same idea, Rule 3.5 says that "if possible", inputs should be on left and outputs on right. The CAN and SPI are both input and output, so their side is not clear. On the other hand, that rule seems to indicate that we should preferably split SPI, which i would hate. (and i think you would too)

Last, if i understood correctly, you are indicating that control pins (for the chip, i guess) ought to be on side at left, while the others ought to be on side at right. Is that an implicit rule? If so, should CS (and eventually SCK) be seen as control pins?
I am facing difficulties in gathering ICs that are correctly defined (in the PIC16 family), so that i can base myself on it. It is even more difficult as i cannot find some of your requests and advices in the convention file. 
I see no problem in these multiple attempts, as i learn how to provide good stuff to the community. I hope your patience will be great enough so that i can comply and learn. 
While the python script helps a lot in sending good results, the rules are somewhat unclear to me. 

Anyway, thanks a lot for your time and coolness. Waiting for your answers to make the (hopefully) last shot.

SchrodingersGat:
Hey @fredbar , I might just add a couple of points here :)

* From the point-of-view of the schematic, SPI should definitely be considered the *input*, and *CAN* is the output (going to the 'outside world'). 
* You are correct not to want to split up the SPI pins. Functional groups should remain together as far as possible!

Take a look at some of the CAN chips already existing in **interface.lib** (as of yesterday they have all been updated and so should give a better idea of what we're looking for). Replace the *TXD/RXD* interface on one side with SPI, and you should be good to go!

SchrodingersGat:
can
can2

e.g.

fredbar:
Hi. thanks for your comments.
I just did what you requested, so hopefully that should be good now.

Funnily, i had MCP2551 waiting in line and i did just what the IC shows. Time to remove it from microchip.lib and use this one. Too bad i did not see it earlier.

SchrodingersGat:
Excellent @fredbar this is looking much better :) @diggit  may have some more to add here. All I will suggest for now is removing the part name from the description field

i.e. change *MCP2515-E/ST Stand-Alone CAN Controller with SPI Interface, TSSOP-20* to *Stand-Alone CAN Controller with SPI Interface, TSSOP-20* etc

fredbar:
Ok, done. I also moved the housing text a bit so that it is positioned like in your example ICs.

SchrodingersGat:
Almost there - I think the layout of the symbol is good :)

* Change identifier IC? to U?
* Remove *MCP2515-E/ST* from the description field of **MCP2515-E/ST** (as you have for the *I/ST* version)
* The E and I versions have different temperature ranges but this is not reflected in the part description (so there are no clues to separate the aliases). Please add temperature range information to the description of each of the two parts.

Cheers

fredbar:
ok, done.
Sorry for delay.

SchrodingersGat:
Excellent, thanks for your contributions :)

PR #508, opened 2016-04-25
LTC3525 is fixed in "regul" library checked using checklib.py script

alperyazar:
Alliases are added (with different output voltages and different datasheets).


CarlPoirier:
Thank you very much for your contribution!

alperyazar:
Your welcome. Thank you for merge!

PR #511, opened 2016-04-28
Lineartech add

jps64:
Hi there,
-I'd like to add NXP CAN and LIN drivers to interface lib
-derived the NXP LPC175x family from the LPC176x in nxp_armmcu lib
-suggest to add lineartech symbol lib and specific footprints (Housing_lineartech.pretty) for certain LTC parts
I'm specially interested in your opinion about the multi-part opamp approach: the last part is supply only and may be added to any sub-part or kept separate.

The libs passed the checklib.py script (no extra-check option).


diggit:
Your PR is too large. Closing.
Use separate PR when you add/mod/improve something that has nothing in common.
Right below repo title: *... Note that the footprint libraries are the *.pretty repos themselves.*
Keep PRs clean, files like `.gitignore` does not belong here.

PR #514, opened 2016-05-02
New components to regul.lib: LTC3429/LTC3429B

alperyazar:
LTC3429 and LTC3429B (as an allias) are added to regul.lib. Footprint filters are added according to comments in #455. They are created by modifying LTC3525.


keruseykaryu:
People. Don't make a mess. This is not a linear regulator, so it shouldn't be addressed to `regul.lib`. As I remember there is separate library for switching converters. This applied also for LTC3525.

alperyazar:
I didn't think that regul.lib is for only linear regulators because it is rather a general name and there are some existing shwitching regulators too.Now I think dc-dc.lib is more suitable for this IC. If it is, I can move LTC3429 to dc-dc.lib plus other existing switching regulators in regul.lib if they are in wrong library.

BTW, is there any list explaining which library should contian which type of components or should I infer only from library name?

Thank you.

SchrodingersGat:
Hi @alperyazar yes *dc-dc.lib* would be a better fit for this. Unfortunately at this stage there is not a list describing library contents. It is something I want to work on, I think it is a very good idea.

alperyazar:
I moved them to regul.lib. Could you please check @SchrodingersGat @keruseykaryu

SchrodingersGat:
Hey @alperyazar  this looks very good, just two slight things I would change:

1. Can you please move the GND pin to the bottom of the part (see image below)

2. Can you please set the default footprint to one that actually exists in the default .pretty libraries? The purpose of the default footprint is so that the user can select this part and not have to then select the correct footprint. Also please set the footprint to invisible (see image below)

eg

Thanks!

alperyazar:
Hi @SchrodingersGat 

Thank you for suggestions. I applied both of them. Could you please recheck?

I also checked the footprint name suggested by you comparing with the datasheet. I also found that it is suitable. Notice that, I moved the location of footprint to another place that doesn't violate the rule EC03.

alperyazar:
Hi @SchrodingersGat 

Nice point. I modified the LTC3429B's description. Could you please confirm that it is OK?

Thanks.

SchrodingersGat:
Looks good to me, merging :)

PR #516, opened 2016-05-03
Add Max481 to Maxim

penoud:


penoud:
Why violating KLC 3.8? I add description.

SchrodingersGat:
@penoud I believe the KLC errors are there because the aliases you have created (EPA / ESA / MJA) do not have description information.

Also, try to match the visual style of the 485 chips that already exist. I like the visual function indicators that you have used, but they are quite small.

Have a look at, for example, the MAX13432* chips in interface.lib (although these have separate transmit and receive pairs, but you get the idea).

The other 485 chips in interface.lib need to be updated to match the same style.

The basic idea is that a user can swap function-compatible ICs out of the schematic with as little effort as possible.

diggit:
Follow what @SchrodingersGat wrote. Match your symbol with already exiting ones. Unfortunately there is not good 485 driver symbol yet. You can get some idea from already mentioned **MAX13432**\*.

use 0.01" outline width.

IMO you should do 1 symbol without those package specific suffixes, just MAX481. Then create FP filter for all package variants. Otherwise you'll have to create separate symbol for each variant to set appropriate FP field and it's too much identical symbols...
Then you can add aliases: MAX483, MAX485, MAX487, MAX1487 because they share same pinout ~ symbol and fill they description.

diggit:
Use checklib.py from [kicad-library-utils](https://github.com/KiCad/kicad-library-utils) to check your symbols, it has **-vv** mode which tells you exact problem.

Output is quite straightforward:
```
checklib.py -vv --enable-extra  -c "MAX481CPA" library/maxim.lib                  Tue 24 May 2016 07:10:08 PM CEST
library: library/maxim.lib
checking component: MAX481CPA
  Violating Rule 3.8
    documentation OK
    checking alias: MAX481EPA
    checking alias: MAX481MJA
    field: reference, posx 70, posy 100, recommended posx 0, posy 175
    field: name, posx 225, posy -400, recommended posx 0, posy 100
    field: footprint, posx 575, posy -475, recommended posx 0, posy -400
    Component line thickness 0, recommended 10
```
*Extra checking is not so important, but 10mil outline width (EC04) is required almost every time.*

penoud:
Thank you for your comment I will try to update according the MAX13432 and thank for the tip -vv

penoud:
I didn't understand the FP filter advise and it's implications. You mean add inside the component description filter as for example : SOIC-*, DIP-* and so on according the datasheet. And then I add aliases in the component. But how differentiate aliases relationship with footprint?

SchrodingersGat:
Hi @penoud sorry for the late reply!

There is no way to tell KiCad which ALIAS matches which footprint. If a different version of the part has a specific footprint, it is best to duplicate the original part and give that it's own footprint filter.

Does this make sense? Let me know if you have any other questions.

hackscribble:
Hi @penoud 

Many thanks

PR #518, opened 2016-05-03
Added NXP LIN driver TJA1021 and TJA1029, added NXP CAN driver TJA1049

jps64:
I'm looking for recommendations where to place name and device reference fields.
When adding a symbol to the schematic the fields often loose position and alignment.
Is this depending on the definition in the symbol?


SchrodingersGat:
@jps64 regarding field placement, since some recent version of KiCad the fields get placed automatically when you add the component to a schematic, based on a few different rules. Looking at the symbols you have added, I'd say your placement is pretty good.

Nice work on these symbols, just a couple of thoughts:

The footprint field should point to the exact footprint for the particular part. e.g. for the part "TJA1021T" you have the footprint:

> SO-8

which, when you try to associate a footprint with CvPcb, will produce an error because there is no FP named "SO-8".

Instead, this should be (for example)

> Housings_SOIC:SOIC-8_3.9x4.9mm_Pitch1.27mm

which evaluates correctly during footprint association

The example set in https://github.com/KiCad/kicad-library/pull/488 for the MCP2551 has been improved here https://github.com/KiCad/kicad-library/pull/488 but not merged yet.

I hope this helps. Let me know if you have any other queries

jps64:
Hm..., I looked around in some other files and often found the SO-8. I.e. this is the one :)
I updated my symbols to reference SOIC-8 for all SO-8 and changed all footprint search pattern
accordingly - the DFN-8* as well.
I do not agree/understand that FP shall be
Housings_SOIC:SOIC-8_3.9x4.9mm_Pitch1.27mm
Shouldn't it find all SOIC-8 - e.g. if you have a private/company FP lib?

SchrodingersGat:
The footprint *filter* is used to match multiple footprints (eg if you have a private FP lib, as you say). So in the filter field you'll have a reg-ex filter to match a genericized component name.

The default footprint field should link to the associated footprint in the KiCad footprint repositories.

As an example:

untitled

Here the filter allows you to match against any FP with "SOIC" in the name, but will be set to the default footprint by "default"

SchrodingersGat:
@jps64  the fixes to the other CAN chips have been merged into the master branch now (https://github.com/KiCad/kicad-library/pull/488). If you now have a look at the CAN chips in interface.lib (there are a lot more now) then you should have a good example of how to adjust yours.

Cheers!

jps64:
Hi, I've made the fixes...
The TJA... pass the lib check, while there seems to be a lot other stuff in the file meanwhile which fails.

SchrodingersGat:
@jps64 these are looking really good. There is just one issue that I can spot.

* For each part, you have labelled RXD as pin 4 (it should be pin 1) and TXD as pin 1 (it should be pin 4).

* Please place the TXD pin at the top, as below:

pins

jps64:
Hi, I swapped the TXD & RXD pins. Seems that, the message triggers attention ;)

SchrodingersGat:
Thanks @jps64 , merging

PR #522, opened 2016-05-07
Add unshielded RJ12 & RJ45 symbols

inio:
Adds unshielded variants of the RJ12 and RJ45 symbols.  Minor edits to be KLC compliant.

Diff will look bigger than it should until #521 or equivalent is committed.


diggit:
There is more work to be done on RJ connectors.
Here are some thoughts, do you think @SchrodingersGat  @pointhi and others (contributors too) ?

- 0.01" outline
- fill with BG
- pin alignment to 100mil grid
- naming RJxx_STH1_STH2... , where:
 - **xx** - number [10,12,45,...]
 - all **_STH**\* fields if present in symbol
   - \_SHIELD (pin 9 if no LEDs, otherwise 13, or special non numeric value for shield?)
    - \_LEDS (pins 9-,10+,11-,12+)
    - \_MAGNETICS (internal drawing might not be necessary)
- same RJxx\* = easily interchangeable symbols (shield pin on same place for w/ and w/o LEDs variations)

This all will result in FP changes. There are upside down RJ jacks which have 1~8 pins in reverse order...

hackscribble:
Hi @inio 

Many thanks

PR #524, opened 2016-05-08
Added symbol for ON semi MC74VHC1G07

Bitbored:


diggit:
None 74???1G07 from **74xgxx** library suit your needs?? These single gate chips comes from several manufacturers.

Bitbored:
Oh, I did not know this library exists. I'll check if it's in there. And if not, it probably makes more sense to put the generic symbol in there instead of making a vendor specific symbol.

I don't know a lot about 74 or 74xgxx devices, the reason for this pr is that I plan to use this symbol in a project I want to open source...

Bitbored:
I think the 74LVC1G07 should fit my need perfectly, thanks for the information.

I only looked in the 74xx library, since I did not think to find a 74**xxx**gxx device in the 74xgxx library. It's a bit confusing if you don't know what the `g` means.

PR #526, opened 2016-05-10
SPX5205 added

greenscreenflicker:
Added SPX5205, please check, as it's my first commit here.


greenscreenflicker:
Why has this pull request not merged? Please provide details. (I just renamed a pin of an existing IC, and ready it is...)

greenscreenflicker:
Corrected footprint. Please merge now.

SchrodingersGat:
hi @greenscreenflicker - firstly thanks for your contribution, it's great to have more people actively contributing parts to the libraries.

There are a few things you can improve here to move this towards being able to merge:

For the moment I'll just focus on the SPX5205 part...

1) Pin Types - VIN should be "Power Input"
2) Aesthetic look of the symbol. Here, try to match other parts in the library. Generally, the with of the symbol lines should be 0.01"  and the shapes should be filled with "background" 

(Note that the regul library is lagging behind a bit in terms of meeting KLC requirements, so a lot of the parts in there need improvement. Have a look at AP131-15, for example 

3) Footprint filter. As a general rule, the assigned footprint filter should match any and all relevant footprints. In this case, we want to match "SOT-23-5" in addition to other footprints that may exist (e.g. "SOT-23-5_HandSoldering"). So in this case, a good FP filter would be "SOT-23-5*". A good way to test this is to assign the FP filter, and then run CvPcb to assign a footprint to the new symbol. If you filter by keywords, only the appropriate footprints should be shown in the right-hand column

4) Default footprint assignment. As this component is only available in a single footprint option, you are correct to assign a default footprint. However, KiCad cannot resolve the footprint "SOT-23-5". If you run CvPcb, it will error and say it cannot find a footprint. It is better to browse to the correct footprint and select it explicitly - in this case, "TO_SOT_Packages_SMD:SOT-23-5"

5) Aliases. If you have the time, it would be helpful to add the possible aliases of this part (available on p8 of the part datasheet). Each one will need a unique description, this can be easily achieved by making a master part, then for each alias selecting "copy description from parent" and changing only the necessary details.

I hope this isn't all too overwhelming :) Let me know if you have any other questions.

SchrodingersGat:
This is looking much better! A couple more points, though.

* Please fill out the component description fields (Description, keywords and datasheet)
(Have a look at how AP1117 has been formatted, for e.g.)

* Move the (invisible) FP field out of the way of the component (once again e.g. AP1117)

SchrodingersGat:
Hi @greenscreenflicker you'll need to merge kicad/master back in to fix the conflicts, thanks :)

hackscribble:
Hi @greenscreenflicker 

Many thanks

PR #527, opened 2016-05-10
Added Maxim DS3232M RTC

swalense:
The symbol is mostly a copy of the DS3231 added a few weeks ago, with the following exceptions
- SOIC-8 package (no NC pins on this one)
- The 32 kHz pin has the type "output" as it is a push-pull construction rather than an open-collector.

https://datasheets.maximintegrated.com/en/ds/DS3232M.pdf


SchrodingersGat:
Thanks @swalense, this looks good

PR #528, opened 2016-05-11
Added Diodes Inc. PAM8301

swalense:
This pull request adds a symbol for the Diodes Incorporated PAM8301 class D amplifier. Since there is a library for audio components this part is added to the latter rather than a Diodes Inc. specific library.

Note, the datasheet mentions a TSOT-26 package which is a Thin SOT-23-6, hence the FP and filter.

http://www.diodes.com/_files/datasheets/PAM8301.pdf


SchrodingersGat:
@swalense nice work, this passes all KLC and the FP and filter are good.

PR #530, opened 2016-05-12
Add ACPL-P434 and HCPL063 to opto.lib TLP250 and HCPL3120 description minor fix and text position changes

El3ctr0nic:
opto.lib ACPL-P343 and HCPL063 added


El3ctr0nic:
Hi,
Why my PR is hanging ?
I made something wrong ?
Regards,
Marcin

SchrodingersGat:
Hey Marcin these look great!

The only request I have is to add pin names to each part. You can set the component options to "hide pin names" to keep the "clean look" of the parts, but then at least users can click on a pin to quickly check the name/function of the pin. 

Other than that, really good :)

SchrodingersGat:
Thanks :)

PR #531, opened 2016-05-13
Added Fairchild N-Channel mosfets in 5x6mm SON8 package

SchrodingersGat:
This PR adds a number of Fairchild N-Channel MOSFETS, in the 5x6mm SON8 package.

A single symbol has been added, with multiple aliases, each with a full description.

https://www.fairchildsemi.com/products/discretes/fets/mosfets/#N=4294910111+4294819895+4294819848&&Nao=0&&showAll=true&&showHrd=false


CarlPoirier:
Hi Oliver,

Your changes to transistors.lib are abnormally huge, can you fix this? Otherwise it looks great.

SchrodingersGat:
Hi @CarlPoirier this is due to saving the transistors.lib file with a newer KiCad version. I don't think there's much I can (or should) do about that. Looking through the diff, it's just lines that have been re-ordered in the file.

CarlPoirier:
Okay, thank you!

PR #533, opened 2016-05-16
Add IRS2103 gate driver to "ir" library

PetteriAimonen:
Hopefully I've managed to follow all the KLC rules; atleast it seems to pass the checks.


SchrodingersGat:
Hi @PetteriAimonen it's always a good idea to commonize parts where possible. All the other gate drivers in ir.lib have a common graphical layout, with the Vcc/Vb pins separated, a narrower package and shorter pins.

If you can adjust IRS2103 to match the other (similar) parts such that a user could substitute one for the other in a schematic, that would be great :)

PetteriAimonen:
Ok, should now be fixed.

SchrodingersGat:
Thanks! Could you also change the graphical style of the LIN pin by setting the graphical type to normal and changing the text to ~LIN

The keywords could also be improved, by adding the "half-bridge" keyword and changing "gate driver" to "gate-driver"

Also, while I did say to match the other aesthetic style, we are also trying to move to having power rail pins added vertically. For this device this will just require putting the GND pin at the bottom of the package as shown below

The other symbols in this lib should be changed too, but if you make it something like the image below, we can merge and deal with them at another time.

untitled

Thanks for your work :)

hackscribble:

PetteriAimonen:
Sorry, but lost interest already.

PR #535, opened 2016-05-16
Fixed default footprints and footprint filters

SchrodingersGat:
This PR fixes default footprint associations and improves FP filters for a number of RS485 ICs in interface.lib


diggit:
@SchrodingersGat I know, that waiting for your PRs being merged might be long, but you should let other librarians to do it for you.
Eg. there is [wide and narrow](https://en.wikipedia.org/wiki/Small_Outline_Integrated_Circuit#Wide_SOIC) variant of SOIC-8. FP filter should take care of it. (matching dimensions and pitch)
I didn't require it earlier, still learning too...

I am busy because of exams, but if you mention me in message, I'll try to look at your PRs.

SchrodingersGat:
Hi @diggit this was merged because of my git n00biness as mentioned here - https://github.com/KiCad/kicad-library/pull/544. After reverting it, these PRs are no longer in the master branch.

I will have to re-issue all my outstanding PRs, but at least they are not still in the master branch.

This goes for all the other PRs that you will see that I have "merged". They have all been undone and I will issue them again.

Cheers,

PR #538, opened 2016-05-19
New element CP2112

PiotrCzCc:
New element CP2112. No other changes in the lib.
cp2112-kicad


PiotrCzCc:
Please merge the commit.

PiotrCzCc:
Please merge the commit.

SchrodingersGat:
Hi @PiotrCzCc  this is a good looking symbol, nice work. Just a couple of points:

* As you have already marked ~SUSPEND with a bar over the top, please change the pin visual type to *normal*
* Add in the NC pins as invisible, and set their type to no-connect
* The current value in the footprint filter should be moved to the default footprint value. In the filter, you should have a regex which matches the required footprint while also offering some other options (e.g. some user may design some custom footprint for this part with a name that matches this filter).

e.g.

cp2112

PiotrCzCc:
All done and some small modifications.

cp2112-kicad_v2

SchrodingersGat:
Thanks :) keep the contributions coming! Merged.

diggit:
Please use full FP format Library:Footprint, 0.01" outline width and FP filter next time. Even when you say to fix something, check it in the end.

PR #539, opened 2016-05-20
Added 3P and 4P crystal symbols

SchrodingersGat:
I think there's unncessary confusion with the current Crystal symbols. This PR does the following:

Removed CRYSTAL_SMD, and renamed to Crystal_3Pin
Copied this and made a version called Crystal_4Pin which has a second GND pin (pin 4)
(This is to match up with 4-pin crystal footprints, we should be matching the datasheet pinout not arbitrarily duplicating pin numbering to satisfy the weird schematic symbol)
e.g. https://github.com/KiCad/Crystals.pretty/pull/4
Adds descriptive text to the new crystal symbols
Changed designator on existing crystal symbols from "Y?" to "X?"
Resaved with new KiCad library version

Having a 3-pin and a 4-pin version allows the generic symbols to match the available footprints better.

This will require a new PR to /Crystals.pretty to renumber duplicate PIN3 to PIN4.


diggit:
What happened to this PR? Did you change your mind?

SchrodingersGat:
@diggit I swept this PR up into another one which did some other crystal fixes too. I'll have to re-issue this PR.

PR #540, opened 2016-05-21
The 3D models of five Epson 32.768 kHz crystal oscillators.

anpaza:
These are the 3D models of five Epson clock crystals.
The footprints themselves are in a pull request to the Crystals.pretty repository.


SchrodingersGat:
Thanks @anpaza  these look good :)

PR #549, opened 2016-05-24
Added ADC and DAC chips

SchrodingersGat:
This PR adds multiple ADC and DAC chips to the adc-dac library

* ADS1118
* ADS1120
* MAX1161x
* MCP4728 (DAC)


SchrodingersGat:
Added multiple ADC / DAC chips as above. Bump :)

diggit:
I'm on it!

Why do you use various designs of ADCs? Because some are differential and others not?

- Use 0.01" outline thickness.
- try to center symbol square around origin
 - MAX1161* 100mils right
 - MCP4728 at least 200mils up
- **MCP4728 (DAC)**
 - you probably meant *quad output* in description
 - match dimensions and pitch in FP filter`*SOP*3x3mm*Pitch0.5mm*`
- **AD1118**
 - it is possible to use inputs as differential, arrow style symbol is ok, but quite large. I know, intersymbol drop-in compatibility...
 - again, match FPs with dimensions, it is necessary for *FN ans *SOP packages
- **AD1120**
 - same thing with FP filter
- __MAX1161\*__
 - again, filter...
 - add package info to description, this is QSOP only symbol

SchrodingersGat:
All done, finally :) LMK what you think now

SchrodingersGat:
Fixed merge conflicts

jkriege2:
Hi!

I can merge this, but some stuff ist still missing:
- MCP4728: could you add the MSOP-package (device is MSOP-only)
- ADS1118: add VSSOP/X2QFN Footprint-Filters (both have the same pin-numbering)
- ADS1120-RVA: could you add the package?
- MAX1161*: would it make sense to split the symbol between the 8- and the 12-channel types and remove the NC pins from the 8-channel types? I find it otherwise a bit confusing ...

Best,
JAN

SchrodingersGat:
## MCP4728

Done, added default package

## ADS1118

Added filter for the QFN footprint

## ADS1120-RVA

This QFN package does not exist in the library as yet...

## MAX1161x

Good idea, I have split these up as you suggested.

SchrodingersGat:
I have now fixed conflicts - ready to merge!

PR #553, opened 2016-05-25
Ipt012n mosfet fix

SchrodingersGat:
This PR adjusts the schematic symbol for the IPT012N08N5ATMA1 mosfet, to match the updated footprint.

The PG_HSOF-8-1 package (http://www.infineon.com/cms/packages/SMD_-_Surface_Mounted_Devices/P-PG-HSOF/P-PG-HSOF-08-1.html) calls for multiple pins to be connected together. 

The footprint (https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/12) looks like this:

hsof

This matches the adjustment that has been made to the schematic symbol for this transistor:

symbol


SchrodingersGat:
Actually I've come up with a better solution. Closing

PR #556, opened 2016-05-26
Added IRF77x9 mosfets

SchrodingersGat:
Associated footprint "DirectFET_L8" is here: https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/16


diggit:
This should be quick. Just fix FP filter to 'IR-directFET_L8-Outline_Names*'.

What happened in https://github.com/KiCad/kicad-library/pull/556/commits/1fd6477f92774c72953632724ed8ea1c7d5cc457 ? There is a lot of changes just in position. Can you try to revert it and set FP and FP filter again, to get rid of unnecessary changes?

PR #559, opened 2016-05-26
RF shielding symbol

pointhi:
I added two rf-shielding symbols (one-part and two-part type), similar to the one used for the hackrf: https://github.com/mossmann/hackrf/blob/master/doc/hardware/hackrf-one-schematic.pdf (last page, left side)

rf_shield

I also added [Shielding Cabinets.pretty](https://github.com/KiCad/Shielding_Cabinets.pretty) to fp-lib-table


SchrodingersGat:
Seems like a good way of indicating it. Merging.

PR #567, opened 2016-06-04
Fixes for LM35 temperature sensor

SchrodingersGat:
* Removed old LM35 symbol
* Added LM35 in TO-92 package
* Added LM35 in TO-220 package
* Added LM35 in SOIC-8 package


diggit:
Thanks for such quick FIX! Can you set default FootPrints for TO-92 and TO-220 too?

SchrodingersGat:
Which ones are the "default"? There are multiple options to choose from in TO-92 and TO-220 footprints which is why I have left it blank

https://github.com/KiCad/TO_SOT_Packages_THT.pretty which ones to choose?

diggit:
You're right, I didn't realize various positions of those FPs. Thanks :)

PR #76, opened 2014-12-15
Added IRF5303, IRF9530N, IRLZ44N mosfets (try #2)

emesik:
As before, the new mosfets, this time applied to the latest upstream version.

I added the fourth pin, another drain connection, which is the big rear tab in TO-220.

PR #568, opened 2016-06-07
update NCP1117 series regulators to KLC compliance, add DPAK parts

Telemin:
As in title. SOT-223 parts updated to KLC compliance and DPAK parts added.


SchrodingersGat:
Hey @Telemin , thanks for your work. A few notes:

**SOT-223*
* Add TAB pin (pin 4)
* Add default package, invisible, as shown
* Add asterisk to footprint filter, the idea of the footprint filter is that it will match any footprints that contain the text SOT223

sot

**DPAK**
Similar to the SOT223
* Add TAB pin
* Add default footprint (in this case it is TO-263-3)
* Fix footprint filter (you still have it set to SOT223)

dpak

SchrodingersGat:
Actually, there seems to be a few opinions on whether the TAB should be added as a separate pin, or not - https://github.com/KiCad/kicad-library/pull/76

@CarlPoirier  what are your thoughts on this? The TO-263-3Lead part has the tab marked as Pin4. Do you agree that it should be enumerated as a separate pin in the symbol?

Telemin:
Okay, footprint filter updated and default footprint added. I've held off on TAB addition until @CarlPoirier has replied.

Telemin:
my bad, committed to the wrong branch

CarlPoirier:
Hi guys,

Regarding the issue about the tab, we will go with the fourth pin. As per KLC 4.4, it should be indicated in the name using the suffix "_MountingTab". A version without the extra pin will be needed for the user who doesn't want to connect the tab.

Thank you for the great work.

Regards,

Carl

Telemin:
Done.  I couldn't find a reference part in the library that actually had a _MountingTab pin as described, but hopefully this will be acceptable.

ncp1117dtag
ncp1117dtag_notab
ncp1117stat3g
ncp1117stat3g_notab

SchrodingersGat:
Hey @Telemin, I would suggest tackling it as follows:

Start with the "_NoTab" version that you have made - this is a nice compact symbol. 

The "_MountingTab" version of the symbol should have the same visual style as the "_NoTab" version (see below!)

mount

Here, just add pin 4 underneath pin 2, also called VO. 

The "_MountingTab" suffix should go in the symbol name, not in the pin name.

Cheers

Telemin:
Okay I understand now, that makes more sense than what I was trying.  Now fixed.
Maybe adding more detail to the wiki on how to make correct symbols would be a good idea? Don't get me wrong the KLC is great and following it produces really good symbols, but as a newcomer to the project I have found it initially a little confusing in places.

I am happy to create a wiki page visually demonstrating the KLC requirements if that would be considered useful?  What do you think?

CarlPoirier:
Hi @Telemin.

Don't use the "_NoTab" suffix for the corresponding symbol since it is just the normal symbol.

As for the wiki, we have the KLC [here](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention). Adding pictures is a good idea and I will definitely take your work into consideration if you do so! As a few guidelines, I think the pictures should be small and concise as to prevent too much scrolling needed. Also, having only one or two sets of symbol/footprint scales displayed on screen would be great and ideally displaying a 100mil grid to match the convention. I suggest you open up an issue in kicad-library for further discussion and sharing.

Regards,

Carl

Telemin:
Hi @CarlPoirier, 

The latest versions already follow that convention, with no suffix for the 3 pin version and "_MountingTab" suffix for the 4 pin version.  (see below)

Re the wiki: Thankyou for your comments. I shall open an issue once I have found some time to choose/create some appropriate graphics.

Regards,

Phil

ncp1117dtag
ncp1117dtag_mountingtab

SchrodingersGat:
Great work here @Telemin . Keep us posted on your thoughts on improving the KLC. I'd love to help you out on this, it's a great idea. Mergin'.

PR #569, opened 2016-06-08
added VDDA and VSSA power symbols

hyOzd:
I've added VDDA and VSSA power symbols by simply copying and editing VDD and VSS symbols.


SchrodingersGat:
Thanks @hyOzd

PR #573, opened 2016-06-09
added horizontal alternative to SW_PUSH_SMALL

hyOzd:
I find SW_PUSH symbol just too big compared to other passive components. SW_PUSH_SMALL is kinda right but its 45° rotated. I've added a similar looking but straight version of it. Here is how it looks.

kicadsmallbutton

As you can see above, I tried to make it close to the size of a resistor. Couldn't think of a good name so I've just added '_H' suffix indicating 'horizontal'.


SchrodingersGat:
@hyOzd  I like it. Can you add a description to the part, and also some pertinent keywords (button, switch, for starters)

Also, can you add the same description and keywords to the existing SW_PUSH_SMALL

If you search for "switch" or "button" in the part selector, SW_PUSH_SMALL doesn't appear because it doesn't have any metadata.

Thansk

hyOzd:
@SchrodingersGat I've added keywords.

SchrodingersGat:
@hyOzd ok thanks. This component breaks some KLC rules, as the pins are not located on a 100mil grid. However, the same is true of the standard R and C symbols.

I'm going to merge this in, as we have to continue to work towards improvement of device.lib anyhow.

PR #574, opened 2016-06-09
add texas instruments TPS2513/4 usb charge controllers

Telemin:
This commit adds the TPS2513(A) and TPS2514(A) charge controllers from TI. 

[http://www.ti.com/lit/ds/symlink/tps2513.pdf](url)

tps2513
tps2514


SchrodingersGat:
Hey @Telemin , great symbols. May I request that you change the footprint filters for each from *SOT-23-6* to *SOT-23-\** - this will allow the footprint filtering tool to match other versions of the footprint e.g. *SOT-23-6_HandSoldering*.

Cheers,

Telemin:
done.

SchrodingersGat:
Further inspection reveals a few other issues that need to be addressed:

* Part fails KLC "extra checking" - Run *python checklib.py texas.lib -c=TPS2513 --enable-extra* gives:

```
checking component: TPS2513
```

EC01 - Pin-type of pin GND should be set to Power Input
EC01 - Pin-type of pin IN should be set to Power Input

EC04 - Set symbol outline to 0.01" (currently set to 0.00")

* Fix GND pin numbering - GND should be pin 2 but you have it set to pin 1 (duplicate of DP1)

* Capitalize pin names to match datasheet (should be "GND" not "Gnd" and "IN" not "In")

* Commonize the pin positions - the symbol layout for TPS2513 is good, I'd like to see TPS2514 have the same shape / pin locations, except that pins 3/4 are "NC" and invisible. Having a common symbol aesthetic allows users to swap similar component symbols with minimal schematic changes, if that makes sense.

Cheers :)

Telemin:
Okay fixes added for EC01 and EC04 failures.

I'm not sure exactly what the EC03 failure means, is the KLC going to get updated at some point to explain all these things?

SchrodingersGat:
Thanks, I'll merge this, these look good.

Regarding EC03, it is still very confusing - we hope to update the KLC rules to be clearer in the future.

diggit:
@Telemin @SchrodingersGat EC03 is about text fields positions and alignments. These are only suggestion and not KLC rule.
If you wand to know more about what EC03 (or any other rule) doesn't like on your symbol, use `-vv` argument fot `checklib.py` script.

PR #575, opened 2016-06-12
Added generic rotary encoder symbol to device.lib

SchrodingersGat:
* Version without button
* Version with button

Example:
sw


SchrodingersGat:
@ashtonchase care to glance over this one?

diggit:
Looks good to me.

It woud make sense to have pins "numbered" as "A","B","C",... to simplify FP connection, but it is not recommended (at least EC02 say so) and it was abandoned for transistors (probably for some reason).

Nice work you did, merging.

PR #576, opened 2016-06-14
Added MCP23017 & MCP23S17, I2C & SPI IO Expanders

ianrrees:


SchrodingersGat:
Hey @ianrrees thanks for contributing. A few things to run through here :)

1. Have you seen the KiCad scripting utilities for checking if your symbols meet KLC requirements?

SchrodingersGat:
Hey @ianrrees , thanks for contributing :) These are great symbols, but there are a few minor things to go through here:

Have you seen the KiCad scripting toolset for checking if your symbols meet KLC?

https://github.com/KiCad/kicad-library-utils

Running your symbols through *checklib.py*, I get the following errors:

```
library: c:\kigit\microchip.lib
checking component: MCP23017
  Violating Rule 3.6
  Violating Rule 3.8
```
* **3.6** Field sizes (pins / ref / text) must be 50mil, you have some set to 60mil)
* **3.8** You should put the datasheet into the documentation field for each part

datasheet

Once the KLC checks are out of the way, there are a couple of other things I would suggest!

* Set NC pins to *hidden* - when hidden, the NC pins won't show up in the schematic view
* Put power pins at top/bottom of symbol (VDD and VSS should be aligned at the top/bottom of the part). Many current symbols do not conform to this but we are trying to make this the standard for new symbols.
* INT pins - I would suggest moving these to the left of the symbol, as they are functionally separate from the GPA/GPB pins. This also allows the symbol to be a bit smaller

e.g
symbol

* Finally, the footprint filters will not work "as intended" as they are currently, as there are no footprints in the footprint library with explicit names like "PDIP-28". You need to make these filters more generic as they are really regular-expressions which are used to match against multiple footprint names.

e.g.

datasheet

Hope this helps :)

SchrodingersGat:
Also I see that other I/O expanders present in the library start with Portx0 at the top, and Portx7 at the bottom, so it would be good if you could match that standard :)

ianrrees:
Thanks for reviewing the submission and for the helpful suggestions!

diggit:
@ianrrees do you plan to commit fixes?

ianrrees:
Yes, I did already after Oliver made those suggestions.

diggit:
I can see only 1 commit, not newer than Oliver's response. You modified existing commit?
Ok, @SchrodingersGat can you review changes?

ianrrees:
I made the changes and rebased, so that the commit to the KiCad library would only have the one message if this Pull Request is accepted.  Sorry if that's not the usual protocol for the KiCad libraries, and I should've made my comment "Thanks for reviewing..." a bit less contextual!

FWIW - Github shows what happened by placing the commit where it fits chronologically in the discussion thread.

SchrodingersGat:
You still need to add the wildcard (*) character to the end of each footprint filter, so that they function correctly.

ianrrees:
Whoops!  Sorry about that.  Asterisks added.

SchrodingersGat:
Thanks @ianrrees for the fix.

@diggit  for some reason I cannot open this PR on my machine (all others are working fine, I can't pull this one :/) Are you able to do final review on this one? Thanks :)

ianrrees:
No problem!  What sort of problem are you having with looking at the PR?  Just a guess - might be worth trying `git checkout 6d933f0` after a fetch of my repo.

diggit:
I am on it!
Please solve conflicts and...

Fix these please (applies for both symbols):
- set outline to 0.01
- improve FP filters to: 
 - `DIP*W7.62mm*`
 - `SOIC*7.5x17.9mm*Pitch1.27mm*`
 - `SSOP*5.3x10.2mm*Pitch0.65mm*`

**MCP23S17**
- fix description and keywords, this is SPI version

jkriege2:
Note: Both devices are already in microchip.lib!
see: https://github.com/KiCad/kicad-library/pull/727

Best,
JAN

PR #577, opened 2016-06-14
Add Traco Power TSR 1-2450 symbol

Schlumpf:
Here comes my first symbol. Another DC/DC regulator.
http://www.tracopower.com/products/browse-by-category/find/tsr-1/

I would like to add the other TSR-1 models, footprints and 3D models, too.


ashtonchase:
Hi @Schlumpf,

Please update the documentation file name with the PDF link.This will clear up violation of [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) Rule 3.8. 

Also, have you considered creating aliases for the other available output voltage versions. No means a requirements, just a thought.

reportingsjr:
test this please

reportingsjr:
test this please

reportingsjr:
test this please

Schlumpf:
Thanks for the tips.
Now I defined the datasheet twice, in the "field properties" and the "component properties". Is that correct? KLC 3.8 is fixed.

And I added the aliases.

Schlumpf:
I added footprints and 3D models. See my other PR here: https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/5

ashtonchase:
Everything looks good here. Thanks for you contribution! 

Also, double thanks for the aliases!

diggit:
Not bad for first symbol.
There is several things to fix.
- FP should point to existing footprint (use FP selector), otherwise leave empty, invisible
- FP filter should match all suitable existing footprints, use asterisks to match only necessary parts of FP name
- Dastaheet field - leave empty, otherwise it collides with more possible ALIASes

Schlumpf:
I can clear the datasheet field. Is this field deprecated?
But what's about the FP? The device has a custom footprint (https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/5). The pins are compatible to a TO-220 FP but the size of the housing doesn't fit. The TSR1 FP is larger.

diggit:
>I can clear the datasheet field. Is this field deprecated?

Yes

> But what's about the FP?...

Set FP field to full format _Library:Footprint_ I guess `Converters_DCDC_ACDC:DCDC-Conv_TRACO_TSR-1` (you can fill it by selecting proper FP) and FP filter to `DCDC-Conv_TRACO_TSR-1*` to allow modified footprints like *DCDC-Conv_TRACO_TSR-1_LargePard* and so

Schlumpf:
I fixed it with a new PR: https://github.com/KiCad/kicad-library/pull/616

PR #579, opened 2016-06-15
New symbol for linear -> LTC6102 (current sense amplifier)

marcoster:
Hello,

this adds the Linear Technology LTC6102 current senes amplifier.
Symbol was checked with *checklib.py*, hope the rest also fits to the defined standards!

Best regards,
Marco


SchrodingersGat:
Hi Marco, thanks for your contribution. A few things I'd like to see improved somewhat:

* Footprint options cannot be represented using common symbol

The DFN footprint has 9 pins, whereas the the MSOP only has 8. You have footprint filters for both parts but really they should be separate symbols, each matching the appropriate footprint.

* Device options should be represented using separate symbols

The LTC6102-1 device has a different pinout to the LTC6102/LTC6102HV. While one has an ENABLE pin, the same pin on the other device should be grounded. This means that the relevant pin on each symbol should be in a different location based on its function.

As a quick example:

normal
dash1

* Symbol could be a bit smaller

The size of the symbol seems quite large for the small number of pins. For comparison, have a look at some of the other current shunt sensors in the library.

* Keyword improvements

Currently you have the keywords *curren sense amplifier* which are already present in the device description (library search is performed on the keywords and on the description). I would suggest that the keywords could be improved by adding other related terms such as *current sensor shunt*

marcoster:
Thank you for your quick and great response.

Regarding the first (just a bit) and second points - what variants should then be created?
The three main parts (normal, HV and -1) in combination with the packages would result in 6 parts, but there is also the temperature-range in the device-number leading to a total of 18 parts.
Should all of those be created seperately or should I use some kind of wildcard character for the temperature-range?

Wouldn't be a problem for me (I guess using aliases shlould do the trick - the symbol itself only differs cause of the package anyway) - I'm just not shure if I should *bloat* up the library?

SchrodingersGat:
I would suggest part numbering such as this:

* LTC6102xDD
* LTC6102xDD-1
* LTC6102xMS8
* LTC6102xMS8-1

for a total of 4 symbols

SchrodingersGat:
Nice work. I think the only thing remaining is to add some identification to the description fields of each part - if a user is perusing the library and sees four symbols which each say *Precision zero drift current sense amplifier*, how will they know which one to pick?

I would suggest something like the following:

select

1. For each component, add the "HV" alias variant
2. Identify the separate components as I have shown above, indicating functionality differences and footprint differences

Also, as the symbols are now specific to a given footprint, you can add a "default" footprint from the kicad pretty libraries:

msop

dfn

Set the footprint text to invisible, and ensure that the full path of the footprint (library:part) is specified.

SchrodingersGat:
Excellent work, thanks for your efforts :)

PR #584, opened 2016-06-17
adding TCA9555 16 bit I2C IO expander

hyOzd:
Added Texas TCA9555 16 bit IO expander IC. 2 symbols added for two different packages, 1 has an alias (in total 3 listed).

tca9555

[Datasheet](http://www.ti.com/lit/ds/symlink/tca9555.pdf)


SchrodingersGat:
Hey @hyOzd , symbol looks good (at first glance, I'll check it out in further detail).

Regarding the .dcm file:

1. Remove commas from keywords
2. You should try to reduce the number of "filler" words in the component description (and use commas here ;) ). This allows for the information to be a bit more readable when a user is browsing the component library.

e.g.

*16-bit I/O expander, I2C and SMBus interface, interrupts, QFN-24 package*

hyOzd:
@SchrodingersGat, thanks for suggestions. I've just implemented them.

SchrodingersGat:
Looking good. Two small additions and we'll be merging!

* Change thickness of body rectangles to 0.01" (rather than 0.00")
* Add a default footprint for the TSSOP-24 package:

fp

Thanks

hyOzd:
Done.

SchrodingersGat:
Sweet, thanks :)

PR #586, opened 2016-06-22
Added 4pin header footprint (47053-1000) & USB6B1 ESD-IC

matyro:
Added PC-4Pin Fan pwm header. Step file is available for 3D view
[Link](http://www.molex.com/molex/products/datasheet.jsp?part=active/0470531000_PCB_HEADERS.xml&channel=Products&Lang=en-US)

Added STm USB-ESD protection
[Datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/3e/ec/b2/54/b2/76/47/90/CD00001361.pdf/files/CD00001361.pdf/jcr:content/translations/en.CD00001361.pdf)


SchrodingersGat:
Hi @matyro , for future PRs I would suggest splitting them up into individual requests so that they can be handled separately. I'll have a go at what you have got here.

For the USB6B1, there are a few issues:

* The Vcc and GND pin types should be set to "Power Input" rather than "Bidirectional"
* Please set the symbol fill to "background" and the width to 0.01"
* Add a default footprint for the shape, with visibility set to "Hidden"
* Move the reference and name out of the symbol, as shown:

usb

I'll have a look at your PR for the Molex footprint before commenting on the models

SchrodingersGat:
Nice work :)

PR #587, opened 2016-06-23
Created LTC4417

openelectron:
Created LTC4417
Added aliases for the variants with the same pin out.


SchrodingersGat:
Hey @openelectron I believe that these parts would be better suited to the "Power_Management.lib" library.

openelectron:
I can move them over. I did look for any guidelines on how to decide which library to put parts in but couldn’t come to any real conclusions.

SchrodingersGat:
The symbol looks pretty good, and doesn't conflict any KLC which is great :)

Just a couple of tiny suggestions and we'll be able to merge!

* Please add in "QFN-24 package" to the description. In the future if someone adds the SSOP variant, there should be an easy way to differentiate within the library browser
* The pins are a little bit crowded as they are only 0.1" long. I would suggest making each pin 0.15" long, and bringing them out such that the ends of the pins are at +/- 0.5. This will necessitate making the symbol body a little bit wider - I think it's a little narrow so this would also be good.

Apart from that, it looks great :)

openelectron:
Updated as requested. 

Thanks

SchrodingersGat:
Please add the "QFN24" addition to the aliased parts descriptions, too :)

openelectron:
all done.

SchrodingersGat:
Nice work, thanks :)

PR #588, opened 2016-06-23
created IRF7324

openelectron:


SchrodingersGat:
Hi @openelectron , why are there so many changes to these files? What version of KiCad are you running?

openelectron:
I'm using 4.0.2.

Somehow the order of the components in the library file has changed. I'm going to close this and make a new PR from a new branch.
Thanks.

SchrodingersGat:
@openelectron the reason for the rearrangement of parts is that version "4.02" is quite old compared to the newest (nightly) versions which have some change to the way the parts libraries are stored.

If you want help with the libraries (and we would be grateful if you could) then you will have to download the latest nightly release so the libraries don't get mixed up. Thanks

openelectron:
@SchrodingersGat  thanks. I will get a nightly version for library work.

PR #589, opened 2016-06-23
added irf7324

openelectron:


ashtonchase:
Thanks for your contribution. Merging.

diggit:
Is IRF7309IPBF somehow different? It looks like alias would be enough this time...

openelectron:
They require different schematic symbols as IRF7309 has an N channel and a P channel mosfet in a single package. IRF7324 has a pair of P channel mosfets.

diggit:
Yikes, sorry, I didn't compare it to your transistor but some else. Never mind...

PR #592, opened 2016-06-26
Add minnowboard lowspeed lure template

mwelling:
This is a lure template of the Minnowboard MAX/Turbot lowspeed expansion header.

Signed-off-by: Michael Welling <mwelling@ieee.org>


mwelling:
Please looks this over and let me know if anything needs to be changed before merging.

template/minnowboard-ls-lure/minnowboard-ls-lure.pro
```diff
@@ -0,0 +1,60 @@
...
+BoardOutlineThickness=0.100000000000
+ModuleOutlineThickness=0.150000000000
+[cvpcb]
+version=1
+NetIExt=net
+[general]
+version=1
+[eeschema]
+version=1
+LibDir=/home/michael/projects/kicad/kicad-library/library
```
mwelling: Does these local paths need to be removed?

diggit: Yes, as it becomes useless on different machine.

diggit:
Thanks for your template design!

**Needed changes:**
Manually added FootPrints in PCB are not good idea. They disappear on next pull of changes from schematic. You need to use some mechanical symbols ans assign it mounting holes you use in PCB.
No FP from Mounting_Holes library suits your needs?

What is the purpose of those keepout areas on B.Cu layer? If you want cutouts for connectors, modify Edge.Cuts drawing adequately.

diggit:
And you should lock these FPs to prevent accidental move.

template/minnowboard-ls-lure/minnowboard-ls-lure.pro
mwelling: Okay. Will remove them before the next merge attempt.

mwelling:
Okay. I will look at the Mounting_Holes library and see if they match, I am sure these are standard enough mounting holes.

Those keepouts are because of the proximity of connectors on the base board. The are meant to keep through hole and bottom side components from being used in those areas to avoid collisions with the connectors below.

Plan of action:
I will add the mounting holes in the schematic, remove the local paths, and lock the mounting holes in the PCB. I will use a standard mounting hole from the Mounting_Hole library if one matches the needs of the board.

Please comment on the bottom side keepouts and anything else that may be blocking the merge.

mwelling:
I was informed that there is n0t standard schematic symbol for mounting holes.

Here are the options as I see them:
- Create a local schematic symbol for the template.
- Push the template without the schematic symbol for mounting holes and lock the mounting holes in the PCB.
- Create a new standard mounting hole symbol and wait for it to be merged.

What is the best option?

diggit:
Waiting for symbol may take a while.
Locking symbol seems to be ok as solution for now.

diggit:
Or you can use some generic 1 pin component if you wan to connect mounting holes. (eg. to GND)

diggit:
Template looks good. Any comments @SchrodingersGat @ashtonchase ?

mwelling:
What is holding up this merge request? Still need some more changes?

ashtonchase:
I will try to take a look at this in a day or two.

ashtonchase:
Mounting hole schematic symbols were added to the library in https://github.com/KiCad/kicad-library/pull/742 .  We had a related conversation about this in https://github.com/KiCad/kicad-library/pull/765.  You can see the two versions added in the bottom right of the screenshot : PR 742

Can you utilize this?

mwelling:
Sure I can add them if you deem it necessary.

mwelling:
Now the schematic page has the new mounting hole symbols. Let me know if anything else is required.

ashtonchase:
I did a check of the board dimensions against their provided [3D model](http://wiki.minnowboard.org/images/e/ec/Turbot_F200_Board_STEP.zip). 


I believe you need to adjust your keepout area for the USB connectors. 

screenshot from 2016-12-30 20-40-57


screenshot from 2016-12-30 20-36-13
screenshot from 2016-12-30 20-38-25

jkriege2:
Also KiCAD has problems opening the logo.file ... is it possibly a JPEG, not a pNG? KiCAD complains this is not a PNG ...

Best,
JAN

PS: I'm using a fairly recent nightly.

mwelling:
The idea was to make the keepouts the same size on both sides for simplicity. Turns out I did not place them quite right to cover both connectors.

If I shift them both down by ~1.5mm they should both cover the connectors while remaining the same size for simplicity.

Is this acceptable?

As for the icon it appears that it is a JPEG in PNG clothes:
icon.png: JPEG image data, JFIF standard 1.01, resolution (DPI), density 72x72, segment length 16, progressive, precision 8, 64x64, frames 3

I will fix the icon and adjust the keepouts based on your feedback to question above.

mwelling:
Updated the icon and keep out drawing.
Decided to just use the dimensions ashtonchase provided.

Please advise if further modifications are needed.

ashtonchase:
Hi @mwelling.

Here is your current layout:

screenshot from 2017-01-03 21-50-51




I have pondered and tinkered on this one for a little bit. Here is my list of proposed changes:

- There seems to be an error on the dimension on the right hand side of the board. I believe the right most X dimension needs to be 195.58 mm. 
- Since you took my literal dimensions, I don't think there is enough margin in the keepout areas. I suggest swelling them by 1mm all the way around the ethernet and usb connector. 
- I took a better look at the dimensions of the board, and the cut-in dept of the keep out isn't quite right. You can see that with the image above. 
- I think it would be good if the keepout area and the edge cuts are made such that you could easily swap the box of the keepout over to a board edge. See my two pictures below 



screenshot from 2017-01-03 21-58-21

screenshot from 2017-01-03 21-35-30

- That allows you do thinks like this: 
screenshot from 2017-01-03 21-36-39


I have incorporated all the suggested changes that I'm recommending here in the most recent **3** commits: 
https://github.com/ashtonchase/kicad-library/tree/mwelling-master
You could merge those changes into your pull request or I can automatically merge them if you wish. You can always do a diff and compare the changes I made to yours

mwelling:
Thanks for taking a closer look. I pulled the changes and pushed them to my master.

Let me know if further action is required.

ashtonchase:
Thanks for your contribution! I've merged it.

PR #594, opened 2016-06-26
add nxp_tja library for NXP CAN transceivers

HubertD:


SchrodingersGat:
Hi @HubertD there is no need to create a library specific to this part - please add it to interface.lib / interface.dcm

There are already some TJA CAN parts within interface.lib, please model your symbol off those.

Thanks :)

HubertD:
sorry, i forgot to pull changes from upstream - was not aware of the newly introduced TJAs

PR #595, opened 2016-06-27
add aliases for TJA1042, which is almost identical to TJA1049

HubertD:


SchrodingersGat:
Hi @HubertD you will need to add descriptions for the new aliases, too. Thanks

HubertD:
thanks for your patience. will send a new pull request, also including tja1051

SchrodingersGat:
@HubertD  you don't need to send a new PR, just push a new commit to this one with the updated description :) All the changes are logged sequentially so you can keep making changes as you go.

SchrodingersGat:
But don't worry about that for now - I'll have a look through your other PR. In the future, you don't have to make separate PRs for each change :)

PR #598, opened 2016-06-29
added symbol for USBLC6-2SC6 ESD protection diode

hyOzd:
Datasheet: http://www.st.com/content/ccc/resource/technical/document/datasheet/06/1d/48/9c/6c/20/4a/b2/CD00050750.pdf/files/CD00050750.pdf/jcr:content/translations/en.CD00050750.pdf

usblc6-2

I've used the same pin name for corresponding pin pairs, this was intentional and inspired by components datasheet.


SchrodingersGat:
Hey @hyOzd a similar chip already exists in the ESD_Protection library. I think that is a better location for this. Can you move it across?

hyOzd:
: ) I've completely missed that library because I've been working inside an old project, not all libraries were listed.

Any other problem?

SchrodingersGat:
A couple of things stick out:

a) Change the pin type of GND to "Power Input" (this appears in the checklib.py output)
b) Change the footprint filter from SOT-23-6 to SOT-23-*

hyOzd:
> Change the footprint filter from SOT-23-6 to SOT-23-*

But why? If I do so, `SOT-23-5` will be listed in footprint choices as well.

diggit:
We let pin count filter do this kind of job. It really counts pins, so you don't rely on some text.

SchrodingersGat:
@diggit has it right. There is also a second reason. The * is there so that if a user has some custom footprint that has text after the number it will be matched too.

e.g. SOT-23-6_Custom

hyOzd:
> The * is there so that if a user has some custom footprint that has text after the number it will be matched too.

I agree with this one. But in this case filter should be 'SOT-23-6*' in my opinion. Nevertheless I've already pushed the changes.

SchrodingersGat:
Thanks @hyOzd :)

PR #600, opened 2016-06-30
Added 74LVC125 symbol based on the existing 74LS125 symbol.

### LINKED ISSUE #599: 74xx is missing 74LVC125
Opening this issue for adding a low-voltage tolerant version of the 125.

ajf58:
Resolves #599. N.b. this fails to some of the Kicad Library Conventions, but is consistent with the other 125 device.


ajf58:
OK so the CI build has pointed out that this violates the conventions. What's the policy here then? Should I fix the new part, the new part? I think there are 330 pre-existing violations in the 74xx.lib currently. Perhaps a separate issue should be raised to fix the existing symbols.

SchrodingersGat:
Hey @ajf58 I agree that this library needs a lot of work. There's no need for you to fix all the KLC errors in this PR (although if you are willing to assist in cleaning up this library it would be most appreciated). I've had my eye on cleanup of 74xx for quite some time.

Regarding your addition of 74LVC125:

* This symbol is an exact copy of the 74LS125 (except for the text description fields). This is what the part ALIAS is for. You should add 74LVC125 as an alias of 74LS125, so that the graphical symbol will be used for both parts, and you can add unique descriptions for the new aliased part.

If you can adjust this, we can go from there. I'm in two minds as to how far we should go with KLC requirements due to the extensive existing violations...

Maybe some of the other librarians would like to weigh in here @ashtonchase @CarlPoirier @ricardocrudo

ajfeatherstone:
OK, I wasn't sure from the description in [this PDF](http://bazaar.launchpad.net/~stambaughw/kicad/doc-read-only/download/head:/1115%4016bec504-3128-0410-b3e8-8e38c2123bca:trunk%252Fkicad-doc%252Fdoc%252Fhelp%252Ffile_formats%252Ffile_formats.pdf/file_formats.pdf) whether that was what the alias is for. I'm not against tidying up as we go along, but a quick glance at [the KLC](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt) felt like it was very restrictive to the point of making compliance hard. For example, I think [rule 3.1](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt#L32)

"""Using a 100mils grid, pin ends and origin must lie on grid nodes (IEC-60617).""" 
is difficult to fully implement on the 74xx125 parts. A middle ground might be to say that pin ends (the bit that connects to a wire) must be on a 100mils grid, and pin origins  _should_ be on a 100mils grid.

I completely understand that slowly making things a little bit worse doesn't help anyone.

ajf58:
Sorry, I replied earlier using the wrong account. Here's my reply again

OK, I wasn't sure from the description in [this PDF](http://bazaar.launchpad.net/~stambaughw/kicad/doc-read-only/download/head:/1115%4016bec504-3128-0410-b3e8-8e38c2123bca:trunk%252Fkicad-doc%252Fdoc%252Fhelp%252Ffile_formats%252Ffile_formats.pdf/file_formats.pdf) whether that was what the alias is for. I'm not against tidying up as we go along, but a quick glance at [the KLC](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt) felt like it was very restrictive to the point of making compliance hard. For example, I think [rule 3.1](https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt#L32)

"""Using a 100mils grid, pin ends and origin must lie on grid nodes (IEC-60617).""" 
is difficult to fully implement on the 74xx125 parts. A middle ground might be to say that pin ends (the bit that connects to a wire) must be on a 100mils grid, and pin origins  _should_ be on a 100mils grid.

I completely understand that slowly making things a little bit worse doesn't help anyone.

SchrodingersGat:
@ajf58 I'll merge this - if you feel like helping improve this library further it will be much appreciated :)

ajf58:
I'm not against helping, @SchrodingersGat, but where do you see the problems with this library?

PR #603, opened 2016-07-06
Add ADS1251 adc to adc-dac

marcoster:
This adds the ADS1251 24-Bit Sigma-Delta ADC to the *adc-dac* library.

Greetings
Marco


SchrodingersGat:
@marcoster looks great, all I would suggest is that you add the footprint filter:

SOIC*

SchrodingersGat:
Thanks :)

marcoster:
It was a pleasure :-)

PR #604, opened 2016-07-06
Add Cortex Debug Connector to conn.lib

samdolt:
This pull request contains two version of the same Cortex Debug Connector:
- Used with SWD
- Used with JTAG

I have made two version to have shorter name in pin leads.
cortex-debug

Reference: http://infocenter.arm.com/help/topic/com.arm.doc.faqs/attached/13634/cortex_debug_connectors.pdf


diggit:
Hi @samdolt, thanks for your work. There is several things to fix:

Using checklib.py from [kicad-library-utils](https://github.com/KiCad/kicad-library-utils) found these problems:

```
checklib.py -vvv --enable-extra -c "Cortex_Debug_JTAG" library/conn.lib
04:29:48 PM CEST
library: library/conn.lib
checking component: Cortex_Debug_JTAG
  Violating Rule 3.1
    pin: GND (3), posx -400, posy 150
    pin: TCK (4), posx 400, posy 150
    pin: TDI (8), posx 400, posy -150
    pin: GNDDet (9), posx -400, posy -325
    pin: nRESET (10), posx 400, posy -325
  Violating Rule 3.6
    field: UNKNOWN at posx 0 posy 0 size 60
    field: UNKNOWN at posx 0 posy 0 size 60
    pin GNDDet (9): w (POWER OUTPUT), expected: W (POWER INPUT)
```

3.1 - your pins are not aligned to 100mil grid
3.6 - each field must be 50mil text size (even empty one)
EC01 - i am not familiar with this pin usage, might be correct...

(note that version in master has bug, there is PR waiting to fix it - your symbol discovered bug :)

**...things, checklib.py didn't find:**
- FP filters _FTSH-105-01-L-DV-K_, _Pin_Header_SMD_1.27mm_2x5_
 - where are they located? I can't find any matching FP
 - if there were existing FPs, filters should match only necessary parts eg. `Pin_Header_SMD*1.27mm*` if there were any "Pin_Header_SMD"
- missing pin 7
 - if some pin is not used, set invisible. Don't omit it, unless it uses special FP without such pin.

samdolt:
Thank you for your review.

`checklib.py` give me " No violations found" so I hope I have corrected all errors 

I have removed the FP filters, I drew this symbol a long time ago  but never the corresponding footprint.

For the missing pin 7,  this connector come with a pin 7 but it should be cut by hand. So I followed your tips and I have set this pin to invisible.

diggit:
Thanks for quick changes, 'checklib.py' checks only some of KLC rules ;)

In compare to common 2x5 pin connector, this is quite large.

*AVR ISP has pretty bad design, it is just for size comparison*
Note that left bottom rectangle is different from the rest, why?
Also polylines connecting squares to pins are multiple times at some places.
Please set reference alignment to left so it won't travel when someone decides to use ref like P1001 or so.
Also outline thickness 0.01" is preferred.

I played with it's design a little bit.

*Just an idea of miniaturization*

Or "slim" version


My favorite is last one, what do you think?

Sorry about such rework, but I am trying figure out, what is best.

samdolt:
Sorry for the long delay,

I have closed this pull request, as I didn't find a better approach than using CONN_02X05.

jkriege2:
Hhmmm ... a pitty ... I also liked the last Version from @diggit ... and that would be a chance to also change all the other ISP-headers to a common design?

Best,
JAN

PR #611, opened 2016-07-11
Interfaces: Add MAX3072 MAX3075 and MAX3078

myfreescalewebpage:


SchrodingersGat:
It appears that these parts could simply be added as aliases for the SP3481CN part that already exists in the library - all the pins are the same and it is also in an SO-8 package.

myfreescalewebpage:
I have not seen the SP3481 when creating the MAX3072. I'm used to search for maxim ICs for this kind of device.

I have fix according to your comment. Also fixed footprint for the SP3481 (SOIC8).

SchrodingersGat:
Great work, thanks for the footprint fix also.

PR #613, opened 2016-07-11
Regul: Add MAX16910, MAX882, MAX883, MAX884

myfreescalewebpage:
KLC rules fail on other devices.


SchrodingersGat:
@myfreescalewebpage  can you fix the conflicts here :)

myfreescalewebpage:
@SchrodingersGat you mean about the KLC rules for the other devices ? Yes I can.

myfreescalewebpage:
Done.

SchrodingersGat:
Your parts look good, but if you have't made any changes to dc-dc.lib can you please revert these files so that only regul.dcm and regul.lib have changed? Otherwise, looking good.

myfreescalewebpage:
Strange, it's the last conflicts resolution which changed the dc-dc library.... I will proceed.

myfreescalewebpage:
Done, no more conflicts.

SchrodingersGat:
Thanks @myfreescalewebpage  : Some notes on these symbols:

## MAX16910CASA
- Center symbol @ (0,0)
- Increase thickness of symbol box-line to 0.01"
- Please place GND pins at the bottom
e.g.
bom

## MAX88x
- GND pins at bottom of symbol
- The description for all these parts are the same - Are there any functional differences between these variants? They should be reflected in the symbol descriptions for each one.

myfreescalewebpage:
@SchrodingersGat: all comments integrated.
Thanks.
Joel

SchrodingersGat:
Still a few issues here:

### MAX16910CASA
1. You have an extra GND pin (No. 9)
1. Please add MAX16910CATA variant (QFN package)
1. Please add package information to the *description* for each variant (to allow user to differentiate between variants in the part selector)

### MAX88xxxA
All the variants that you have added here should include information (in the description) to describe what the differences are between the variants. In this case the differences are the package and the temperature rating.

So, for example:
MAX882EPA - Please add `-40C to +85C, DIP package`
MAX882CSA - Please add `0C to +70C, SOIC package`

myfreescalewebpage:
@SchrodingersGat 

MAX16910:
- The pin 9 is the exposed pad, it is normal.
- CATA variant added.
- packages description added.

MAX88x:
- Temperatures and packages added.

Also, just two questions for my understanding:
1- how do you organize pull request reviews ? I have made several pull request, and it's always a long work for you and for me to finalize it. Each time I have made a correction you have another issue to report. To save your time and also mine, isn't it possible to make a single review with all issues ?
2- I have now understood that KLC do not describe all rules to make new parts. Where is it possible to find the complete rules ? Or the "perfect" example to follow ? I have the feeling that the rules are always changing....

And a suggestion: why KLC checking and all rules are not integrated in the Kicad editor ? I mean, it is possible to add fields for temperature, package name etc etc so when we add a new part we complete the fields (if not, the editor refuse to save it) and then we save time to review the pull requests. What do you think about that?

SchrodingersGat:
I did ask in my first review that descriptions be improved - but reading this again I see that it was a little ambiguous. Thanks for making these changes.

1. Yes it is simpler for everyone if all issues are flagged in the first review. Sometimes we miss them ;)
1. The KLC is currently being reworked (significantly) to include all the "rules-of-thumb" that the librarians have developed that have never been formalized. Here is the current draft: https://github.com/SchrodingersGat/kicad-library/wiki/KLC

I like the idea of the KLC being integrated into the editor, although I do not see that working, especially as the KLC is still somewhat fluid.

The current DRC included in the symbol editor, for e.g. could be significantly improved (currently only checks for off-grid and duplicated pins).

SchrodingersGat:
Looks good, thanks for the contribution.

PR #616, opened 2016-07-13
Fix TRACO TSR-1 symbol

Schlumpf:
Fix symbol as @diggit wrote in https://github.com/KiCad/kicad-library/pull/577#issuecomment-231536171.
It seems, that KiCad reordered the symbols in the lib on saving my changes.


Schlumpf:
@SchrodingersGat any comments?

SchrodingersGat:
Hi @Schlumpf let's see what we can do:

* Please change type of Vout pin to "Power Output"
* Change the footprint field to "Invisible" (it will become grey instead of purple)
* I would suggest simplifying the description a little bit:
`5V 1A, step-down regulator, input range 6.5 - 36VDC, TO-220 compatible footprint, pin compatible with LMxx linear regulators, -40°C to +85°C`
could become
`5V 1A, step-down regulator, 6.5-36Vin, TO-220 compatible, replacement for LMxx linear regulators, -40°C to +85°C`
* Keywords shouldn't contain the part specs, and could be a bit simpler. Also, please add the 'buck' keyword
e.g.
`5V 1A DC/DC TRACO TSR-1`
could become
`dc-dc traco buck`

SchrodingersGat:
Oh so these components are already merged, but is this PR just updating the footprint field and removing the datasheet field?

Schlumpf:
Yes, but I can fix the things you mentioned, too.

SchrodingersGat:
Great! Also the dc-dc library has now been re-ordered so you'll have to merge kicad/master back in to fix conflicts.

Cheers

Schlumpf:
@SchrodingersGat done

SchrodingersGat:
Great work thanks

PR #617, opened 2016-07-13
adding symbol for Micro SD card socket

hyOzd:
Modified version of SD Card connector for Micro SD cards.

microsd

A datasheet: http://katalog.we-online.de/em/datasheet/693072010801.pdf 

I also have the footprint for the module in this datasheet. I'm planning to make a PR for Sockets footprint library if you agree to merge this one.


SchrodingersGat:
Nice, thanks for contributing :)

PR #620, opened 2016-07-18
DAC7513

mpkopec:


SchrodingersGat:
Hey @mszypulska  a couple of small changes:

* Please change the pin-type of VOUT to power-output
* Change the visual type of the SYNC pin to normal and invert it with a ~ character in front of the name
* Add the _DCN suffix to the part name to indicate this is the SOT-23-8 version of the part
* Change the footprint filter from "SOIC*" to "SOT-23-*"

mpkopec:
Hi @SchrodingersGat actually I have changed the Vout to output as normally DACs are not directly delivering power because of their low driving capabilities. They usually provide a precise reference for error amplifiers and so on. I am wondering what is actually the motivation behind assigning Vout to power output?

SchrodingersGat:
Ah good point regarding Vout. checklib.py threw a warning at me and I didn't actually confirm that it made sense. No worries :)

mpkopec:
Done. @SchrodingersGat Btw. I found a crude way how to reload a lib without restarting KiCad. Simply remove wanted library from the library list (e.g. adc_dac) and then add it again. It works but has a disadvantage of ruining alphabetical order in your lib list which at some point will be a pain. Then you'll need to remove all the libs and reload them, as this will load them in the correct order.

SchrodingersGat:
Nice work, thanks :)

PR #622, opened 2016-07-22
Added six-channel isolator ADuM7643

mpkopec:


SchrodingersGat:
@mpkopec this symbol looks very good, nice work! :)

The only thing I have to say is that the footprint field should be reserved for "Exact match" footprints (i.e. ones that exist already within the KiCad .pretty libraries). 

I don't think a matching footprints exists for this part. If it does, link to it in the footprint field with an exact match string of the format "Library:FootprintName"

Otherwise leave it blank or it will cause some errors in CvPcb association.

Thanks!

mpkopec:
It is possible that matching footprint does no yet exist, but I think it can be generated by the footprint generator, I'll generate it and create a pull request in the corresponding footprint library ;)

SchrodingersGat:
Ok I'll merge for now - once you have made a footprint please come back and adjust this symbol with an accurate reference :) Nice work.

PR #624, opened 2016-07-22
Added opamp - LMV601

mpkopec:


SchrodingersGat:
HI @mpkopec  this is mostly pretty good, just a couple of small suggestions:

eg

* You can add a default footprint for this part (TO_SOT_Packages_SMD:SC-70-6) as there is only one possible footprint (as per the datasheet)
* Make sure you set the default footprint to "invisible"
* Some additional text in the description
* Center the symbol as per the image

mpkopec:
Thanks for the tips, I'll do that this week probably. By the way, do you know any easy way to refresh a library after changing branches without closing whole KiCad?

SchrodingersGat:
No I don't, it's a real pain! I think that it loads the library afresh when you launch KiCad. However if you find a way to do it please let me know :)

mpkopec:
I'll let you know, ofc! We should create a ticket somewhere to the developers somewhere to put a button 'Refresh libraries' :)

SchrodingersGat:
Thanks for the changes, but you have also added the datasheet to the "datasheet" field. This would seem like a good idea, but if anyone tries to make an ALIAS of this part then the datasheet will follow. Having the datasheet in the part documentation field is enough :) If you revert that single change it's ready to go :)

mpkopec:
Done ;)

SchrodingersGat:
Awesome, thanks :)

PR #625, opened 2016-07-22
adc_dac: Added MAX1274, differential ADC

mpkopec:


SchrodingersGat:
Hey @mpkopec once again this looks pretty good. Just a couple of things:

* Add the PAD pin (13)
* GND pins at bottom, Vin pins at top
* Please add an alias for the MAX1275
* In the description field, MAX1274 should say "Unipolar" and MAX1275 should say "Bipolar"

Unfortunately the Kicad libs don't have the correct footprint for this part yet. If you feel like it you could make and submit one to https://github.com/KiCad/Housings_DFN_QFN.pretty

eg

Thanks :)

mpkopec:
Done :)

SchrodingersGat:
Now that I've merged your other PR it has conflicted with this one! Fix the conflicts and we're cooking :)

mpkopec:
I've merged, solved conflicts and tested in the library editor. I cooked an alias for MAX5139 which is MAX5138, the only difference is the resolution. Sorry for not doing it on the proper branch, but it's already merged ;) I'll try to make the footprint for this some time, but I have some problems properly doing that in the wizard so I'll just do it and then await your feedback in the footprint section ;) Thanks for all the help!

SchrodingersGat:
Great work! thanks

PR #627, opened 2016-07-24
Add LT1945

yaqwsx:
Add LT1945 (low power DC-DC converter with positive and negative output). Screenshot below. I don't know why, the checklib script reports violation of rule 3.8 even the description, keywords and datasheet are specified correctly. I think there might be a bug in the check script.

vystrizek


SchrodingersGat:
@yaqwsx what version of KiCad are you running? There are numerous parts in the files that have been re-ordered, making this PR really hard to determine exactly what you have changed.

This could be because you are using an older version of KiCAD - the lastest nightly release will be the most up-to-date for the purposes of library merging...

yaqwsx:
I use Windows build 4.0.2-stable. I noticed the rearrangement, however I have no idea what could cause it. I'll try to investigate it.

SchrodingersGat:
Ah yes there have been some changes to the way that the libraries are arranged since 4.0.2 stable. As the libraries are now "ahead" of the stable release we are trying to use the nightly builds to ensure that the libraries stay consistent. 

With 4.0.3 just around the corner, the libraries should be as "up-to-date" as possible.

yaqwsx:
To fix it, I should upgrade to 4.0.3 and re-import the part?

SchrodingersGat:
4.0.3 is not out yet (shouldn't be too far away now though).

In the mean time, you can use one of the nightly development builds (http://downloads.kicad-pcb.org/windows/nightly/ if you use Windows). All you will have to do is open the KiCad library and re-save it which will re-order the components properly.

yaqwsx:
Even with the latest nightly build, the library gets reorganized and the PR doesn't contain only my changes (I even tried to export my part and import it back to library from current HEAD of master).

SchrodingersGat:
Ah @yaqwsx it actually looks like the "latest" version of the dc-dc library is itself pretty old.

Sorry for hassling you about that - we'll just have to accept the re-ordering of parts in the library.

Moving on!

Just a couple of notes, the symbol is mostly very good :)

* Please change SHDN1 and SHDN2 to ~SHDN1 and ~SHDN2 (see pic below)

invert

yaqwsx:
Ok, updated (I didn't know the trick with ~ to mark inverted pins. Thanks!)

SchrodingersGat:
Great work, thanks :)

PR #628, opened 2016-07-25
74xx: Added SN74LVC1G332.

mpkopec:
KLC violations were made on purpose to comply with the other gates.


SchrodingersGat:
Thanks @mpkopec

keruseykaryu:
The single logic devices have separate library called `74xgxx.lib`. Merging this to the main TTL library `74xx.lib` was wrong and will create a mess.

PR #630, opened 2016-07-28
Add Altera EPM2210F324 (MAX II)

GyrosGeier:
This is the 324 ball, 210k variant of the MAX II


hackscribble:

hackscribble:
Posting image of one unit from the symbol for reference.

image

hackscribble:
Hi @GyrosGeier 

Sorry your pull request has been waiting so long for review. 

If you are still interested in contributing, please check your PR against the latest KiCad Library Convention here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention.

The Travis CI checks are flagging up a number of KLC errors: https://travis-ci.org/KiCad/kicad-library/builds/249617954?utm_source=github_status&utm_medium=notification

Many thanks

GyrosGeier:
Hi @hackscribble,

what is the convention on parts with an even number of pins on one side?

If I distribute the pins evenly along the side, I end up with pins on (x, -50) and (x, 50), which violates the rule about pins on the 100mil grid, and if I move the symbol, I end up violating the rule about the origin being the center.

poeschlr:
In this case either move the symbol or move the pins such that there is an empty space in between. (If you have different functional groups just increase the distance between them.)

hackscribble:
@GyrosGeier is this the right datasheet for the device pinout?  https://www.altera.com/content/dam/altera-www/global/en_US/pdfs/literature/dp/max2/epm2210.pdf

GyrosGeier:
@hackscribble Yes, that's the one.

hackscribble:
In some of the units, the pins are ordered by number then letter ...

image

In other units, the order is letter then number ...

image

Is this intentional, maybe to do with how the escape routing works for the BGA?

GyrosGeier:
@hackscribble Yes, that's based on escape routing, every bank is ordered outside to inside. Bank boundaries are on the diagonals.

hackscribble:
> Yes, that's based on escape routing, every bank is ordered outside to inside. Bank boundaries are on the diagonals.

Thanks, that makes sense.  

Another question (sorry, I am not familiar with Altera devices) ... the pin names include the pin (ball) number e.g. `IO_C3` on pin number C3.  I guess that's the usual convention for Altera CPLDs?  Xilinx FPGAs for example have pin names that are different from the pin number e.g. `IO_L09P_0` on pin B11.

Let us know if you need any help on sorting out the Travis KLC errors.  And, again, thanks for the contribution.

GyrosGeier:
@hackscribble To do it correctly, we'd need multiple configurations. `IO_L09P_0` sounds like the non-inverted half of a differential pair to me, which is a possible configuration on Altera chips as well.

Personally, I find the pin numbers easiest to use, because in the Altera toolchain I do pin assignment also by pin number, even for differential pairs (where I define the non-inverted pin, and get an automatic definition for the inverted pin).

This means that inside KiCad, I need to explicitly make the differential pair visible for the PNS router, but that's not a big hurdle.

SchrodingersGat:
@GyrosGeier we have begun the transition process to a new consolidated symbols library at https://github.com/kicad/kicad-symbols 

The Altera library has now been transferred - please re-open this PR at the new repository. Sorry for the inconvenience.

PR #637, opened 2016-08-04
[enh] add 3d models for würth shielding cabinets.

pointhi:
source: http://katalog.we-online.de/de/pbs/WE-SHC/
legal: Würth gave permission to use their 3d models inside KiCad

*.stp are the original version from the website
*.fcstd are converted models which are positioned and recolored
*.wrl exported from the *.fcstd files

wuerth_smd_3d
wuerth_tht_3d

related pull request: https://github.com/KiCad/Shielding_Cabinets.pretty/pull/6


SchrodingersGat:
I think before we add these we need an accompanying legal notice as obtained from Wurth

We'll also need somewhere designated to put such notices.

@CarlPoirier  @ricardocrudo  any thoughts on this?

jkriege2:
what about having a subdir ` modules/packages3d/Shielding_Cabinets.3dshapes/license/`and putting a copy of the email there + here in the thread for reference?

Best,
JAN

SchrodingersGat:
Given the new licensing agreement, I am unsure that we can legally re-distribute these models. I will close this issue pending any further comments from @pointhi who has not commented here in a while.

PR #629, opened 2016-07-27
Added MCP1640 boost regulater

mkleinschmidt:
This pull request adds the MCP1640 bost regulator from Microchip with two different pin configurations/packages.

[MCP1640.pdf](https://github.com/KiCad/kicad-library/files/385803/MCP1640.pdf)

mcp1640-ch

mcp1640-mc

PR #642, opened 2016-08-09
Added MCP1640 boost regulater

mkleinschmidt:
This pull request adds the MCP1640 bost regulator from Microchip with two different pin configurations/packages.

It is the second try since Pull Request #629 did not really work.

[MCP1640.pdf](https://github.com/KiCad/kicad-library/files/409000/MCP1640.pdf)

mcp1640-ch

mcp1640-mc


SchrodingersGat:
Great work :)

PR #464, opened 2016-04-04
Added files via upload

davidsrsb:
1280/2560 added missing pin 61 VCC and pin 62 GND

PR #647, opened 2016-08-15
Atmel: added missing pins to ATMEGA2560-A

mpdanny:
Atmel: added missing pins to ATMEGA2560-A
Missing pins: VCC (61) and GND (62) missing as per discussion here: https://github.com/KiCad/kicad-library/pull/464


jkriege2:
Please declare both pins as POWER_INPUT!

Best,
JAN

jkriege2:
Thanks!
JAN

davidsrsb:
Thanks, I never saw diggits reply to #464 in April

PR #648, opened 2016-08-16
Added LT6657 series of precision voltage references

SchrodingersGat:
**Datasheet**
http://cds.linear.com/docs/en/datasheet/6657fb.pdf


SchrodingersGat:
Bump

jkriege2:
Nice Work!

JAN

PR #652, opened 2016-08-23
adding Invensense motion sensors

hyOzd:
This PR adds below Invensense motion sensor ICs which have very similar pinouts.

* MPU-6000 [datasheet](https://store.invensense.com/datasheets/invensense/MPU-6050_DataSheet_V3%204.pdf)
* MPU-6050 [datasheet](https://store.invensense.com/datasheets/invensense/MPU-6050_DataSheet_V3%204.pdf)
* MPU-9150 [datasheet](https://www.invensense.com/wp-content/uploads/2015/02/MPU-9150-Datasheet.pdf)
* MPU-9250 [datasheet](https://store.invensense.com/datasheets/invensense/MPU9250REV1.0.pdf)

mpu


hyOzd:
In these symbols I've left exposed pad completely unconnected but there is a special consideration regarding the exposed pad of these components. Here is the relevant section from the datasheet of 6000/6050 devices:

> 11.4.2 Exposed Die Pad Precautions
The MPU-60X0 has very low active and standby current consumption. The exposed die pad is not required for heat sinking, and should not be soldered to the PCB. Failure to adhere to this rule can induce performance changes due to package thermo-mechanical stress. There is no electrical connection between the pad and the CMOS.
11.4.3 Trace Routing
Routing traces or vias under the gyro package such that they run under the exposed die pad is prohibited. Routed active signals may harmonically couple with the gyro MEMS devices, compromising gyro response.
These devices are designed with the drive frequencies as follows: X = 33±3Khz, Y = 30±3Khz, and
Z=27±3Khz. To avoid harmonic coupling don’t route active signals in non-shielded signal planes directly below, or above the gyro package. Note: For best performance, design a ground plane under the e-pad to reduce PCB signal noise from the board on which the gyro device is mounted. If the gyro device is stacked under an adjacent PCB board, design a ground plane directly above the gyro device to shield active signals from the adjacent PCB board.

My plan is this: connect the footprint pad to the GND layer, to achieve this add a pin named GND, numbered '25' to the symbol . But modify the footprint package and disable the solder mask and solder paste layers so that chips pad cannot be soldered to the pcb. Should I submit that as a special package to the QFN footprint library?

SchrodingersGat:
@hyOzd these look great! I also like the idea of the footprint adjustments.

Unless you have a better suggestion, perhaps these adjusted footprints should have the suffix "_IsolatedPad" ?

hyOzd:
> "_IsolatedPad"

That sounds alright. I will be updating this thread when I'm done with the footprints.

hyOzd:
I was just about to create the PR for the footprints then I noticed:

epad

Thickness of solder mask and the copper will cause probably the same problems that soldering would cause. So I did more research and found this clear recommendation from invensense:

> It is strongly recommended to define a keep-out layer beneath the MPU, and not place any trace, fill, or via array on the top layer under the exposed pad...

I've also checked their evaluation boards:

invenses_epad

So instead of "_IsolatedPads" footprints, I will add "_NoPads" : ) Or "_NoEPad", I think this one is more clear.

SchrodingersGat:
Great work. Footprints are almost there too.

smallzzy:
The AUX_SDA pin on the MPU-9250 is actually pin 21 instead of pin 6. It is a strange change that Invensense made.
Can someone fix it?

hyOzd:
@smallzzy I will look into this.

PR #653, opened 2016-08-23
TPS61200, TPS61201 and TPS61202 are added to DC-DC library.

alperyazar:
I am not sure about the footprint filters and package field (F2). These chips have custom TI packages and I was unable to find .pretty repository corresponding to these packages.

Also as a general question, how should I fill the F2 field and footprint filters when a component has more than one package available?

Best.
Alper


SchrodingersGat:
Hi Alper! Regarding your question about the footprint fields and filters: If a part has multiple package options, leave the default footprint (F2) blank, and put an appropriate filter for each package.

The S-PVSON-N10 package is awaiting merge here - https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/18

Feel free to submit the S-PWSON-N10 version too :)

SchrodingersGat:
So, looking at these symbols, they are pretty good! There is only one suggestion that I can see at this stage:

* Make F2 field (default footprint) blank

Cheers

alperyazar:
Hi Oliver,

Thank you for response.

As a last question, is FPLIST entry starting with '*' OK (like *S-PVSON-N10*) or should I remove the leading '*' (conver to S-PVSON-N10*)? Actually I don't know whether it makes a practical difference or not? However as far as I understood, we generally don't use leading '*', for example: https://github.com/KiCad/kicad-library/pull/598#issuecomment-230462090

Best.

SchrodingersGat:
It appears that you don't need to use leading OR trailing asterisks/wildcards for the footprint filters. If you run CvPcb and filter the footprints with the search bar at the top, including * at the start or the end of the search term does nothing.

I think there has been some confusion on this issue for the different PRs! You could just use the filters as following:

S-PVSON-N10

The wildcards are useful but not necessarily in this case.

alperyazar:
Thanks for explanation. 

Could you please recheck? It is ready to merge! :)

SchrodingersGat:
Great work, thanks for the contribution :)

alperyazar:
Thank you too for your patience.

PR #659, opened 2016-08-29
Added ACS711EEXLT Hall Effect Current Sensor

floschmid:
acs711eexlt-15ab
http://www.allegromicro.com/~/Media/Files/Datasheets/ACS711-Datasheet.ashx
The EX-Package of the ACS711 needs a new symbol since the pins (1,2) and (3,4) are connected to a single pad.


SchrodingersGat:
If you want to make a new symbol for the QFN version of this part, I suggest copying the ACS711 symbol that already exists, and simply renumbering the pins. 

By giving the pins invalid names such as "1,2" they will not match correctly against pins on the footprints.

Also, copying the existing symbols allows you to maintain the visual style. Thanks.

PR #660, opened 2016-08-29
Added TPS56528 4.5V to 18V Input, 5-A Synchronous Step-Down Converter…

floschmid:
tps56528
… with Advanced Eco-mode

http://www.ti.com/lit/ds/symlink/tps56528.pdf


SchrodingersGat:
* Change keywords from "4.5V to 18V Input, 5-A Synchronous Step-Down Converter with Advanced Eco-mode" to something like "step-down dc-dc buck converter"
* Arrange pins by function rather than physical layout. Have a look at other parts in dc-dc.lib - inputs should be on the left, outputs on the right
* Change text field size on all fields (including invisible ones) to 50mil
* Change line width of rectangle shape to 0.01"

jkriege2:
please fix conflicts!

PR #661, opened 2016-08-29
Added TPS62150 (TPS62151, TPS62152, TPS62153) 3-17V 1A Step-Down Conv…

floschmid:
tps62150
…erter in 3x3 QFN Package

http://www.ti.com/lit/ds/symlink/tps62150.pdf


SchrodingersGat:
Hi @floschmid please fix conflicts before we can continue :)

PR #662, opened 2016-08-29
Added LSM6DS3H Always-on 3D accelerometer and 3D gyroscope

floschmid:
lsm6ds3h
www.st.com/resource/en/datasheet/lsm6ds3h.pdf


SchrodingersGat:
Please run KLC on these before you push a PR! :)

* Set Pin 11 (NC) to invisible
* Add a footprint filter (e.g. *LGA*)
* The default footprint should point to a footprint that is actually in the library. As there are no LGA footprints yet, leave the default FP field blank
* Set the line-thickness of the rectangular shape to 0.01"

jkriege2:
please fix conflicts!

PR #663, opened 2016-08-29
Added INA219 Zero-Drift, Bidirectional Current/Power Monitor With I2C…

floschmid:
ina219
… Interface

http://www.ti.com/lit/ds/symlink/ina219.pdf


SchrodingersGat:
**Super important**
* SOIC and SOT-23 pinouts are different! They are not compatible so you will need to make two parts, one for the SOIC footprint and one for the SOT-23-8 footprint.
* To do this, make a copy of the SOIC footprint (for which the pins are correct) and then change the pin numbers, and re-save with a different footprint

(Fix the issues below before doing this, too :) )

**Less important**
* Field text should be set to 50mils (footprint and datasheet fields are currently 60mils)
* Change line width of background rectangle to 0.01" (10mils)

jkriege2:
Please fix conflicts.

jkriege2:
Hi!

may I ask, why you pulled back all your PRs? Would have been great to get those symbols into the lib ...

Best,
JAN

PR #664, opened 2016-08-29
Added RT9629B Triple-Channel Synchronous Rectified Buck MOSFET

floschmid:
rt9629b
Driver
http://www.richtek.com/assets/product_file/RT9629B/DS9629B-05.pdf
Created a new library for Richtek devices since no other library fitted to this device.


SchrodingersGat:
Probably a good idea making a new .lib for this, it's pretty specific.

Once again, KLC checking script is your friend!

* Align pins to 0.1" (100mil) grid
* Set NC pins to invisible
* Line with of rectangle to 0.01" 
* Move footprint field up/down so it does not interfere with pin text

PR #666, opened 2016-08-29
Added MCP1805T-xx02I 150 mA, 28V LDO Regulator With Shutdown

floschmid:
mcp1804t-xx02i
http://ww1.microchip.com/downloads/en/DeviceDoc/20002200D.pdf


SchrodingersGat:
Hey @floschmid thanks for contributing, a few notes :)

* Have you run these through the KLC checking scripts? See here: https://github.com/KiCad/kicad-library-utils This will enable you to short-circuit a lot of this back and forth as it performs automatic checks on components.
* Pins should be placed on an 0.1" grid
* Pin names should match datasheet (IN -> Vin, OUT -> Vout) etc
* Change type of pin-1 (Vin) to *Power Input*
* Add a default footprint for this part
* Simply the part keywords ("150mA" isn't really a keyword, I'd suggest changing just to "LDO regulator fixed positive")
Also, the part name should reflect a single device (e.g. MCP1804-1802xOT) and make a part ALIAS for different parts.

As an example of what I mean:

You have the part description to *150 mA, 28V LDO Regulator With Shutdown, Fixed Output 1.8V, 2.5V, 3.0V, 3.3V, 5.0V, 10.0V, 12.0V, SOT-23-5*  - I'd suggest making an alias for each voltage variation.

reportingsjr:
test this please

reportingsjr:
```
Please fix these issues:
checking component: MCP1804T-xx02I/OT
  Violating Rule 3.1
Components Added: 1
+ MCP1804T-xx02I/OT
```

PR #670, opened 2016-09-02
Added Raspberry_Pi.lib. GPIO Pin header for RPi 2 & 3

feichtinger:


pointhi:
test this please

diggit:
Thanks, fix these please:
- use **conn** library
- set outline thickness 0.01" and filled with BG color
- place GPIOnn text nearest to pin (right column pin names) 
- use overline for text inside brackets only `(~SOMETHING~)`
- playce 5V and 3V3 pins on top
- merge *raspberry pi* -> *raspberrypi* in keywords
- description *pin breakout* -> *expansion header*
- solve conflicts

feichtinger:
OK, thanks for the hint. Should be fixed now.

> - set outline thickness 0.01" and filled with BG color
Maybe this should be appended to the KLC and ideally also to the default settings of the library editor.

diggit:
Connector seems ok to me.

**MCP7940N**
- keep datasheet pin names, GND -> Vss
- Vss(GND) pin type -> Power Input
- MFP pin type -> Open Collector
- create FP filter (but leave FP field empty):
 - `SOIC*3.9x4.9mm*Pitch1.27mm*`
 - `TSSOP*4.4x3mm*Pitch0.65mm*`
 - `MSOP*3x3mm*Pitch0.65mm*`
 - `DFN*2x3mm*Pitch0.5mm*`
- set FP field invisible
- make it smaller, sth like this



Also, next time, please do not mix multiple libraries changes in one PR. I almost did not notice RTC chip, not single mention here, except commit message.

diggit:
And DIP FP filter, that I forgot...
`DIP*W7.62mm*`

feichtinger:
> set FP field invisible
OK... How do I do that?

> Also, next time, please do not mix multiple libraries changes in one PR. I almost did not notice RTC chip, not single mention here, except commit message.
Sorry for that. The initial pull request was for the Raspberry only. I didn't assume that github would append my new commit to the existing pull request automatically.

diggit:
In symbol editor: big **T** icon - select FP field - ucheck *visible* or sth like that.

You pushed RTC chip to the same branch as RPi header.

*Btw you have to use double new line to leave quote block ;)*

feichtinger:
> In symbol editor: big T icon - select FP field - ucheck visible or sth like that.

What do you mean with the FP field? There is no such field in this component. There is only the reference and the name field...?

feichtinger:
OK, I found the FP field, but it was a bit tricky. In fact, the only way I found to access the field when it is empty (default for a new part), is to edit the lib file directly with a text editor. Here the field is called "F2" instead.

diggit:
Sorry for no response. In symbol editor, click on big **T**, there are all fields.

feichtinger:
Another pitfall in KiCad...

feichtinger:
OK, now everything should be fixed. Are there any other concerns?

diggit:
Sorry, I didn't realize, there are same. Not smart icon usage :/

Can you merge master to your branch? Now it removes screw terminals added recently.
Use [this link](https://github.com/KiCad/kicad-library/pull/670/files?w=1) to view merge changes excluding whitespace changes. (or add **?w=1** to merge preview url)

feichtinger:
OK, I took the kicad/library master and imported RaspberryPi_2_3 and MCP7940N again.

If every part would be stored in a separate file this would significantly reduce the risk of merge conflicts!

diggit:
Thanks, ready for merge!

>If every part would be stored in a separate file this would significantly reduce the risk of merge conflicts!

Yes, current library format has flaws. There is planned new one, unknown when. That depends on kicad developers.

PR #673, opened 2016-09-04
added one valve and a library for power connectors

jkriege2:


jkriege2:
reorganized into  songle branches, will rerequest pull

jkriege2:
reorganized into songle branches, will rerequest pull

PR #674, opened 2016-09-04
memory: add M25PX32 symbol

piotr-wiszowaty:


jkriege2:
Hi!

thanks for the contrib ... could yo:
1. replace the google datasheet link, e.g. with https://www.micron.com/~/media/documents/products/data-sheet/nor-flash/serial-nor/m25px/m25px32.pdf
2. I pressume you're describing the SO-8 version ... so could you add the correct footprint+FPFilter?

Then I'll look more into it ...

Best,
JAN

jkriege2:
Hi/Good morning!

Could you also fix the conflicts?
I saw that there are a lot of changes (moving of `S ...` in all symbols) ... is that necessary? Or did it "just happen" by KiCAD? This makes the DIFF a bit large ;-)

Best,
JAN

piotr-wiszowaty:
Done.

piotr-wiszowaty:
It seems that KiCAD reorders library statements when saving symbols.
Wild guess: different KiCAD versions (I use 4.0.2) have different
statement orders.
I agree - this behaviour is a real pain during merging.

jkriege2:
There are still conflicts here ... could you fix?

+ some other issues:
- could you switch the style of the pins S#, W#/Vpp, HOLD# to inverted input?
- please make the outline 0.01" thick (see KLC)
- could you rename the component to M25PX32-VMW, set its footprint to `Housings_SOIC:SOIJ-8_5.3x5.3mm_Pitch1.27mm` and the footprint filter to `SOIJ*5.3x5.3mm*Pitch1.27mm*` ... Then for the the QFN-version make a copy with the name `M25PX32-VMP` and the footprint `Housings_DFN_QFN:DFN-S-8-1EP_6x5mm_Pitch1.27mm` and accordingly the filter `DFN*6x5mm*Pitch1.27mm*` ... and please countercheck whether I picked the correct packages ,-)

Best & thanks for the effort and work that you put in this already,
JAN

piotr-wiszowaty:
Done: conflicts, symbol split (checked names and footprints). I'd rather 
leave the pin names as they are because they match the datasheet ones.
How do I set symbols footprint?

Somewhat offtopic: conflicts are like these are a real nuisance. Are 
there plans to keep each symbol in a separate
file? It would probably help in such situations.

jkriege2:
Hi!

about the pins: I didn't mean to rename them, just set the pin-style like this:
flash


you can set the footprint by clicking on this button in the toolbar (kind of non-obvious :-(
fp1

Then select the footprint field and enter the name or search with the button below the field:
fp2


ad off-topic: I think there are plans to have a format like the .pretty-libs for PCB, but I don't know the current status of that ...

cu
JAN

piotr-wiszowaty:
OK, done.
The "T" icon is somewhat misleading indeed.

jkriege2:
I think only one last thing is missing:
The `DFN-S-8-1EP`-package has an exposed pad (pin number 9), which is connected to Vss (see datasheet) internally. For that could you add a pinn9 to the schamtic, maybe with the name `PAD/Vss`. This is important, so you find DFN-packages with the EP when searching by pin number (the device really has 9 pins, not 8 ;-)

Otherwise: GREAT WORK ... and I will merge, as soon as the final changes are done.

Best & Thanks,
JAN

piotr-wiszowaty:
All right. Done.

jkriege2:
Hi!

could you please move the VSS+PAD pins onto a 100mil grid?

```
  Violating Rule 3.1
    pin: VSS (4), posx -50, posy -500
    pin: PAD/VSS (9), posx 50, posy -500
```

piotr-wiszowaty:
Fixed.

PR #675, opened 2016-09-04
regul: add MCP1703AT-3302/MB symbol

piotr-wiszowaty:


jkriege2:
Hi!

First of all: thanks for the contribution!
Second: Could you please alter these things:
1. datahseet is wrong ... correct datasheet is http://ww1.microchip.com/downloads/en/DeviceDoc/20005122B.pdf
2. description is wrong (MCP1754ST instead of MCP1703AT)
3. fix conflicts
4. Could you add the other family members with different voltages as aliases? That would be great
5. Could you also add the other packages in the libraries as copies?

Best,
JAN

jkriege2:
Hi!

thanks already ... I think:
1. we don't need the MCP1703A AND MCP1703AT types ... in the end this just encodes the packaging /tape/reel=T) ... This would cut down a bit on the number of ALIASES. Also the description for the ALIAS-parts is still missing.

JAN

piotr-wiszowaty:
Done. I will add other packages if I actually use them.

jkriege2:
Hi!

looks great so far ... I spottet two more minor things:
1. the outline of the component box should be 0.01" thick (see our KLCs)
2. Could you add the fixed output voltage of the regulators to the description of the aliases? + What does the 16V mean? If its the max input voltage, then maybe write `Vin<=16V` ?
3. Could you make your symbol a bit larger to match the other regulators in the lib, e.g. MCP1700:
mcp1700 


Best,
JAN

piotr-wiszowaty:
Done (also keywords and documentation references).

jkriege2:
... and the other stuff?

JAN

piotr-wiszowaty:
1. Ah, the extra rules... They should not be called 'extra'. Done.
     The checklib.py still reports violation of EC03 but without any 
specifics.
2. This is straight from the datasheet, but you are right (obviously). 
And done.
3. Done.

jkriege2:
Hi!

yes, EC03 you can ignore ... it proposes a position for REFDES and name-filed (try the python script with --fix and it will try to make it the way it thinks, which is not for all symbols a good choice ...)

Could you add the output voltages to the description and maybe resort them so that the most important info is in the front (i.e. read first ;-), e.g. `Low Quiescent Current LDO Regulator, 3.3V, 250 mA, Vin<=16V` for `MCP1703A-3302/MB`

Best,
JAN

piotr-wiszowaty:
Done.

jkriege2:
OK, thanks, so what's left is setting the footprint to `TO_SOT_Packages_SMD:SOT89-3_Housing` (+set the footprint-field invisible!)

Best,
JAN

piotr-wiszowaty:
And done.

jkriege2:
Here also one final thing: The FPFilter should be `SOT?89*` ... The package you use is calle `SOT89...` which would not match the filter ...

Best & Thanks,
JAN

PR #677, opened 2016-09-04
xilinx: add XC9572XL-TQ100 symbol

piotr-wiszowaty:


jkriege2:
Hi!

First of all: Did you note that you deleted the description of XC2S64A-XQFG48 in the cmp-file? Could you fix that?

About the symbol:
1. please make change the background color to match KLC
2. please add lirary name to footprint field and hide it, set footprint filter as usual ,-) `TQFP*14x14mm*Pitch0.5mm*`
3. some pins (probably NC) are missing: 
```
Pin 2 is missing
Pin 7 is missing
Pin 19 is missing
Pin 24 is missing
Pin 34 is missing
Pin 43 is missing
Pin 46 is missing
Pin 73 is missing
Pin 80 is missing
```

Best,
JAN

piotr-wiszowaty:
Done. I wasn't aware that my commit deleted XC2S64A. Might that be it 
was deleted automatically
because the name in xilinx.dcm didn't match the one in xilinx.lib file?

PR #679, opened 2016-09-04
device: add Crystal_Small_4pin symbol

piotr-wiszowaty:
This is for keeping pin numbering consistent with manufacturers'
specifications e.g. http://www.abracon.com/Resonators/abs25.pdf.

This symbol is a copy of Crystal_Small.
checklib.py output (not sure whether/how this should be resolved):
library: library/device.lib
checking component: Crystal_Small
  Violating Rule 3.2
    For black-box symbols, pins have a length of 100mils. Large pin numbers can be accomodated by
    incrementing the width in steps of 50mil.
  Violating Rule 3.6
  Violating Rule 3.8


jkriege2:
HI!  I'm currently updating a lot of stuff in device.lib ... do the new 4-pin crystals there match your needs? --> https://github.com/KiCad/kicad-library/pull/742 :
crystals


Best,
JAN

piotr-wiszowaty:
No, but a "Crystal_Small_14" - would.

- Piotr

jkriege2:
Then please have a second look ... I added SMALL-variants ... but removed the Crystal_14/13, as the Crystal_GND24/23 have the same pinout ... OK for your application?

piotr-wiszowaty:
Still no. I've deliberately omitted not connected (as in the referenced 
datasheet) pins in my symbol.

jkriege2:
The pins are there on the package, but are NC ... You can interpret the pins 2 and 3 in `Crystal_GND23(_Small)` as connected to the package (drawn as U-shaped "can" ... they are explicitly not a POWER_INPUT, so just add the NC-cross-label to them and it should be fine ... otherwise we have symbols with 2 pins that have to be matched to a package with 4 pins ... so what happens to the pins that do not appear in the schamtic symbol? It's better to have them there (they actually exist) and mark them as NC explicitly, so it's clear they are not merely forgotton ... What do you think? Where is the problem with your application? Which solution do you think would help more people (by incorporating it in the lib? Yours? Min? Both?)

Best,
JAN

piotr-wiszowaty:
I prefer not to have not-connected pins (for crystals as referenced). 
The "both" variant is the best.

diggit:
Package have 4 numbered pins. You should use Crystal_GND23 or create symbol with hidden NC pins. Never omit pins.

piotr-wiszowaty:
Right. Added.

jkriege2:
Hi could you take part in the discussion ate this PR: https://github.com/KiCad/kicad-library/pull/742

Best,
JAN

PR #685, opened 2016-09-06
regul: add MCP1700TT

ChristianHirsch:
250mA Low Quiscent Current LDO


SchrodingersGat:
Hi @ChristianHirsch I welcome your desire to contribute!

Firstly, have you seen the kicad-library-utils (https://github.com/KiCad/kicad-library-utils)? These scripts perform automatic checks against many of the KLC rules so it can save a lot of time:

That aside, here are a few notes regarding this symbol:

* Pin layout should match function: Generally speaking, inputs should be on the left, (VI) outputs on the right (VO) and GND pins on the bottom (GND).
* Part name should not be duplicated in description: Please remove "MCP1700TT" from the description field
* Description: As there are other options in this series it is helpful to add these data to the description. e.g. "250mA Low Quiscent Current LDO, 3.0V output, SOT-23-3 package"
* Keywords: Other helpful keywords such could be added e.g. linear ldo
* Footprint filter: The /TT part is in the SOT-23-3 package but you have set the footprint filter to SOT-89\* This should be changed to "SOT-23" to match the correct footprints.
* Visual style. I realise that you have copied this from the existing MCP17 parts, however these are somewhat out of date and don't meet KLC requirements. The changes required here are: Set the fill of the rectangle to "background" and set the width of the rectangle outline to 0.01"

Here's an example of most of the points above:

eg

I hope this all helps. Please let me know if you have any other questions.

ChristianHirsch:
Hi @SchrodingersGat ,

thanks for your information! I applied all your suggestions to this element. But for the other pull requests I made, I need to check whether everything is fine. I will also apply your suggestions to those.

Best

SchrodingersGat:
Great work, thanks

PR #690, opened 2016-09-11
added library with 110V/220V AC power connector symbols

jkriege2:


jkriege2:
cleaned-up version is available here: https://github.com/KiCad/kicad-library/pull/718

PR #694, opened 2016-09-12
Changed install path on osx

xzcvczx:
Currently prefix is set to / and KICAD_DATA set to Library/Application Support/kicad This does not match the main kicad repo nor does it match any logical behaviour.

It makes more sense to remove KICAD_DATA as it can not be empty and just put the stuff in the prefix dir supplied. 

The only other sane option would be to have the prefix as /Library/Application Support with kicad_data as kicad/ but it doesn't seem much better.

This may need to be discussed or coordinated with adam prior to merge due to him having to change a line or 2 in his build scripts.


reportingsjr:
```
Please fix these issues:
```

adamwolf:
I am OK with this change happening.

xzcvczx:
any comments?

SchrodingersGat:
Merged after discussion on IRC

PR #701, opened 2016-09-16
added 3d models for resistors (RM2.5/5 vertical and RM7.5 horicontal)

jkriege2:
rm25
rm5
rm75


jkriege2:
cleaned-up version is available here: https://github.com/KiCad/kicad-library/pull/723

PR #703, opened 2016-09-18
Add motors library

jkriege2:
this lib contains diverse motor symbols: AC/DC motors, fans (2,3,4-pin), servo motors, stepper motors (bipolar/unipolar).

There is also a pull request for two additional pinhead footprints for fan connectors (as used on PC mainboards): https://github.com/KiCad/Pin_Headers.pretty/pull/15

The image below shows all available symbols, but additional aliases are defined (e.g. for model plane servos by company):

motors


jkriege2:
cleaned-up version is available here: https://github.com/KiCad/kicad-library/pull/717

PR #706, opened 2016-09-18
Add LT1492 dual-opamp symbol 

jkriege2:
http://cds.linear.com/docs/en/datasheet/14923f.pdf


jkriege2:
cleaned up version is available here: https://github.com/KiCad/kicad-library/pull/712

PR #708, opened 2016-09-20
Added PIC18(L)F46K22-I/PT symbol ... and more components (AD5980 DAC, MCP3428 ADC, AP130 LDO, LM4030 Vref)

neosoyo:
For TQFP  44 pins footprint only


jkriege2:
Hi!

thanks for the contribution. We have some konventions that new symbols should follow: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention.

In order to conform with them, could you please change/check the following:
1. PIC:
  1. set the correct package `Housings_QFP:TQFP-44_10x10mm_Pitch0.8mm`
  2. set package name to invisible
  3. the pins of port RB0..7 are not on a 100mil raster
2. AD5689:
  1. Pins are again not on a 100mil raster
  2. outline should be 0.01" thick
  3. maybe make a nicer symbol that reflects the function (not necessary), as e.g. for AD7171
3. MCP3428:
  1. Pins are again not on a 100mil raster
  2. outline should be 0.01" thick
4. AP130:
  1. please set correct name, as AP130XXX may have different packages. Your verison should be AP130-XXYR ... maybe you could also add the variants with the different volatges as ALIASES with accordingly changed description? See table on p. 8 of datasheet
5. LM4030:
  1. outline should be 0.01" thick
  2. symbol is a bit messy ... maybe make it bigger? Or maybe make it a bit more like TL431?
  3. what does the pin nc_GND mean? Is it NC or GND?
  4. please move to reference.lib


There are some scripts that help you check whether your components conforms to the KLC (https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention): https://github.com/KiCad/kicad-library-utils

Best,
JAN

neosoyo:
Okay Jan! Sorry about the errors, i'm moving to KiCad now (eagleCAD before)
and i'm drawing again my own components in use at work, so is a little
sloppy because the lack of time to do it :(

I will correct my contributions, i've already made others components too.

The LM4030 datasheet says that the pin NC can be connected to GND as well,
that's why i put both names.. any suggestion how to proceed?

2016-10-13 15:51 GMT-03:00 Jan W. Krieger <notifications@github.com>:

jkriege2:
No need to be sorry ;-) ... any contributions are welcome. If you submit more components: I think it would make it easier to group them a bit more, say e.g. by library ... then reviewing (and giving hints) is easier ;-)

Yeah that GND_NC thingy is weird ... ONSemi says "This pin must be left floating or connected to V−." (see http://www.onsemi.com/pub_link/Collateral/LM4041-D.PDF) ... TI draws the device with the pin described as NC (see http://www.ti.com/diagrams/rd/schematic_pmp10596_20150128111242.jpg) ... here (https://cdn-shop.adafruit.com/datasheets/2200techdeets.png) they have a line, but no pin ... so maybe I would do the symbol as the last one (diode up, no box around, draw a line and add an invisible NC pin, leave out the other two NC pins ... they are really not necessary in my opinion? ...This way if you don't know about the pin you'll leave it floating and everything is fine. If you do know about the pin and want it tied to V- for any reason, you can explicitly unhide it ... and if you do it wrong it's the user's falt ;-) ... Anybody else ( @pointhi @SchrodingersGat @...) any suggestions?

Best,
JAN

CarlPoirier:
There are some other components in kicad-library having NC pins marked as invisible, I believe.

hackscribble:
Hi @neosoyo 

Many thanks

PR #711, opened 2016-09-25
modernize microchip ENC28J60 footprint fields. add pad pin to match QFN-28-1EP footprint

ihartwig:


jkriege2:
Hi!

we have some guidelines (KLC, see https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) on how to deal with the footprints: 
1. If the device has onely one package, set the footprint field to the exact footprint, including the lib, e.g. `Housings_SSOP:SSOP-28_5.3x10.2mm_Pitch0.65mm`
2. The footprint filters should in the above case be set as follows: `SSOP*5.3x10.2mm*Pitch0.65mm*`. This way you can find any MATCHING variant of the basic footprint (e.g. with a `_HandSoldering` part in the fp-name).

Could you change it like that?

Thanks,
JAN

ihartwig:
I agree with you that the footprint filters should have wildcards and have corrected this. However, I don't agree that the number of pins in the footprint should be wildcarded or that this is specified in the KLC (which I have read). It seems like that part of the KLC was written before footprint filters were in place since your second bullet makes sense but is not present. I can help update that too.

diggit:
@ihartwig I re-read KLC again and FP filters are not mentioned there. We match only package type, dimension and pitch. Pin count filter is separated in kicad. That filter really counts pins.

ihartwig:
In lieu of having a spec in the KLC I can see that the precedent is to only use the footprint filters to select possible package types (SSOP*, PDIP*, QFN*). Honestly, this is not helping anybody. It is obvious when you're running cvpcb what general bucket of footprint you're looking for but not always clear if you want the 0.65mm vs. 0.5mm pitch or 28-1EP vs. 30(28). We have a good tool in footprint filters to clear this confusion if we choose to use it.

jkriege2:
So how do we proceed?
I would still propose this:
- set the footprint field (which you completely removed now) to the default package ... the user can still select another one by the filters
- set the FPfilter to something like `SSOP?28*5.3x10.2mm*Pitch0.65mm*` instead of `SSOP-28_5.3x10.2mm_Pitch0.65mm*` ... This way at least the combining `-_` are wildcarded so when makign a new package that does not exactly follow the required nomenclature or has something in between, it is still found.

I would also wildcard the pin-number (there is the pin number filter!), but don't have too many problems if it is still there ... it's kind of how this seems to be handled at the moment ...

Best,
JAN

@pointhi @Schlumpf @SchrodingersGat @diggit : What do you think on that topic?

ihartwig:
I have applied a strict interpretation of the KLC that we seem keen on enforcing. Since there is only one obvious footprint for each part variant I filled the full path in the field. To maintain visual compatibility I used text to show the footprint visually in the middle of the part.

Let's focus on the issues that block this part from being used in a new design as is: the existing values in the footprint fields don't match anything in our 4.0+ libraries and the QFN variant is missing the ground pad pin in schematic.

Since this part now has none of the advantages we wanted from applying either set of footprint filters I would encourage having a separate discussion about how best to codify footprint filters in the KLC. I previously highlighted how the way we match now allows picking footprints with the wrong pin geometry and physical size.

jkriege2:
I would still give `1EP`, the width/package size etz. in the FP-Filters. Otherwise you might be presented with the wrong variants! E.g. for DIP-packages I'm now presented with this list with not-applicable variants:
dip1
Even worse for QFN:
qfn1
I used these filter settings:
filt


> Let's focus on the issues that block this part from being used in a new design as is: the existing values in the footprint fields don't match anything in our 4.0+ libraries and the QFN variant is missing the ground pad pin in schematic.
This I don't undertsand ... what do you mean? What I noticed:
- for the SOIC-package it's `SOIC-28W_...` (with a `W`)


Otherwise, here's the output of our check scripts:
```
checking component: ENC28J60-C/SS
  Violating Rule 3.1
    pin: VSS (2), posx -100, posy -1050
    pin: VDDPLL (20), posx 100, posy 1050
    pin: VSSPLL (21), posx 100, posy -1050
    pin: VDD (28), posx -100, posy 1050
  Violating Rule 3.9
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint field 'Housings_SSOP:SSOP-28_5.3x10.2mm_Pitch0.65mm' should be set to invisible.
checking component: ENC28J60-I/ML
  Violating Rule 3.1
    pin: VDD (24), posx -100, posy 1050
    pin: VDDPLL (16), posx 100, posy 1050
    pin: VSS (26), posx -100, posy -1050
    pin: VSSPLL (17), posx 100, posy -1050
    pin: PAD (29), posx 0, posy -1050
  Violating Rule 3.9
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint field 'Housings_DFN_QFN:QFN-28-1EP_6x6mm_Pitch0.65mm' should be set to invisible.
checking component: ENC28J60-I/SO
  Violating Rule 3.1
    pin: VSS (2), posx -100, posy -1050
    pin: VDDPLL (20), posx 100, posy 1050
    pin: VSSPLL (21), posx 100, posy -1050
    pin: VDD (28), posx -100, posy 1050
  Violating Rule 3.9
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint field 'Housings_SOIC:SOIC-28_7.5x17.9mm_Pitch1.27mm' should be set to invisible.
checking component: ENC28J60-I/SP
  Violating Rule 3.1
    pin: VSS (2), posx -100, posy -1050
    pin: VDDPLL (20), posx 100, posy 1050
    pin: VSSPLL (21), posx 100, posy -1050
    pin: VDD (28), posx -100, posy 1050
  Violating Rule 3.9
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint field 'Housings_DIP:DIP-28_W7.62mm' should be set to invisible.
```

So please make sure that all pins are on a 100mil raster and that the footprint fields are set to invisible.

Best,
JAN

ihartwig:
Hi Jan,

I agree with you on how with generic (`SOIC*`, `PDIP*`...) footprint filters allow us to match nonsensical footprints. I tried to make this point 2-3 messages ago. However, the correct usage is not listed in the KLC, which asked me to follow in your initial feedback, and precedent from currently committed changes shows that we prefer generic footprint filters.

What I mean by the symbol does not match the library footprint is that there was no schematic pin for the ground pad. When the footprint library has a 28-1EP pins part the 1EP is listed as pin 29 in the schematic. The part ENC28J60-I/ML, which is specifically for QFN28-1EP, did not have a pin 29 for pad previously. This meant that any design using this schematic symbol left the pad floating, which is not the hookup specified in the datasheet.

I hid the footprint field text. Otherwise, the rules check is appreciated but it is highlighting issues that already existed in the part. I only added PAD (29), and I followed the existing layout. We should not change these pins just because they are 50 mils off grid and break any design currently using them.

ihartwig:
The SOIC variant does use the SOIC-28 Wide package. We codify this in kicad libraries as `SOIC-28_7.5x17.9mm_Pitch1.27mm`. The relevant datasheet page is reproduced below.

screen shot 2016-11-08 at 10 47 30 am

jkriege2:
Hi!

> I agree with you on how with generic (SOIC*, PDIP*...) footprint filters allow us to match nonsensical footprints. I tried to make this point 2-3 messages ago. However, the correct usage is not listed in the KLC, which asked me to follow in your initial feedback, and precedent from currently committed changes shows that we prefer generic footprint filters.
OK, I'm also still new as a maintainer ... I stuck to what seems to be current practice in other PRs ... and I think makes sense ;-) ... 


> What I mean by the symbol does not match the library footprint is that there was no schematic pin for the ground pad. When the footprint library has a 28-1EP pins part the 1EP is listed as pin 29 in the schematic. The part ENC28J60-I/ML, which is specifically for QFN28-1EP, did not have a pin 29 for pad previously. This meant that any design using this schematic symbol left the pad floating, which is not the hookup specified in the datasheet.
OK, but then that's fixed ... so everything OK on that!

> Otherwise, the rules check is appreciated but it is highlighting issues that already existed in the part. I only added PAD (29), and I followed the existing layout. We should not change these pins just because they are 50 mils off grid and break any design currently using them.
If we stick to that we will never get a proper library as we are not allowed to change ANY old symbol ... Every project should hvae a cahe-lib that contains the used parts, so that's not really an argument. Also we are currently doing a big makeover of `device.lib` (see #742 ) that will break even more old designs ... SO I think: If you touch a part, just make it follow the rules as good as possible, so we arrive at a modern, coherent  and  reliable lib!

Best,
JAN

hackscribble:
Hi @ihartwig 

It has been a while since Jan's last comments.  Since then, `microchip.lib` has changed.

If you are still interested in contributing, please rebase this pull request or open a new one in synch with the current master branch, and check your contribution against the latest KiCad Library Convention here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention.

Many thanks

hackscribble:
Closing inactive PR.

ihartwig:
We spent so much time arguing about how to fix a clearly broken library part that I lost interest in contributing.

diggit:
@ihartwig  I am sorry to hear this. KLC changed since then and should be more clear now. All fixes are still welcome.

PR #714, opened 2016-09-25
added MCP3204 & MCP3208 symbols

jkriege2:
datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/21298c.pdf
mcp32048


SchrodingersGat:
@jkriege2 these ones could benefit from KLC checking too. Some notes :)

* Increase outline thickness to 0.01" and background-fill the main rectangle shape
* The symbol could be a lot smaller, there's a lot of wasted space
* Change Vref / Vdd / AGND / DGND to "Power Input" type
* Change Din to "Input" type
* Change the graphical style of CS to "normal" pin and invert the name by renaming it to ~CS
* Center the symbol
* Change all text size to 50mils (down from 60mils)
* Improve the keywords (currently the same as the description)
* Remove text "(tm)" from the description

e.g.

untitled

jkriege2:
Hi!

yes that one is still on my list ;-) I'll try and make the changes you suggested (maybe today, evening). For sure I can squeeze out a bit of space for the 4-input ADC ... and maybe some more for both.

I'll also look at all my other PRs, but that will take some time. So if you look at one of them, maybe it's good to wait until I add a commit that says that they conform to KLC ... otherwise you can be quite sure I didn't touch that one with your hints in mind, yet.

BTW: There are not too many ICs that Show how they are set up on the inside, but especially for ADCs with Input MUX, I think that's nice to have to make schematics more readable (I can easily see that there's a single ADC+a MUX, which is otherwise often hidden by a simple rectangle. Is there any Standard, how you want such things in the Lib? Maybe two symbols (one with rect only, one with internal schematic)?

SchrodingersGat:
No rush, it's great that you're willing to contribute :)

I like the artwork for these, I think it does improve readability for the
symbols. I don't think it should be much more complicated than it is,
though.

There's no "standard" for such aesthetic eye-candy, as such.

I also wouldn't suggest having two different symbols - it's needless
duplication and would also be confusing.

jkriege2:
OK, I made all changes. How do you like them now?

Cheers,
JAN

default

jkriege2:
@SchrodingersGat  : did some more changes OK to merge now?

Best,
JAN

SchrodingersGat:
## MCP3204

* Add a footprint filter for the TSSOP version
* Add the two NC pins (type = NC, set to invisible)

Other than that, very good :)

jkriege2:
Done! ... should be mergeable!

SchrodingersGat:
Great work, thanks :)

PR #716, opened 2016-09-25
Add missing MOSFET transistors

jkriege2:
IRF3205 (http://www.irf.com/product-info/datasheets/data/irf3205.pdf)
IRF4905 (http://www.infineon.com/dgdl/irf4905.pdf?fileId=5546d462533600a4015355e32165197c)
IRLZ44N (http://www.irf.com/product-info/datasheets/data/irlz44n.pdf)


jkriege2:
now the symbols conform to lib styles, but they violate KLC rules 3.1 and 3.9 (I didn't find 3.9 in the KLC), but then most other symbols violate the same rules ... any idea what to do?

jkriege2:
added even more MOSFETS:
2N7000
BS250
BS108

jkriege2:
I noted that the MOSFET symbols in the transistors.lib (which I copied) do not conform to KLC, as the pins are not on a 100mil raster, but the gate pin is on a 50mil raster ... ist this OK, or should the symbol be changed?

best,
JAN

Schlumpf:
Some old parts in the libs are not KLC conform yet. Because they were created before the KLC was written. In general they should be updated over the time. Feel free to fix those parts.

jkriege2:
So are there any Standard symbols for P/N-channel MOSFETS that already conform to KLC? Otherwise I will maybe create some. What about the symbols from device.lib (i.e. Q_NMOS_*, Q_PMOS_*)?

Best,
JAN

jkriege2:
Dear all (@SchrodingersGat, @Schlumpf , ...),
I looked around a little bit in the libs and the current situation is as follows:

mosfets

So the transistors.lib uses non-KLC symbols and distinguishes between high-power (Var1) and low-power (Var2) types by "cooling dashes" (Var1). The device.lib has MOSFET symbols that conform to KLC, but do not have the typical substrate diode (Var3, although I would be willing to contribute symbol variants with the diode, as shown in Var4).

Also the transistors.lib shows the SDG labels and the device.lib does not. Is there any consensus of how one would like to have those symbols?

Best,
JAN

jkriege2:
If you want to look at the proposed new symbols, I created a private branch for them: https://github.com/jkriege2/kicad-library/tree/improve_MOS_symbols_device.lib

jkriege2:
I proposed to put Var4 into `device.lib`
See: https://github.com/KiCad/kicad-library/pull/742

JAN

jkriege2:
I am closing this and reopen it cleaned up here: https://github.com/KiCad/kicad-library/pull/780
JAN

PR #717, opened 2016-09-25
added library with motor symbols lib contains AC/DC motors, servos, stepper motors

jkriege2:
this lib contains diverse motor symbols: AC/DC motors, fans (2,3,4-pin), servo motors, stepper motors (bipolar/unipolar).

There is also a pull request for two additional pinhead footprints for fan connectors (as used on PC mainboards): https://github.com/KiCad/Pin_Headers.pretty/pull/15

The image below shows all available symbols, but additional aliases are defined (e.g. for model plane servos by company):

motors


jkriege2:
conforms to all KLC rules, except 3.8 now, but I can't figure out why ... the descriptions, keywords (and partly datasheet-links) are all there ... Any help?
JAN

SchrodingersGat:
@jkriege2  this is a great contribution. A couple of suggestions.

* Try centering the components at (0,0) as far as is practical
* It'd be great to show the pin numbers where they are currently invisible
* The Fan_4Pin symbol is a bit messy where the pin text covers the symbol itself, might want to fix this one

Thanks!

jkriege2:
improved as suggested,

Cheers,
JAN

SchrodingersGat:
I've worked out why the KLC 3.8 check is failing.

For some reason, in the motors.lib file, you have a ~ character at the front of each name

e.g.

```
#
# Stepper_Motor_unipolar_5pin
#
DEF ~Stepper_Motor_unipolar_5pin M 0 0 Y N 1 F N
F0 "M" 150 100 50 H V L CNN
F1 "Stepper_Motor_unipolar_5pin" 150 50 50 H I L TNN
...
...
```

Note after the *DEF* statement the `~` character.

Please remove this from each DEF line, thanks.

jkriege2:
I did the change you mentioned with a text editor. Now the lib conforms to KLC, except the components that do not have datasheets (they do not make sense for such general components).

Interestingly, the "~" is restored, when I open and save the lib in the library editor ... I looked at the documentation of the .lib-format on the webpage, but didn't find any mention of the ~ after DEF. and in the source tree I couldn't make out a file that implements the .lib reader/write ...

Do you have any idea, what the ~ means?

Best,
JAN

SchrodingersGat:
That's odd - it doesn't seem to be doing it for your other PRs. 

@CarlPoirier  @pointhi do you guys have any idea why there would be a leading `~` character being inserted into this library?

jkriege2:
added an IEC-60617 conformant fan symbol 
fan_iec

jkriege2:
OK; I also figured out the ~-Problem: You get that behaviour, when you hide the component name, but display the label ... now all names are visible and the problem is solved.

In addition I changed the style a bit (thick borders of the motor circle/servo boxes and renamed the symbols from FAN... to Fan... and MOTOR... to Motor...

I also don't get any KLC issues from the script anymore, except for the IEC fan, where it wants to move the labels to the left of the comp (--enable-extra option), but I feel it looks better the way it is shown above ... What do you think?

OK to merge now?

Here are some screenshots of the changed style:
fan2
motor_ac
servo
stepper

SchrodingersGat:
Yeah I think this is pretty good. Great work on this :)

PR #718, opened 2016-09-25
added library with 110V/220V AC power connector symbols

jkriege2:
Symbols conform to KLC (except 3.8 but I can't figrue out why not???) now and were made a bit smaller (still they are quite large ...).

## current style
2017-06-04 12_54_29-eeschema test_addlin _ d__kicad_test_addlin

## style when starting this PR:
wall


jkriege2:
Symbols conform to KLC (except 3.8 but I can't fugrue out why not???) now and were made a bit smaller (still they are quite large ...).

![Uploading 2016-09-27 21_17_53-[ kicad_lib_test _] (D__hardware_KiCAD).png…]()

poeschlr:
@jkriege2 could you take a look at the Travis output? I think there is a lot to do until we can merge this.

jkriege2:
Hi @poeschlr 
nearly forgot about this PR (was one of my first ones ;-) ... and I'm currently still not really srue whether it is good as it is ... Would you like to have symbols like these (I used them for a specific case where I wanted to make clear that something is in between on a 220V cable) ... or would sou use simple screw terminal or pin-head symbols?

@SchrodingersGat @pointhi @CarlPoirier Ideas on this?

Best,
JAN

PS: Maybe I will rework the symbols a bit later ...

jkriege2:
Hi all!

I reworked the symbols:
- There are more general symbols for wall plugs/spckets now and I removed the flashes (will add them in a second PR to a `graphical_symbols.lib`
- All plugs have been renamed to match better the current lib style
- I kept the old "country-specific" symbols (they were based on actual german Schuko/"Schutzkontakt" connectors/wall sockets)
- pin names are `AC_P` (for phase/hot) and `AC_N` (for neutral) + (optionally) `EARTH` now
- EARTH is always PASSIVE now
- `AC_P` and `AC_N` are power-input for sockets and  power-outputs for plugs

2017-06-04 12_44_04-eeschema test_addlin _ d__kicad_test_addlin


So questions:
1. What do you think? 
2. Do we need those at all? IMHO the MALE/FEMALE connector symbols may also easily be used (although they do not have the specific pin-naming, which might come in handy)!
3. Should they be moved to `conn.lib`
4. If we keep the symbols, so we use the "general" symbols, and/or do we do a series of country-specific symbols?

Best,
JAN

jkriege2:
- Note 1: Flash symbols will be available here: https://github.com/KiCad/kicad-library/pull/1283
- Note 2: Travis errors are due to a bug (https://github.com/KiCad/kicad-library-utils/issues/120) in the script (on server) ... The docu is actually there and locally the symbols check out OK (except FPFilters, which we priobably don't need for such general symbols)!:

```
Checking symbol 'Conn_WallPlug':
    No footprint filters defined
Checking symbol 'Conn_WallPlug_ALT':
    No footprint filters defined
Checking symbol 'Conn_WallPlug_ALT_NoEarth':
    No footprint filters defined
Checking symbol 'Conn_WallPlug_NoEarth':
    No footprint filters defined
Checking symbol 'Conn_WallSocket':
    No footprint filters defined
Checking symbol 'Conn_WallSocket_DE':
    No footprint filters defined
Checking symbol 'Conn_WallSocket_DE_NoEarth':
    No footprint filters defined
Checking symbol 'Conn_WallSocket_NoEarth':
    No footprint filters defined
```

SchrodingersGat:
Personally I think the naming is clearer if you have the default option *without* Earth. e.g.

* `Conn_WallSocket`
* `Conn_WallSocket_Earth`

jkriege2:
renamed ;-) ... commenty on my other points?

SchrodingersGat:
I have found the source of the KLC bug. 

It appears that somehow the files have been polluted with '\r' characters (might be the Windows / Linux differences again). Submitting a PR to kicad-library-utils now.

SchrodingersGat:
Regarding the PR itself:

a) I think that there are few enough symbols that they should be added to conn.lib
b) Only a single style is required, I think. Four symbols in total:
* WallPlug
* WallPlug_Earth
* WallSocket
* WallSocket_Earth
c) Please use J? for jack and P? for plug as per [RefDes standard](https://en.wikipedia.org/wiki/Reference_designator)

poeschlr:
@jkriege2 to answer your questions:
I like the general idea of these symbols. Especially the once that look similar to your original symbols because they make it very easy to see what is meant here. Even for someone who does not really have an education in electronics.
(Not sure if i will ever use such symbols but i can see the usecase behind them. That's why i poked this pull request.)

We have some specialized connector symbols already in the library. If i remember correctly you created or at least changed symbols for din connectors. I see no reason why we should not have specialized connector symbols for other types of connectors.

jkriege2:
changes done as requested by @SchrodingersGat . 

Can be merged?

Best,
JAN

SchrodingersGat:
The "plug" side should be P? rather than J? otherwise looks good to me :)

jkriege2:
@SchrodingersGat DONE!

SchrodingersGat:
Thanks Jan :)

PR #719, opened 2016-09-25
Add screw terminals to conn

jkriege2:
added screw terminal symbols to conn library


jkriege2:
improved symbols, fixed some bugs

symbols conform to KLC, except rule 3.8 ... and again I don't know why???

screw

diggit:
checklib.py tells you what is wrong...
```
checklib.py -vvv --enable-extra -c "screw_terminal_4x01"  library/conn.lib 
library: library/conn.lib
checking component: screw_terminal_4x01
  Violating Rule 3.8
    missing datasheet, please provide a datasheet link if it isn't a generic component
```
This is generic component, so not problem here.

FIX:
- Outline should be 0.01" thick and background set to fill with BG color

*I dud just a quick look, not inspected in KiCAD*

jkriege2:
fixed the thickness issue and the back color is yellow now:

screw

diggit:
After further thinking, I am not sure if we need dedicated symbol for screw terminals. Generic symbols for row connectors should be enough. (CONN_??X??)
What do you think @SchrodingersGat  @pointhi ...?

jkriege2:
hmmm ... I made them for schematics where I have some I/Os especially on screw terminals (e.g. motor-IO, power-IO, speakers etz.) and others in simple pinheaders (e.g. AVR-programming interface, ...) In such cases I think it's nice to make clear which one is which in the schematic. You would maybe also use a special sybol for power-jacks, or BNC-connectors, although a 1x02, 1x03 header does the same job ...

What do you think?

Best,
JAN

SchrodingersGat:
Hmm, I like the symbols, and I think that screw terminals are used widely enough to perhaps warrant some differentiation. Maybe a simpler name like "Terminal_01x05" (just as an example?) 

Let's maybe wait for some more consensus on this, @CarlPoirier and @pointhi should weigh in also.

diggit:
This would add symbol for screw terminals. Then what about symbols for spring terminals - that is mater of FP, then isn't this too?

Yeah, symbols look great, but IMO all row connectors with identical pins can be well represented by existing generic CONN_??x??.
I'd really like to see more opinions. It can take some time.

El3ctr0nic:
I think it is good idea to add those symbols. 
When using  standard conn we didn't know type of connector is used.
Standard conn lib is great for all types but using this schematic won't be so monolitic.
I vote 3x Yes :)
Regards,
Marcin

pointhi:
I also like the idea of additional schematic symbols for screw terminals.

SchrodingersGat:
@jkriege2 I think we have a consensus. If you can make the following changes, I can merge (unless I spot something else in the meantime ;) )

* Please change the REFDES (reference designator) from "H" to "J" [Wikipedia](https://en.wikipedia.org/wiki/Reference_designator)
* Set the name for each connector to visible
* Change the name format to Screw_Terminal_1x**YY** where **YY** is the number of connections (and has a leading zero for numbers 1-9)
* Move the name so that it is like the CONN parts already in the library

st

jkriege2:
can do the changes this evening or during the rest of the week.

JAN

jkriege2:
OK, should be OK. Please have a second look and merge if OK ;-)

SchrodingersGat:
Great work, thanks :)

PR #725, opened 2016-09-26
Add 3d models fan pinheader (3-pin/4pin as in standard PCs for chassis/CPU fans)

jkriege2:
- see http://www.formfactors.org/developer%5Cspecs%5Crev1_2_public.pdf for spec
- footprints: https://github.com/KiCad/Connectors.pretty/pull/62

2017-02-25 18_13_21-3d viewer


jkriege2:
I improved the 3D models, so they are nearer to the Intel spec

fanpin3
fanpin4

diggit:
Aren't these some common connectors with lock? I doubt, that they are Intel specific.

jkriege2:
yes, they may be ... sooo?
I'm putting together a motors/fans lib (https://github.com/KiCad/kicad-library/pull/717) and there I need them specifically for fans ... so here are the matching footprints/3D models ... and the Intel doc is the one I found ;-)

best,
JAN

jkriege2:
@diggit : As for the TO-18 packages (see https://github.com/KiCad/kicad-library/pull/734) I also added the STEP models here ... + I changed the pin color to the a bit prettier GOLD. Can be merged?

best,
JAN

matyro:
I added the 4 pin header in pull request #586 maybe you want to take a look there as well

SchrodingersGat:
@jkriege2 this has been sitting here for a while - is this OK to merge in your opinion?

jkriege2:
@SchrodingersGat Should we merge this?

easyw:
@jkriege2 
would it be possible to export the STEP models as a single object fused, with their names as in the file names?
Moreover if you use StepUp to export those FCStd models, you can add material properties also to WRL files...

jkriege2:
I have several 3D models taht I want to covert (with correct scaling then), but no time at the moment ... so it would be great o merge these for now, so they go with the old .prettys and I'll add the new 3D models for the new repo?

Best,
JAN

PR #728, opened 2016-09-28
conn.lib: add SIM card socket component

zgyarmati:
Schematic symbol for adding SIM card. See attached img for quick review.

screenshot_20160928_235048


zgyarmati:
In the meantime i stumbled upon the checklib.py script which uncovered some issues, will send an other PR with the improved version, sorry for the noise.

PR #729, opened 2016-09-29
conn.lib: add SIM card socket components

zgyarmati:
This commit adds 2 SIM card socket symbols: one without SIM card sensing switch and one with.

Checklib.py outputs:

```
checklib.py -v conn.lib -c SIM_Card_SW
checking component: SIM_Card_SW
  Violating Rule 3.8

checklib.py -v conn.lib -c SIM_Card
checking component: SIM_Card
  Violating Rule 3.8
```
The 3.8 violating is due to the missing .pdf docs, but IMHO it's can be ignored, as we are talking about general components here, i think it's even a bug in checklib.py.

For quick reference, find the attached screenshot
sim_screenshot


jkriege2:
Hi and thanks for the contribution!

two things:
1. could you make a clean pull request, that only contains the SIM-sockets ... and not all the other commits, as above
2. could you add invisible NC-pins for the pins that do not appear in your symbol (i.e. 4) ... or could you renumber 5,6,7 --> 4,5,6??? ... Do we already have footprints for the SIM-sockets? then you should orient on the pins named in those.

Best,
JAN

zgyarmati:
Thx for the feedback!
 Apparently  i've messed up the rebase... I'm going to send a new PR when i added  the NC pin for the pin4 (and PIN 8, as that one is also exists and NC). Until that i close this one, sorry for the noise.
 Nope, there is no upstream SIM card socket footprint yet, i made for 2 models, i'M also planning to send them in, but first i wanted to have the schematic symbol in place.

PR #733, opened 2016-10-01
Add 3d models of 6x6mm² buttons with different heights

jkriege2:
added 3D models for 6x6mm² tactile buttons (PHAP33xx) with different heights. They are all the same, but point to different 3D modules.
see http://nal.apem.com/pdf/phap33-series.pdf

phap1
phap13


diggit:
Neat! Can you tweak material properties to make plastic look like plastic and same for metal?

jkriege2:
btn

jkriege2:
- button rounded now
- materials changed
- all created from one FCStd-file


btn1
btn2
btn

jkriege2:
Can be merged?

jkriege2:
@diggit : As for the TO-18 packages (see https://github.com/KiCad/kicad-library/pull/734) I also added the STEP models here ... Can be merged?

best,
JAN

diggit:
I've spotted 2 problems:
- origin placement - FP(SW_PUSH_6mm) has it on top left pin, but your models have it on bottom left. (looking from top)
- Step models have correct scale, but wrl not. Did you change something between exports or is it behavior of kicad? (maybe step carries dimension units and kicad developers used it to fix mm/inch scaling problems)

jkriege2:
The origin I can fix, but as far as I know the dimensions are a KiCAD export problem, as there are no explicit dimensions in vrml.
JAN

diggit:
Ok then, just fix origin.

STEP models are future. I hope, that kicad nightlies are built with support turned on.

jkriege2:
OK, Origin fixed ... + footprints modified accordingly

JAN

diggit:
Did you set material properties? Metal parts does not look shiny. Freecad project says, default.
Can you make them look like plastic and metal? (steel)
Sorry that I didn't spot it earlier. I just saw some correctly set model next to this a realized it.

jkriege2:
Actually ... yes, I set it to Silver, which has a shinyness of 20% in FreeCAD ... choosing Steel makes them even darker, so for me they look OK ... the plastic parts have the properties of plastic (although Freecad resets the choosen material to default every time, but keeps the properties themselves ... Bit strange, but in all they should be fine.

jkriege2:
@SchrodingersGat @pointhi Could you have a look at this one? Would be great if it was still in the tag! It is (IMHO) complete!

If you do, please also merge: https://github.com/KiCad/Buttons_Switches_THT.pretty/pull/10

best,
JAN

SchrodingersGat:
Mergerated

PR #735, opened 2016-10-02
added TLC5940

jkriege2:
- see http://www.ti.com/lit/ds/symlink/tlc5940.pdf
- two symbols: one for DIP and one for HTSSOP-package (see this PR: https://github.com/KiCad/Housings_SSOP.pretty) as pinouts are different!
- both conform to KLC!

tlc5490
tlc5490_tssop


diggit:
Hi, yes symbols passes checklib.py test with no KLC violations, but that does not mean, there are 100% KLC compliant.

- DIP version is in datasheet, but not available, remove it unless you find it in production (I ignore it for now)
- Use 0.01" thick outline
- rename HTSSOP version to TLC5940PWP
- set correct FP filter `HTSSOP*4.4x9.7mm*Pitch0.65mm*`

jkriege2:
I made the requested changes ... although No. 1 I don't understand: I have ~1 pieces of that in DIP lying around somewhere here (and actually that's the reason why I want that symbol!) ... don't knwo whether they are still in production, but they are widely used for e.g. 3D LED-cube projects ... and most people seem to have it in DIP package!

Best,
JAN

diggit:
Interesting, TI does not sell DIP version anymore. Anyway I found correct name for DIP, it is **TLC5940NT**, so please rename it.
Make all those symbols to have same pins on same position (no SPI block moving), place THERMPAD on bottom and you can shorten it to PAD only.
I'll review it tomorrow, hopefully.

jkriege2:
Done ;-)

jkriege2:
@diggit : Can this be merged? Anything missing?

PS: If you have time: I posted a general question on editing libs: https://github.com/KiCad/kicad-library/issues/741

diggit:
Needed to fix:
- In f9128399ecbdefaa7ec5b1f82abcfdbe53f4f29c you somehow put letter **F** on the beginning of file which renders it unreadable for KiCAD library parser.
- pin BLANK is not inverted
- use datasheet matching pin names: DCPROG -> DCPRG, Vprog -> VPRG
- pin VPRG type -> Input

**TLC5940NT**
- assign generic FP, not modified one -> `Housings_DIP:DIP-28_W7.62mm`
- set FP filter to not match pin count -> `DIP*W7.62mm*`

jkriege2:
OK, done ...

jkriege2:
One question on the FPLIST: Are there any rules on how to fill it? Especially: Why should the pin-count be left out (yes there is a pin-number filter, but that's not always on). Let's take a simple 7400 as example ... there are no versions in a DIP-16 and DIP-20 housing, but only a DIP-14 housing, so why not mention it? This filters down the shown footprints to what really makes sense ... and if I want something completely different, I'll have to deactivate the filters by hand, but usually that shouldn't be needed, especially in the example case ...

Best and thanks for an answer/explanation,
JAN

diggit:
We omit pin count in FP filters because that information is filled in by human, it is more secure to rely on pin count filter which really counts pins.

jkriege2:
OK,
merge conflicts fixed and made minor changes ... all points should be OK for merge now.

JAN

jkriege2:
OK,
merge conflicts fixed and made minor changes ... all points should be OK for merge now.

JAN

diggit:
Only **BLANK** remained inverted and it does not act so. When BLANK=H, all outs goes of, so no inversion here...

jkriege2:
Done!

JAN

diggit:
You forgot to change Graphic style.

jkriege2:
fixed

diggit:
Nice work you did, merging.

PR #739, opened 2016-10-06
USB connectors: VCC renamed as VBUS

diegoherranz:
The standard name for USB power rail is VBUS and not VCC.


SchrodingersGat:
Thanks :)

PR #742, opened 2016-10-08
added several (many!!!) missing devices

jkriege2:
Dear all, I added all devices that I could think of and that were still missing in KiCAD + I reworked some of the existing devices:

add_missing

What do you think? Is anything missing? Do you like the style? Shoot at it ;-)

Best & have a nice weekend,
JAN


SchrodingersGat:
This is some great work - I shall try to have a better look at this soon :)

jkriege2:
Comments on KLC:
- Thermocouple: OK
- Thermistor_PTC/NTC: missing datasheet => OK (general comp.)
- R_NET: OK
- Darlingtons: violate rule 3.2, as do all NPN/PNP transistors
- UJT: missing datasheet => OK (general comp.)
- Photo-components (diode, transistor, resistor): missing datasheet => OK (general comp.)
- Meters: missing datasheet => OK (general comp.)
- Antennas: missing datasheet => OK (general comp.)
- Lamps: missing datasheet => OK (general comp.)
- ..._Small components: pins are shorter (0.05") than specified in KLC and are not on the grid (but this is true for all _Small components, e.g. C_Small, R_Small ...)
- Batteries/Solar Cells: missing datasheet => OK (general comp.)
- speakers, microphones, buzzer: missing datasheet => OK (general comp.)
- switches: missing datasheet => OK (general comp.)
- switches_Small: same problems as for other _Small components (grid/pin length)
- EMI: missing datasheet => OK (general comp.)
- Resonator: missing datasheet => OK (general comp.)
- CTRIM_DIV: same problems as C (grid/pin length)
- Resonator_Small: same problems as C (grid/pin length)

-

jkriege2:
About all those `_Small` components: They all do not conform to KLC, because their pins are never on the grid and the pins are too short (usually 0.05" instead of 0.1"). The first thing could be fixed in most cases by moving the anchor away from the cpomponent center ... but the second issue cannot be fixed with the current KLC ... how should we deal with that? Should possibly the KLC-rule be extended to say that `..._Small` components are allowed to have 0.05" long pins that may be on a 50mil grid?

Best,
JAN

maximeborges:
Nice job !
I think putting the 'R?' of the resistor on the top of it would be better (like on Eagle for example).
image
For the resistor networks, maybe putting the pin number on the left of the pin. Some numbers looks like too much collapsed to the pin and symbol, and number with two figure overlap the symbol.
image
For inductors I think having the loops on the top looks better, but that's a personal preference.
image

For the rest, it seems to be ok. Still, well done ;-)

jkriege2:
About the R-Netwerks: Which sytle/psoition for the REFDES and Value do you prefer?

CURRENT VERSION:
r_net8_current

VERSION 1:
r_net8_v1

VERSION 2:
r_net8_v2

VERSION 3:
r_net8_v3

VERSION 4:
r_net8_v4

VERSION 5:
r_net8_v5

PS: I'll change the REFDES to RN now ;-) ... just in case somebody would have complained ;-)

JAN

maximeborges:
I believe the version 2 is the better one. Never used a R-network in my projects yet but I think putting it vertically would be better, so the version 2 seems to be the better one I'm my opinion.

Schlumpf:
I prefer the com-pin top variants. It's not a footprint, so it's better to order the Pins logical.

jkriege2:
I'm also experimenting with the Ferrite Beads ... which version do you like better?

ferritebead

jkriege2:
@Schlumpf : Which version? 1 or 2 (note the different text locations!)

feichtinger:
> I'm also experimenting with the Ferrite Beads ... which version do you like better?

According to IEEE 315 this symbol should be used for ferrite beads:
[https://upload.wikimedia.org/wikipedia/commons/thumb/2/2d/Ferrite_bead_ring.svg/240px-Ferrite_bead_ring.svg.png](url)

...which is very close to the third version.

I suggest that the ferrite bead symbol should have the same pin distance (300mil) and orientation (vertical) as the standard R and C components.

PS: good job, thanks for your work! :+1:

jkriege2:
Thanks for all the feedback so far ... I'll work over the addition probably on the weekend ... Please: Give me more suggestions on what you like/don't like/should be changed ...

JAN

jkriege2:
e.g.:
1. what's about the changed/added switches?
2. Should we have two flavours of DIP-switches? If no, which one do you like better (I've seen noth alternatives in EDA-tools)?
3. Should we have two LDRs? If not, is removing the circle (not really standard I think) OK with you?
4. Phottransistors OK?
5. Batteries/Solar Cells OK this way?
6. Meters OK? Any missing? Should we have a oscilloscope symbol
7. should the 3-/4-wire symbols for NTC/PTC be in here, or better in sensors.lib? Symbols OK?
8. Microphone/Speaker lib OK? Should we maybe replace the old speaker symbol with the samller new one?
9. Are the new filter symbols OK? Anything missing?

Best, JAN

maximeborges:
1/ is the SW-SPDT_x2 
2/ For the DIP switches It hink you could maybe provide splitted multi-symbols components like this:
image
I'm more for the left one btw, using standard representation.
3/ You talk about which components ?
4/ Not really familiar with them so I will say ok
5/ Ok
6/ Is the arrow for the meters actually used anywhere ? Never seen them.
7/ Sensors.lib has only specific references of components, so there are better in device.lib imo
8/ The old speaker is really too big, so replacing it with the Speaker_Small seems good to me. Maybe adding a smaller one for the _Small since your Speaker_Small looks coherent with the rest of the normal sized components.
9/ Can't help that much for this
10/ Screenshots of the actual one ?

jkriege2:
@maximeborges 
ad 10/ They are both in the screenshot at the top ... e.g. R_NET8 and R_NET8_Small

cu&thanks for teh feedback,
JAN

CarlPoirier:
Hi Jan,

This is pure awesomeness! I shall look into it in more details soon. For now I have just glanced over the discussion and one thing I can say is about the _small components and transistors. Some rules apply only to blackbox symbols so these are not considered.

Carl

pointhi:
Because you asked, could you also improve the fuse symbols, and add one for poly switches. From what I found out they are similar to the thermistor one:  ([see here](http://electronics.stackexchange.com/questions/237437/what-is-the-schematic-symbol-for-resettable-fuses))

There is also a wiki page which could lead to some ideas which other symbols could be useful (german page): https://de.wikipedia.org/wiki/Liste_der_Schaltzeichen_(Elektrik/Elektronik)#Elektrische_Bauelemente

I'm a bit unsure if we have an symbol standards we follow. From what I see we use IEC for most parts, but I haven't seen a recommendation of it in KiCad.

Nevertheless, @jkriege2 good work :+1:

jkriege2:
OK, here are several changes ... some based on all your suggestions:
- more antennas
- improved fuses, added polyfuses
- changed Ferite_Bead
- reworked R_NETs (generated by a script now, so can easily be changed)
- added RTRIM, reworked Dula-Potentiometer
- replaced old (large) speaker symbol
- reworked resistive sensors
- removed old EMI_Filter symbols

add_missing

Please have a look ... and make more suggestions ;-)

diggit:
Awesome improvement!
Switches: It would make sense to unify them even more. Identical circles, spacing etc.
DIP: I prefer primary (left) version.

Do we really need _small variants when basic variants won't be huge anymore?

jkriege2:
OK, again an updated version:
- several new/missing diodes
- diode bridge symbols are now available for different pinouts (as e.g. dual-Schottky-diodes)
- Dual-Diodes (analogous to Dual-Schottky-Diodes)
- added memristor
- added mising transformer
- added flash lamp
- improved on switches (all look alike now in the normal and small-series, removed alternative DIP-switches)
- added mayn missing switches (temperature, Dual, ...)
- added frequency counter and oscilloscope
- all LED symbol look alike now (arrows!)

add_missing

Any more comments? Anything missing? Should we move towards merging ... for that, it would be great if all readers of this thread had a (very) close look and give comments/bugreports.

One question: For the parallel resistors in SIP-package: Should we leave the symbol as ist (pindistance=100mil, resistordistance=300mil), or should we move them closer together (resistordistance=200mil), or should we have a completely different symbol (e.g. like DIP-symbol, but with changed pin-numbering, although I would find that confusing).

cu
JAN

pointhi:
Hy @jkriege2,

what I spot on a quick look is the differences of the numeration style of some devices. For example ```SW_SWDT``` uses blue color for the pin numbers, while the (new) ```SW_SPST```uses red for them. Probably this should be unified.

pointhi:
For review, I suggest we review category wise. Kicad library maintainers should be able to edit this post.

When someone checked a category and it look fine simply check the corresponding checkmark in this comment and add your username to the corresponding line. This approach should be clearer than reviewing all at once.

* [ ] Meters
* [ ] Antennas
* [ ] Lamps
* [ ] LDRs
* [ ] Photo-Semiconductors
* [ ] Solar Cells
* [ ] Batteries
* [ ] Speakers, Microphones, Ultasound
* [ ] Switches, etc.
* [ ] Resistor Network/Packs
* [ ] Ceramic Resonator
* [ ] EMI components
* [ ] Inductors
* [ ] Transistors
* [ ] Fuses
* [ ] Capacitors
* [ ] Misc
* [ ] Semiconductors

jkriege2:
@diggit : first thanks for the comments so far ...
- switches display pin numbers now
- about the cathode: I merely copied that nomenclature from the existing components ... I think they use K (either because some germean guy made the symbols or) because that allows to distinguish the cathode (K) from the common/center pin (C) ??? Any idea, for a better nomenclature? Maybe A-node, C-athode, M-id? or B-ranch or may T-ee?
- Should we use TVS or Suppressor? I (german speaking) would have searched for suppressor ... but then you also find the TVS-symbols, as "suppress" appears in the description/keywords ... SO should we take TVS, because it's shorter and was used before?
- the real pinnumbers for bridges I don't understand ... could you give an example?

Thanks,
JAN

jkriege2:
added electromagnet + updated image ABOVE!!!
JAN

jkriege2:
- added laser diodes
JAN

Image here again:
add_missing

diggit:
@jkriege2 sorry, they are ok, I thought they have wrong color.

jkriege2:
@diggit : According to wiki (https://en.wikipedia.org/wiki/Reference_designator), the REFDES for Loudspeakers/Buzzers is LS ... so I took that ;-)

jkriege2:
OK; I also reworked the crystals still a bit:
- renamed CRYSTAL to Crystal
- added symbols for 3- and 4-pin Crystals with GND connection (not oscillators)

add_missing
(image also updated above!)

jkriege2:
I reworked the multi-colored LEDs

add_missing
(image also updated above)

jkriege2:
- changed size of LDR to match size of normal R (same as done before for NTC/PTC)
- removed alternative R_PHOTO symbol ... was non-IEC standard!


add_missing
(image also updated above)

diggit:
LEDs LED_CRGB and LED_RCBG have GC instead of C on common cathode.

jkriege2:
@diggit: thanks, fixed ;-)

jkriege2:
- added more crystal symbols (with pins1-3 and 1-4) for such devices as: http://www.abracon.com/Resonators/abs25.pdf

JAN

jkriege2:
@pointhi @diggit : Could you re-check the crystals/resonators, please? I made some more changes there .... they look more similar now and are more versatile (no longer exlicit GND pins,  but can be GND or NC or package ... or whatever ;-)

- I reworked the ceramic resonators (as small as R and R_Small now
- I added small Crystals, analogous to the large crystals
- removed Crystal_13 and Crystal_14, as they are merely the same as Crystal _GNDXY if you use the "GND" pins as NC!

crystal

Thanks,
JAN

diggit:
@jkriege2 rechecked, looks ok

@pointhi @SchrodingersGat @CarlPoirier and others...
Do we really need **_small** component versions?

jkriege2:
still found siomething missing:
- added peltier element
- added a heatsink symbol with electrical connection
- and renamed `D_Suppressor...` to `D_TVS...`

tec_heatsink

jkriege2:
Uuuppps ... somehow the resistor symbols vanished from my large sheet ... I readded it:
res

jkriege2:
also did some minor changes:
- renamed Components in ALL-UPPERCASE to Normal-Case 
- removed old/unneeded components (e.g. RVAR as we now have RTRIM, like CTRIM, LTRIM ...) 
- reworked Varistor symbol

JAN

jkriege2:
One question: Which Thermocouple should we use? Or should we keep both?

Best,
JAN

jkriege2:
I still did some minor alterations to the STYLE of the crystals/resonators: 
- the electrode lines are thicker now
- the central crystal is a bit thinner and the electrodes moved towards the center, so there is more space for the pin numbers.

crystal


JAN

jkriege2:
Should we have two different symbols for trim pots like this:
trimpot
and "normal pots"?
pot

JAN

jkriege2:
Hi all!

to make the potentiometer question a bit more concrete, I drew up some design:
pots

Which one's should we take?
1. keep the old one (I'll change the symbol a bit, as the outline of the resistor should be 0.01" thick, as for the standard R symbol)
2. use the IEC-conformat version (I think they don't have the trimmer, but I like the possibility to distinguish between a 3-pin trimmer (on the PCB) and a potentiometer (e.g. fronpanel)
3. use the nearly IEC-conformant version ... I think it's close to what is often used in germany

I would (I think) vote for 2 (cause of the standard), although I like the symbols 3 (personally) better.

What do you think?

Also: Please give feedback to the other open questions:
4. Which Thermocouple symbols should we use?
5. Do we really need the **_Small** components?
6. AAny idea how we can distinguish **C**athode from **C**ommon in the Names of the 2x-Diodes/Schottky-Diodes?

Best,
JAN

jkriege2:
I small note: I improved the varistor symbol with a "U":
varistor

JAN

jkriege2:
Maybe one argument pro/contra the `R_Small`: I'm just adding some instrumentation ampifiers and the symbols look like this:
ina_rsmall

So with an R_Small we can conveniently add the A-setting-resistor where the manufacturers do it in their datasheets ... What do you think?

JAN

jkriege2:
Also a question concerning this PR: https://github.com/KiCad/kicad-library/pull/679

Should we add 3/4-pin Quarz variants without the can that I added above and with additional pins hidden? Or should we only take the symbols with can and say you can NC the pins explicitly?

JAN

jkriege2:
... and please fill in @pointhi 's post above, so we can move towards merging ;-)

JAN

jkriege2:
Hi!

I've been looking a bit more into the abbreviation for cathode thing: It seems that although Cathode is written with a C, a "K" is commonly used as abbreviation: https://www.google.de/search?q=diode+schematic+symbol

So if nobody objects, I'll replace the C for Cathode with K for Cathode ... OK? THat should also solve some naming issues with 2x-diode-symbols!

cu
JAN

CarlPoirier:
@pointhi I don't see any issue in having these _small symbols.

@jkriege2 I'd go for the thermocouple without the rectangle. Have you seen the other used often?

@jkriege2 Don't care about the NC pins for now. That can be added later on if needed.

@jkriege2 K is good. I guess there could be an IC with a cathode and a collector on the same package. [This](http://www.onsemi.com/pub_link/Collateral/NTLJF4156N-D.PDF) is almost the case.

jkriege2:
@CarlPoirier 
- OK, I removed the blocks ... didn't find too many schematics with them on Google ... so we only need to decide between Thermocouple and Thermocouple_ALT
- I went back from C to K for Cathode, which solves all naming issues also. (double-TVS devices)
- I would also say we leave crystals selection as is.

Best,
JAN

PS: What do you think on the potentiometers?

jkriege2:
BTW: Given the examples above I would also vote for keeping the `_Small` devices ...

samdolt:
Great work :)

Personnally, I use the  _Small components more often than the normal one.

#### Historical fact:

The first _device.lib_ clean-up was started upon issue #97.

The _*_Small_ component and the first device.lib clean-up was  merged in #104 and in the same time the KLC was updated to allow such components

> Revision 0.10, March 1st 2015
> 1. Moved the footprint value property to the fabrication layer
> 2. Pin length is enforced only for black-box symbols
> 3. Some reorganization has been done.

diggit:
Ok, small component variants of R, C, X are reasonable, but I can't see reason to have small variants of switches.

CarlPoirier:
Denser schematics?

jkriege2:
So, what about simply removing the large switches and only using the small ones (renamed without the small)?

JAN

jkriege2:
@diggit raised the question, which lamp symbols to use ... Do you think we need the _ALT symbol?
lamp2

It can be found via google: https://www.google.de/search?q=lamp+schematic+symbol&num=20&tbm=isch&tbo=u&source=univ&sa=X&ved=0ahUKEwj75Pbhi-_PAhXHlCwKHfBHAMMQsAQIHQ&biw=1543&bih=832

But I think the symbol with the X should be enough ... whatd o you think?
JAN

jkriege2:
@diggit: The C_Feedthrough issue is fixed

jkriege2:
@diggit:
- "not sure if needed D_Temperatur_Dependent_90deg": I made it, because the vartheta looks weird, if rotated ... 
- "DIAC vertical lines thickness inconsistent": What do you mean? To me they look the same as other diodes ...
- POT: outer line thicknes will be fixed in final symbols ... please vote on my proposals above!

JAN

jkriege2:
@diggit: numbering fixed

diggit:
> "DIAC vertical lines thickness inconsistent": What do you mean? To me they look the same as other diodes ...



AFAIK only POT question remains.

Thermocouples with block should remain, it is different if you have plain thermocouple or using it with isothermic block. Never seen non-alt version...

diggit:
> "not sure if needed D_Temperatur_Dependent_90deg": I made it, because the vartheta looks weird, if rotated ...

ok

> But I think the symbol with the X should be enough ... whatd o you think?

Yeah, X symbol is enough.

> AFAIK only POT question remains.
 I vote for full IEC version. "nearly IEC" symbols are specific for rheostats only.

CarlPoirier:
> Thermocouples with block should remain, it is different if you have plain thermocouple or using it with isothermic block. Never seen non-alt version...

Interesting to know, it's a good reason to have both.

> AFAIK only POT question remains.

I read on wikipedia that there are rheostats with ratings up to several thousand watts and that the term potentiometer is more general. How about having both symbols?

diggit:
[Here is something about this topic.](http://www.resistorguide.com/trimpot/#Trimpot_symbols)

We can have both. Then description must be precise what is what.

jkriege2:
OK; DIAC line-widths fixed (dodn't see it at first)!

jkriege2:
Lamp_ALT removed

jkriege2:
- OK; I added the POTs, removed Lamp_ALT, improved CTRIM, added C_Variable
- Switches: I'll do that tomorrow ...

JAN

jkriege2:
latest versioN:
add_missing

diggit:
Can you position component names of POTs to not overlap symbol?
Nice work on capacitors.

SchrodingersGat:
@jkriege2 I have added some comments to the review comment by @diggit (way up in the message history)

jkriege2:
@SchrodingersGat : I fixed most REFDES as proposed ... The speakers are all LS (see https://en.wikipedia.org/wiki/Reference_designator) ... the old SP was just a non-updated schematic!

cu
JAN

jkriege2:
OK!

all things mentioned above are fixed now ... see commit comments and here's the latest version:
add_missing

SchrodingersGat:
Ok, I am in favour of merging. Let's see what the others say. Great work :)

jkriege2:
Hi all!

I still found one thing worth changing:
- there were no MOSFET-transistors with a substrate diode (although it is commonly drawn in mayn schematics). therefore I added a series of such devices:
mosfet

- also I fixed the label positions for all transistors (now all left-aligned on the right of the device).

That should be all from my side (can't think of anything that needs fixing right now in this PR ;-)

Best & thanks for the extensive review to ALL!
JAN

So please merge, if you're OK with everything!

Here's a final image of the full lib changes:
add_missing

CarlPoirier:
Some more comments:

- You can use "Ammeter" instead of Amperemeter. It's shorter and still complete.
- For resistor packs, we once decided only use "_x8" as a suffix. See KLC 4.5. Of course this is debatable, however. What do you think? I must admit I do like the pack vs. net distinction.
- I can't see how parallel resistors in SIP package are a net. I'd go for some nomenclature like "R_Pack3_PAR" or "R_x8_Par"
- Same for switches, that would be "SW_Pack6" or "SW_x6". There is no need to add the package name in.
- Diggit mentioned a distinction could be made between thermocouples and thermocouples with isothermic block. It was suggested to keep the symbol with the block for the latter.

jkriege2:
@CarlPoirier : 
- ad 1): Hmmm ... I like the longer version better (clearer, what is meant) ... what do the others think?
- ad 2)/3): I would also keet the R_NETxx ... so what do you think on this:
  - true networks (terminator/star topology): R_NETxx_...
  - multipacks: R_PACK_xXX (maybe even R_PACK_xXX_DIP) and R_PACK_xXX_SIP
- ad 4): so you think of using SW_DIP_xXX ???
- Yes, I think there is one (https://www.omega.com/temperature/Z/pdf/z021-032.pdf), which was also my intention when initially designing both symbols, but there are other possibilities to implement such a symbol ... I made this quick sketch for two possibilities:
tc_block
  So basically the idea is to have an isothermal piece of metal, of which often the temperature is measure independently (above with a T-diode). So we would have several options:
   1. have no block symbol in the lib and create it with lines in the schematic
   2. have an isothermal block symbol without extra temperature sensor (possibly like heatsink with an additional PAD pin) and sensors should be put next to it and maybe connect with lie (see above)
   3. have separate symbols with different temperature sensors (D_Temp, NTC, PTC ... the we still ahven't covered digital sensors ...)
  I would be in favour of 1 or 2 ... but can't really decide for one, so possibly 2, so the user can choose ...

Best,
JAN

What do you think?

jkriege2:
@CarlPoirier : 
- ad 1): Hmmm ... I like the longer version better (clearer, what is meant) ... what do the others think?
- ad 2)/3): I would also keet the R_NETxx ... so what do you think on this:
  - true networks (terminator/star topology): R_NETxx_...
  - multipacks: R_PACK_xXX (maybe even R_PACK_xXX_DIP) and R_PACK_xXX_SIP
- ad 4): so you think of using SW_DIP_xXX ???
- Yes, I think there is one (https://www.omega.com/temperature/Z/pdf/z021-032.pdf), which was also my intention when initially designing both symbols, but there are other possibilities to implement such a symbol ... I made this quick sketch for two possibilities:
tc_block
  So basically the idea is to have an isothermal piece of metal, of which often the temperature is measure independently (above with a T-diode). So we would have several options:
   1. have no block symbol in the lib and create it with lines in the schematic
   2. have an isothermal block symbol without extra temperature sensor (possibly like heatsink with an additional PAD pin) and sensors should be put next to it and maybe connect with lie (see above)
   3. have separate symbols with different temperature sensors (D_Temp, NTC, PTC ... the we still ahven't covered digital sensors ...)
  I would be in favour of 1 or 2 ... but can't really decide for one, so possibly 2, so the user can choose ...


On more thing: In this PR (https://github.com/KiCad/kicad-library/pull/765) @ashtonchase proposed to have a symbol for mounting holes/screws, which would match the mounting_hole.pretty library that we have. The proposed symbol is:
mounting_hole
Should we think about incorporating such a thing here? Is device.lib the right location (maybe conn.lib would be better?) I would propose a simple symbol, such as something like this: http://www.fedevel.com/welldoneblog/wp-content/uploads/2016/07/mounting-holes.jpg

Best,
JAN

What do you think?

ashtonchase:
@jkriege2,

I'm on board with including a mounting hole symbol in this request. If so, I will remove it from my pull request.

jkriege2:
Here's a proposal from my side: 
mounting_hole2

1. What do you think?
2. Should we put it here?
  2.1 Leave it here in `device.lib`
    :+1: 
  2.2 Put it into `conn.lib`
    :+: 
  2.3 Put it (together with the Heat_Sink) into a new lib `mechanical.lib`
    :+: @jkriege2

Please vote ;-)
JAN

jkriege2:
Hi all!

still some questions open:
- MOSFETs with substrate diode
- ThermoCouple with blocks + do we need the alternative ThermoCouple
- What about proposed Mountin_Hole?
- naming of DIP-switches and R_PACKs

I have a lot of time on friday (but won't be available until mid next week after that), so if we get to a decision until then, I can still implement all changes/wishes ... and move this forward ;-) 

Best,
JAN

jkriege2:
Ohhh ... this is never ending ... also the TRIAC/Thyristor symbol are in need of an overhaul ... my proposal:
triac

jkriege2:
latest version

add_missing

jkriege2:
OK, so I made some changes (can be redone anyways ... it's a GIT ;-)
- I created `mechanical.lib` and put Heatsink(s) and new MOunting Hole(s) in there
- I added the MOSFET symbols with substrate diode
- I readded the THermocouple with Block
- I added new TRIAC/Thyristor symbols and removed the old ones.

... SO if all this is fine with you, there are only the naming issues remaing ... What do you think?


add_missing

diggit:
- AFAIK there are no MOSFETs without substrate diode.
- Thermocouple block is fine. I would not complicate it.
- R_NET and R_PACK makes sense to me
- mechanical lib is good idea
 - What do you think about some simple symbol for housing? Otherwise, there is no symbol which we can assign eg. dummy footprint with housing 3D model set.

diggit:
I noticed, there is library **switches.lib** with DIP switches. There are 2 options:
- **replace** original switches.lib with new switches from device lib
- **delete** switches.lib

SchrodingersGat:
I would suggest migrating all switches to the switches lib.

There are some switches in device.lib that require alternative pinouts, and
it's starting to get cluttered.

CarlPoirier:
@jkriege2 OK, let's keep _NETX and _PACKX and _PACKX_PAR. However, I don't see any reason to specify the package (_DIP or _SIP).

Also, I suggest you stop adding things, we discuss the remaining issues and merge this. Then you can open up another pull request. :)

jkriege2:
Interesting about the switches.lib ... for some reason I didn't see it before ... ???
Here's a comparison:

swdip

So I would simply replace the old ones (they are basically the same, except for non-KLC and the switch symbol does not match the new "normal" switches (see SPST for comparison)). I'm also in favour of moving all switches to `switches.lib`

So: About the MOSFETs? I've seen both symbols used ... I also think that all commercial MOSFETs that I've seen have the diode ... hhmmmm ... so what do we do? Keep both? Replace one? Delete one?

I also still have the housing symbols that @diggit proposed ... should I still push them?

JAN

jkriege2:
@CarlPoirier : 
ABout specifying the package (SIP/DIP), I think that's important due to the pin-numbering ... The SIP-parallel resistors have pin 1-2, 3-4, 5-6, ... and the DIP have 1-N, 2-(N-1), 3-(N-2), ... so how to distinguish them otherwise?

JAN

jkriege2:
Since nobody answered so far ... I started by moving all switches to `switches.lib` ... If that's OK with everybody, we could move on ... IMHO there's only the question of the MOSFETs open?
JAN

El3ctr0nic:
I think that diodes in MOSFET should be drawed. All have it in physical struct. When draw H bridges or 3 phase bridges it's important to have them - they work as freewheeling diodes and should be on schematic

jkriege2:
But the question remains: Do we keep both variants or only one?

JAN

El3ctr0nic:
I think only one :). Sorry long conversation i didn't read all ;/

diggit:
Keep only version with diode and remove D from name.
A have nothing else against merging.
@SchrodingersGat @Schlumpf @CarlPoirier any commets?

jkriege2:
Done!

jkriege2:
now done!

CarlPoirier:
@jkriege2 About specifying the package (SIP/DIP):

What will happen when we will want to use aother package, say BGA? We will end up using a symbol ending in _SIP but associate it with a BGA footprint. We suggested a suffix to indicate a different pin ordering without the package, see KLC 4.4. We can find a better suffix, but I don't agree in using a package name.

CarlPoirier:
Also, there are many comments left in @pointhi's post in the middle of the conversation. Before merging, I would like to see them updated/removed by their author if issue has been fixed.

As for my comments, there is one that's not been addressed yet. I was saying that I can't see how symbols in section "parallel resistors in SIP package" are a net. I'd use "R_PackX_Par" instead of "R_NetX_Par"

diggit:
@CarlPoirier
Ok, I've cleaned up [that post](https://github.com/KiCad/kicad-library/pull/742#issuecomment-254044173). All solved requests were removed, even @SchrodingersGat 's. (sorry)

Now I see what is the problem with package in name. I didn't realize it at first. Not idea about solution.

NET is confusing, presumably, it was meant as NETwork. NET -> PACK is solution here...

KarlZeilhofer:
Very Nice Proposal!

But please draw the diodes without any fill, and with a line through the body. 

Please refer to the standard EN-60617, 
Or Wikipedia: https://de.wikipedia.org/wiki/Liste_der_Schaltzeichen_(Elektrik/Elektronik)#Dioden

KarlZeilhofer:
Very Nice Proposal!

But please draw the diodes without any fill, and with a line through the body. 

Please refer to the standard EN-60617, 
Or Wikipedia: https://de.wikipedia.org/wiki/Liste_der_Schaltzeichen_(Elektrik/Elektronik)#Dioden

KarlZeilhofer:
Very Nice Proposal!

But please draw the diodes without any fill, and with a line through the body. 

Please refer to the standard EN-60617, 
Or Wikipedia: https://de.wikipedia.org/wiki/Liste_der_Schaltzeichen_(Elektrik/Elektronik)#Dioden

diggit:
Interesting, I probably never seen this diode symbol and it is IEC standard. :/

jkriege2:
@KarlZeilhofer @diggit ...
I can change the diode symbols in the coming days ...seems to make sense to me ;-)
JAN

jkriege2:
@CarlPoirier @diggit 

About the R_NET-naming ... I'm not yet sure, what a really good solution would look like, but what about this as a proposal:
- `R_NETX_PAR_YYY` -> `R_Pack_xX_YYY` or `R_Array_xX_YYY`
- `R_NETX_DIV_SIP` -> `R_Network_Dividers_xX_SIP`
- `R_NETX` -> `R_Network_Star_xX_SIP`
- move all resistor networks (maybe except the simple volatge divider) into a separate library `resistor_networks.lib`, so they don't clutter `device.lib` (same as already done for `switches.lib` and `mechanical_devices.lib` Then we only have the basic devices in `devices.lib` and the derived ... more specialized stuff in separate libs.

Comments: 
1. This will make the topology more clear from the names. 
2. I would really like to keep the `_SIP`, `_DIP` and even add it to all networks. The reason is IMHO simple: the pin-numbering is what makes the devices different. This is especially obvious for the parallel resistors (SIP: 1-2, 3-4, 5-6, ... DIP: 1-N, 2-(N-1), 3-(N-2), ...). And as far as I've seen , most commercially aravailable R arrays are either in a package with one line of input-pins (i.e. single in-line package = SIP) or with two parallel lines (Dual in-line Package =DIP, if we also understand SO/SOIC and other SMD packages with two parallel columns of pins, not just THT). THT and SMD-variants of DIPs have usually the same pinout. Taking the SOlution of KLC 4.4 would leave us with the problem of which package is the "original" package and which one is the modification ... and how to describe a complete reordering except with saying: OK, this is for devices in a DIP-package, not a SIP ... (or vice-versa)
3. Having all the R_Networks in a separate lib would also make it easier to add more variants (there are variants where the common-pin of the R_NETX is in the center, there are R-2R and ladder-networks ... that are all not yet covered by the proposed symbols ...

What do you think? ... I'm also open to other proposals for a naming scheme ...

Best,
JAN

jkriege2:
OK, so I fixed the diodes to be IEC-conformant:

add_missing

I reworked some of the other diode symbols also (e.g. ZENER-diode, `D_Capacitance`) to match IEC

OK for everybody? Please have a look and double-check!

... then only the problem of R-Network-naming remains?

Best,
JAN

KarlZeilhofer:
Great diodes! Thanks for your work!

Schlumpf:
If we use IEC as standard (and I support that :+1: ), we should change the design of the switches, too. Here is a doc describing the differences between IEC and other (NEMA ICS/ANSI/IEEE): http://www.moeller.net/binary/schabu/SB2011-9_online_DE.pdf (german)

jkriege2:
Mmmhhh ... I would like to move forward towards merging ... what about doing the conversion to IEC in a second PR?

Best,
JAN

jkriege2:
@Schlumpf @SchrodingersGat @pointhi @diggit :
I just looked a bit into that switchy-thing ... For the normal switches and push buttons, I don't care which version to use, but expecially for the 1x12, 2x6, ... rotary switches the IEC symbol would look something like this:
multiswitch
I think in that case the circles really improves readability of the schematic ... but then that's just my opinion ... What do you think?

JAN

diggit:
Switches are ok.
I am not fan of new diodes design. In 90% cases, I saw hollow triangle and I am more familiar with them. Even old design was ok, what was wrong on old design?

maximeborges:
For me the old one are really more explicit 
image
But I think that using the IEC by default is the best thing to do for KiCad. Maybe still having these alternative symbols could be useful though.

jkriege2:
@diggit : I changed them to make the diodes IEC-conformant (which requires a non-filled triangle) ... but please: I want your opinions on this ... as I'm doing electronics as a hobby, I'm not bound to any standard and usually choose the symbol that pleases my eye more (see switches!)

Best,
JAN

maximeborges:
IEC for standardizing the libraries, _ALT for more explicit symbols like big switches and speaker (IMO).

CarlPoirier:
@jkriege2 I'm fine with your proposal for the resistor networks except one thing and it is the _DIP and _SIP suffix. Again, this makes no sense when one uses a BGA package.

jkriege2:
@CarlPoirier 
Hhmmm ... looking around I didn't find any BGA resistor arrays that could be used together with one of the proposed symbols. Usually they have 3 or 4 rows of balls (see e.g. https://www.koaglobal.com/en/product/br.aspx) ... so in the end we would need an additional symbol for resistor networks anyways!

Do you have any proposal for another designator, except `_DIP` and `_SIP`? Maybe `_2PinRows` instead of `_DIP` (but which is more or less the same ;-)

I'm open for suggestions ...

Best,
JAN

PS: What do you think on the swicthes and the diodes topic?

Schlumpf:
For the diodes/switches topic: In my opinion, @maximeborges's idea could be a good way...

jkriege2:
at all:
To make the remaining topics a bit more visible, here's a list:

## Switches
  * [ ] switch to IEC-conformant symbols before merging
  * [ ] switch to IEC-conformant symbols after merging in a new PR
  * [ ] add IEC-conformant symbols as alternative before merging
  * [ ] add IEC-conformant symbols as alternatives after merging in a new PR
  * [ ] don't add IEC-conformant symbols for now (can be done later, if someone wants it)
    - @jkriege2

## Diodes
  * [ ] keep IEC-conformant symbols (non-filled triangles)
    - @jkriege2
  * [ ] switch back to non-IEC-conformant symbols (filled triangles)
    - @jkriege2
  * [ ] add non-IEC-conformant symbols (filled triangles) as `_ALT` alternatives


## Resistor Networks:
Proposals for naming scheme:
* [ ] keep `_DIP`, `_SIP`:
    - @jkriege2
* [ ] change `_DIP`, `_SIP` (please give proposed new appendix!):
    - 

Best,
JAN

PS: Please add questions if I missed some!

CarlPoirier:
@jkriege2 I suggest making the _DIP ones the default and calling the _SIP ones _pinswap1 or any other suffix.

jkriege2:
@CarlPoirier : The problem is _" _pinswap1 or any other suffix"_ ... i.e. WHICH suffix.

Maybe anothzer proposal:
We use R_PACK_n_ for the DIP-style packages and R_PACK_n__SIP for the SIP-style (i.e. old R_NET_n_PAR_SIP ... in the end they are only available in SIP with this pinout (I think)). The drawback is you don't get all if you search for R_NET

Suggestions?

JAN

CarlPoirier:
@jkriege2 This is a fair proposition. So: "R_PackX" like now and and "R_PackX_SIP".

jkriege2:
Hi!

OK, resistors are renamed now:
* R_NetworkXX for standard networks ("star topology")
* R_Network_divider_xXX for volateg divider style networks
* R_PackXX for parallel resistors in a DIP-style package
* R_PackYY_SIP for parallel resistors in a SIP-style package

So I think mainly the question of the diodes remains ... What do you think on that? SHould we have both or only one alternative?

Best,
JAN

CarlPoirier:
Awesome. For diodes, I'd go for non-IEC conformant as _ALT. Everyone will be happy, then.

jkriege2:
@CarlPoirier OK, I added now `_ALT`-sybols for the basic diodes (D, LED, D_Photo, D_Shottky, D_Zener ...) ... not yet for the more complex devices (Bridges, TRIAC, Thyristor, ...) as that would blow up the library. Do you think we should double them too?

cu
JAN

add_missing

CarlPoirier:
I think we can merge like this for now!

Anyone else have comments before?

Schlumpf:
I agree with @CarlPoirier. Other possible improvments can be discussed in own PRs.

jkriege2:
go ahead :-))) :+1:

CarlPoirier:
Here we go!

Congrats Jan for the hard work and to all the reviewers!

jkriege2:
Thanks also from me to everyone who contributed 👍 

JAN

PR #744, opened 2016-10-11
Added FAN7842, high and low side mosfet driver

crackerizer:


diggit:
Please read [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention)

Then use checklib.py from [kicad-library-utils](https://github.com/KiCad/kicad-library-utils) to check/fix some problems for you.

crackerizer:
Hi diggit, I have fix the sch part with the util.

jkriege2:
Hi!

why do you delete so many components (see DIFF)?

Best,
JAN

PR #745, opened 2016-10-11
Add AS5045B

El3ctr0nic:
Add AS5045B to sensors.lib
If my style (key words etc) will be ok i change rest of magnetic
encoders description.


jkriege2:
Hi!

looks good to me and checks out with KLCs. So two things remaining:
1. Could you pull the master into your brach or rebase it, so the conflicts are fixed? (Only then I can merge)
2. What about making the "magnetic" of the sensor explicit by altering the symbol slightly like this:
as5045b

Best,
JAN

PS: Go ahead with any other changes ... best in a second PullRequest, then we can already merge this one!

El3ctr0nic:
I try fix conflicts again pls. check if I made it. I'm still week in Git. 
Regards

El3ctr0nic:
I made it :) Wow :)
Pls check. If all OK i redraw rest of symbols

jkriege2:
Hi!

please look at your DCM-file ... seems like you committed in some remenants from DIFF/PATCH?

cu
JAN

PR #746, opened 2016-10-12
Added L5973D step down regulator

feichtinger:
The L5973D is a 2.5A step down switching regulator in a SOIC-8 package with exposed pad.

bildschirmfoto vom 2016-10-12 22-44-00


jkriege2:
Hi!

thanks for the contribution!

some extra issues:
```
D:\KiCAD\kicad-library-review\library>..\..\kicad-library-utils\schlib\checklib.py -vv --enable-extra regul.lib -p L5973D
checking component: L5973D
    pin VREF (6): O (OUTPUT), expected: W (POWER INPUT)
    field: reference, posx 0, posy -100, recommended posx -100, posy 525
    field: reference, justification C, recommended R
    field: name, posx 0, posy 100, recommended posx -100, posy 450
    field: name, justification C, recommended R
    field: footprint, posx 0, posy 0, recommended posx 150, posy -450
    field: footprint, justification C, recommended L
```

- footprint filter should be `SOIC*1EP*3.9x4.9mm*Pitch1.27mm*`
- set footprint in component properties to `Housings_SOIC:SOIC-8-1EP_3.9x4.9mm_Pitch1.27mm` (there is only this one package)
- PAD seems to be gonnected to GND (the layout on p. 10 of the datasheet suggests that) ... so maybe rename the pin to PAD_GND and make it a POWER_INPUT?

cu
JAN

feichtinger:
I found out that there is also a dc-dc.lib (but it was not included in KiCad) so I moved the part there.

> pin VREF (6): O (OUTPUT), expected: W (POWER INPUT)

VREF is a 3.3V reference voltage OUTPUT. See page 6 of the datasheet.

>    field: reference, posx 0, posy -100, recommended posx -100, posy 525
>    field: reference, justification C, recommended R
>    field: name, posx 0, posy 100, recommended posx -100, posy 450
>    field: name, justification C, recommended R
>    field: footprint, posx 0, posy 0, recommended posx 150, posy -450
>    field: footprint, justification C, recommended L

OK, I repositioned the fields. But does it really look better this way? By the way, the y coordinate has the wrong sign in the library editor.

> PAD seems to be connected to GND (the layout on p. 10 of the datasheet suggests that) ... so maybe rename the pin to PAD_GND and make it a POWER_INPUT?

Hmm, probably the exposed pad has no electrical connection to the chip at all. The datasheet does not specify that. I think the recommendation in the datasheet to connect it to the groundplane is due to practical reasons. (Usually you have a ground plane around a switching regulator.) Anyway, I renamed it to PAD_GND. But I don't think that the pad is suitable as a POWER_INPUT so I better set it to UNSPECIFIED.

jkriege2:
Great work! I'll merge it in this version :+1: 

Thanks,
JAN

PR #748, opened 2016-10-12
Texas Instruments devices for USB hub

razykov:
This pull request adding few devices compatible with **TUSB2036**:

- TPS2044
- TPS2054
- SN75240/SN65240
- SN65220

It need for hub the Universal Serial Bus construction.


jkriege2:
Hi!

Thanks for the contribution! Could you do some minor changes to your components:
1. The outline of the IC-box should be 0.01" thick (see KiCAD library Conventions at https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention ... ther are also scripts to check the KLCs https://github.com/KiCad/kicad-library-utils ... enable the --enable-extra option to get additional hints, the basic KLCs check out for your components).
2. The TPS2044 and TPS2045 seem to have two independent components inside ... maybe have them as two units in the component, so they can be placed independently??? (just a suggestion from me)
3. Another suggestion: I personally like it, when the symbols somehow represent their function ... may that be an option for the TPS-components? Examples:
exfunction
4. put the REFDES- and NAME-Labels above the components (is now usually done that way, see example image above)
5. remove `SN65240` from keywords of SN75240
6. the supprossors are basically ESD protection diodes, so maybe they fit better into the library ESD_Protection.lib ... And here especially it would be great to have the different independent components split and the symbols showing their function, example from ESD_Protection.lib:
esd1 and from device.lib: 
shotky

Would be great, if you could do those changes!

Best,
JAN

razykov:
`5. remove SN65240 from keywords of SN75240`

Why do it? SN65240 and SN75240 have equivalent functionality.

jkriege2:
@razykov: Yes, but why have it? If I'm looking for the one ... why find both?
JAN

razykov:
OK. I will remove it.

razykov:
Can you check my work again? 
Thanks for the help!

jkriege2:
Looks great so far ... some minor things remaining though: 
- keywords should be separated by whitespaces, not by commas (there are commas in all your keywords)
- exact component names are TPS2044D and TPS2054D (note the D ... could you change that still?
- as the TPS2044D and TPS2054D are both ONLY available in SOIC package ... could you set that explicitly as footprint under the properties? Correct package should be `Housings_SOIC:SOIC-16_3.9x9.9mm_Pitch1.27mm`
- SN-devices: footprint filters should be `DIP*` and `TSSOP*`
- SN75240: did you mix up the GND-pins? Shouldn't GND=1,3 belong to C=2,4 and GND=5,7 to C=6,8? Would make more sense to me ... Could you please double-check?

Also I'm getting this output from our check-scripts:
- all components check OK, except:
```
checking component: SN65220
    Pin 1 is missing
    Pin 3 is missing
```
Maybe you could (optionally) add the two missing pins 1 and 3 marked as NC and invisible.

Best and thanks for all the work,
JAN

diggit:
@razykov 
>> 5. remove SN65240 from keywords of SN75240

> Why do it? SN65240 and SN75240 have equivalent functionality (see page 6 par. 10.2).

We have aliases for this purpose. Same symbol, separate documentation.

diggit:
Do not omit pins, symbol must have same pin count as their package. If pin is NC, set is as NC and invisible.

diggit:
> SN-devices: footprint filters should be DIP* and TSSOP*

You should match package dimensions and pin pitch too(if applicable).

razykov:
@diggit 

> You should match package dimensions and pin pitch too(if applicable).

Did you mean this: `DIP*7.62mm*`, `TSSOP*4.4x3mm*0.65mm`, etc... ?

diggit:
Yes, with trailing asterisk.

razykov:
@diggit OK. Thank you for the comment!

razykov:
It seems that everything is OK.

jkriege2:
Hi mostly fine (IMHO) ... I spotted one small error still: In the TPS-devices, the outputs are inverted, but the OC-pins are not ... that should be swapped, as shown here:
tps

Does anybody see anything else? If not and that is fixed I will merge later (or this evening ;,-)
JAN

razykov:
@jkriege2  It was a very serious bug. Thanks for the attentiveness!

jkriege2:
Hmmm ... some last things I just spotted:
- GND and IN are Power Inputs, but OUT ar only outputs, not POWER OUTPUTS -> could you change that?
- one cold think about having the OC pins as Open-Collector (although there's technically a MOSFET inside) ...

cu
JAN

jkriege2:
Thanks for all the work and the contribution ... hope to see more ;-)

Best,
JAN

diggit:
I just noticed TPS2044D and TPS2054D has not FP filters set.
These filters allows you to find all compatible FPs like SOIC-16_3.9x9.9mm_Pitch1.27mm**_Handsoldering** (just example) even when you assign default one. They are not meant to be replacement of FP field for multi FP parts.

razykov:
@diggit  #755

Issue #750, opened 2016-10-13
Mistyping in pin name of microchip/pic16f145X ICs


fredbar:
Small thing. The TX pin on RC4 is written as TK. 
Corrected on all members of the family and checked also all libs for same mistake. 
Pull request on the way.

fredbar:
pull request done.
https://github.com/KiCad/kicad-library/pull/751

PR #751, opened 2016-10-13
fix for #750

fredbar:
TX on RC4 is spelled TK.


SchrodingersGat:
Hey @fredbar please try to refrain from addressing multiple issues in a single PR - it can become really hard to track and merge multiple fixes this way. 

Each unique change should be in its own branch :)

fredbar:
Yes, sorry i had a brainf*ck moment and messed up things.

It's also that i need to work with both those components and doing so in separate branches is problematic.
I need to find a workflow that:
- let me work with the all the components i create while they are not merged
- not have to mess with multiple merges on my own repo
- not force me to have a working directory and a patch submission repo from whom i would extract data.

If you have any proposal, i'm in.
Thanks.

fredbar:
Closed as i need time and method to clean all the mess i am accumulating. BRB.

PR #753, opened 2016-10-15
added Microchip Ethernet PHY KSZ9031RNX

ashtonchase:
**KSZ9031RNX**

Device Datasheet:

http://ww1.microchip.com/downloads/en/DeviceDoc/00002117C.pdf

Product Page:
http://www.microchip.com/wwwproducts/en/KSZ8999


jkriege2:
Hi!

first of all: Thanks for the contribution!

second: We have the KiCAD library Conventions at https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention that explain how symbols should be made to fit into the lib... there are also scripts to check the KLCs https://github.com/KiCad/kicad-library-utils ... enable the --enable-extra option to get additional hints

now, a more detailed review:
- the border of your IC should be 0.01" thick
- remove the `DOCUMENTATION` from the Datasheet field in the component properties
- maybe rename `P_GND` to `PAD/GND` to make the bottom-PAD more explicit
- finally the footprint filter should be (IMHO): `QFN*1EP*7x7mm*Pitch0.5mm*`
- also I would suggest to move the VALUE-field (IC name) to the center, as shown below:
ksz


otherwise the comp checks out with our KLC.

Best,
JAN

jkriege2:
... and P_GND should be a POWERINPUT, as VSS is!
cu
JAN

ashtonchase:
Thanks @jkriege2. 

I've incorporated all recommended changes.

jkriege2:
Hi!

I think the datasheet FIELD is unnecessary ... The datasheet was already mentioned in the component props.

Also our check-script (see my first post) changes the component (using option --fix) to this alignment of the REFDES/Value/Footprint fields ... If you like that better, maybe do it like that?
ksz


cu
JAN

Otherwise: looks good to me

jkriege2:
Please fix conflicts (and implement final remarks)! ... Then I can merge!

JAN

ashtonchase:
@jkriege2.

Ready to review again.

jkriege2:
Hi!

OK; I merged this! Thanks again :-)

JAN

PR #762, opened 2016-10-21
Add AS5045B

El3ctr0nic:
After mess i made i try add it second time


jkriege2:
Great!

thanks for the contribution.

JAN

PR #763, opened 2016-10-21
Add ACPL-C79x optical isolation amplifier and ACPL-C87x optical isolation voltage sensors

El3ctr0nic:
I dont have good idea to draw nice symbol ;/ Pls help :)
Here are my propositions


El3ctr0nic:
vote

jkriege2:
I like variant C best ... although power-supply and output pins should not meet on the output side!

Best, JAN

SchrodingersGat:
Another vote for version C. And @jkriege2  is correct, please separate the supply and output pins from (visually) connecting.

El3ctr0nic:
vote2

Fixed C and add D version :) 

On schematic version C is a little bit too small. It is indvidual feeling I think,

No changes anymore :) 
Pls give feedback and i prepare finall version of component

jkriege2:
:+1: OK for me ...
JAN

El3ctr0nic:
vote3
I think that is something what i looking for :)

jkriege2:
Hi!

not bad, but I think what I like more in former option D is that the whole thing is a complete amplifier symbol, just split by a gap. the symbol above looks like two ...

cu
JAN

jkriege2:
Nice work!

JAN

PR #699, opened 2016-09-13
Lm285z adj

piotr-wiszowaty:

PR #698, opened 2016-09-13
Lm285z

piotr-wiszowaty:

PR #764, opened 2016-10-22
Add missing references

jkriege2:
added these reference families in different packages:
- [ ] LM285
- [ ] LM385
- [ ] LM4040
- [ ] LM4030
- [ ] REF30_xx_

Note: All devices violate KLC rule 3.1 ... but then all diodes in devicxe.lib do it too ... so should I make the 3.1 compliant, or leave it like that? I like the smaller symbols better.

Note: Also these devices match the IEC-diode style (non-filled triangle) like also #742 ! 

Note: I fixed several other issues in this lib (footprints/fpfilters, diverse KLC-things, ...)

Best,
JAN

PS: This should also make some more PRs closeable, as the devices are included here (with more variants!):
- https://github.com/KiCad/kicad-library/pull/699
- https://github.com/KiCad/kicad-library/pull/698

ref


SchrodingersGat:
Jan the multiple invisible pins look quite cluttered:

clutter

I have been considering how we approach this issue, as ideally, we should *not* include NC pins in the symbols! However we of course need them for doing match-by-pin-count functionality.

Maybe the best idea is to wait for the new symbol format which allows multiple pin numbers to be mapped to a single pin?

In any case, maybe we can place all the invisible pins in the same spot? It looks ugly in the symbol editor but I think it is *less* ugly than the current solution.

capture

jkriege2:
Wouldn't the latter option connect all the NC pins?

SchrodingersGat:
I just tested this and it would appear that you are correct - pins that are on top of each other are implicitly connected even if there are no wires going to them.

In that case, the way that you have drawn them current is going to be the best (until the new and improved file format comes along).

SchrodingersGat:

PR #767, opened 2016-10-25
Added CP2104 USB to UART bridge

SchrodingersGat:
https://www.silabs.com/Support%20Documents/TechnicalDocs/cp2104.pdf


jkriege2:
Hi!

```
checking component: CP2104
    pin NC (10): P (PASSIVE), expected: N (NOT CONNECTED)
    field: reference, posx 300, posy 1050, recommended posx -300, posy 925
    field: reference, justification L, recommended R
    field: name, posx 300, posy 950, recommended posx -300, posy 850
    field: name, justification L, recommended R
    field: footprint, posx 300, posy 850, recommended posx 150, posy -950
```
... maybe just the EC01 is really valid?

Best,
JAN

SchrodingersGat:
In the [datasheet](https://www.silabs.com/Support%20Documents/TechnicalDocs/cp2104.pdf) it says that the NC can be left unconnected **or connected to Vio**.

For this reason I have left it visible, as it may be connected to the other pin.

jkriege2:
Then maybe mark it as NC and call it NC_OR_VIO ??? Currently its PASSIVE ... But my proposal is also not the best solution ;-) What do you think?

JAN

SchrodingersGat:
Personally I'd leave it as is - it's up to the user to read the datasheet and decide what to do with the pin ;)

jkriege2:
OK, but is the pin PASSIVE or NC? Or even POWER_INPUT, if it can be optionally tied to Vio?

SchrodingersGat:
I will change it to power_input then :)

jkriege2:
Cool with me ... I'll merge when done ;-)

SchrodingersGat:
fixed :)

PR #769, opened 2016-10-25
Added more SP050x chips

SchrodingersGat:


SchrodingersGat:
@jkriege2  added some more of the SP050x series chips here

jkriege2:
I was thinking a bit about these devices ... would it maybe make sense to have them not separated, but as a single symbol ... like a resistor network or the x2-diodes in device.lib ... otherwise there is the dange that the user connects the common pin once to a signal, say Vcc and once to another signal, say GND ... the symbol does not prevent that or make that obvious ... but this will create a shortcut ...!

Best,
JAN

SchrodingersGat:
Yeah good idea. I have made this change as suggested.

jkriege2:
Great!

JAN

PR #770, opened 2016-10-26
Added various analog comparators

SchrodingersGat:
TS881
ADCMP35x


jkriege2:
Hi!

- ADCMP...: I think you swapped pins 4 and 1 ... accoridng to datasheet 1 is Vin and 4 is Vcc
- rest is OK ...

cu
JAN

SchrodingersGat:
Ah, great catch, thanks Jan. Now fixed.

jkriege2:
One more proposal for the ADCMP comparators:
What about showing the reference source in the symbol? That would make (IMHO) the function of the device clearer, when reading a schematic, e.g.:
comp

JAN

diggit:
Why does output pin have inverted graphic style? It does not match datasheet.

jkriege2:
Because the 350 is as I proposed above, but for the other two devices the REF is connected to - and the input is + ... so 350 has technically the inverted function as the others ... I think my proposed symbol would solve that problem ;-)

JAN

diggit:
I think this is a bit confusing. I'd stay consistent with datasheet where I read about which version have which input used for ref and input.

jkriege2:
@SchrodingersGat : How about this one? Could you fix conflicts and see whether we get it done?

best,
JAN

SchrodingersGat:
Conflicts have been fixed

jkriege2:
OK! nice work!

JAN

PR #776, opened 2016-10-28
Added LPC822M101JDH20 to nxp_armmcu.

michalkona:


diggit:
Please fix these:
- set VREF pins as inputs, there is no power delivery.
- move GPIO names (PIO...) to closest place to pins/symbol edges
- assign default FP: `Housings_SSOP:TSSOP-20_4.4x6.5mm_Pitch0.65mm` and set FP field invisible (deselect _show_)
- add FP filter `TSSOP*4.4x6.5mm*Pitch0.65mm*`

keruseykaryu:
> set VREF pins as inputs, there is no power delivery
Why inputs? Passive.

diggit:
Thanks, I wan't sure here. @michalkona please change it to passive as corrected.

michalkona:
What do you mean by moving GPIO names (PIO...) to closest place to pins/symbol edges? I am limited by the rules stated in the Kicad Library Convention (rule 3.1).

jkriege2:
Hi!

I think @diggit means that you should reduce the pin name offset:
2016-11-26 09_44_41-part library editor d__kicad_kicad-library-review_library_nxp_armmcu lib

If that's fixed, I'll merge this PR.

Best,
JAN

jkriege2:
nice work!
Thanks,
JAN

diggit:
I meant when you have pin name eg. **PA4**/UART1_RX/SDA/INT1 on **right** side of symbol, change it to UART1_RX/SDA/INT1/**PA4**. (just example) It really improves readability. Leave offset at defaults.

PR #777, opened 2016-10-29
Added LFxxABDT-TR to the regul library.

michalkona:


diggit:
Please add FP filter: `TO-252*`
Why did you skip 6V and 8.5V versions?

michalkona:
I used table on this web site http://www.st.com/content/st_com/en/products/power-management/linear-voltage-regulators/low-dropout-ldo-linear-regulators/lfxx.html and I added to library only the LFxxABDT-TR with explicitly assigned output voltage. if it is necessary I can add all LFxx voltage regulators in DPAK package.

michalkona:
I think all is done.

michalkona:
What is the reason why the "need to improve" label has not been erased yet?

jkriege2:
Nice work! Thanks,
JAN

PR #784, opened 2016-11-06
Wiznet W5100 and Microchip LAN8720

cafee:
Hi i have drawn two schematic symbols:

Wiznet W5100 in new wiznet.lib
Microchip LAN8720A in microchip.lib


jkriege2:
Hi!

first of all: Thanks for the contrib. 
Could you pleasre have a look at your PR and merge in the current repository version correctly? Your PR deletes several symbols from microchip.lib (see: https://github.com/KiCad/kicad-library/pull/784/files)!

Otherwise some comments on the symbols (output from our check-scripts and explanation below):
````checking component: W5100
    Component line thickness 0, recommended 10
    Pin 3 is missing
    Pin 60 is missing
    Pin 61 is missing
    Pin 62 is missing
    Pin 66 is missing
    Pin 67 is missing
    Pin 70 is missing
    Pin 71 is missing
    Pin 72 is missing
    Pin 73 is missing
    Pin 75 is missing
    Pin 76 is missing
```
So:
1. set the outline width of the surrounding box to 0.01 and its fill color to background
2. Since there is only one package on this device, please set the footprint-field correctly in the form `library:footprint` (i.e. `Housings_QFP:LQFP-80_12x12mm_Pitch0.5mm` for your device) and the the footprintfilter to `LQFP*12x12mm*Pitch0.5mm*`
3. please add all the missing pins. Even if they are NC. In the latter case name them NC, select NC mode and set the invisible



```
checking component: LAN8720A
    pin VSS (25): P (PASSIVE), expected: W (POWER INPUT)
    Component line thickness 0, recommended 10
```
```
So:
1. set the outline width of the surrounding box to 0.01 and its fill color to background
2. correct Pin 25
3. set footprint filters to `QFN*1EP*4x4mm*Pitch0.5mm*`

Thanks,
JAN

cafee:
Hi,
 sorry for problem with merge, i didn't noticed it before creating pull request.

 For WizNet W5100 i corrected 1 and 3.
    Footprint for W5100 is LQFP-80 Pitch 0.4mm (don't ask me how i found that :) ). This footprint is not currently available in QFP footprints. (I don't have correctly drawn footprint)

For LAN8720A corrected 1.2.3.

I there be anything more wrong please let me know.

jkriege2:
Hi!

Then codify the footprint as it should be ... and if you have time, maybe you can even create it for the lib?

Everything else looks fine: Thanks!

JAN

cafee:
Hi

Footprint association is with name as it should be.

Lukas

PR #785, opened 2016-11-06
motorola: add MC68000FN (PLCC)

piotr-wiszowaty:


jkriege2:
Hi!

thanks for the contrib ... and I checked it quickly:
```checking component: MC68000FN
    pin NC (18): B (BIDI), expected: N (NOT CONNECTED)
    Component background filled with TRANSPARENT color, recommended is BACKGROUND color
    Pin 31 is missing
```

SO I think the main things to change are:
1. NC-pin (18,31) should be NC, not BIDI and invisible
2. fill outline with background
3. Pin31(NC) is missing
4. Set footprint to `Sockets:PLCC68`
5. set Footprint-filter to `PLCC*`

Best,
JAN

piotr-wiszowaty:
Done. Questions:
1. I've searched for PLCC68 footprint (not a socket) but couldn't find 
any. Is there one?
2. Why broad PLCC* filter? This symbol is 68 pin only.
3. Are you using the newest checklib.py? I've just pulled it and it 
crashes on motorola.lib (on character ²).
Error message:
Traceback (most recent call last):
   File "/home/p/bin/checklib.py", line 82, in <module>
     lib = SchLib(libfile)
   File "/home/p/soft/kicad/kicad-library-utils/schlib/schlib.py", line 
296, in __init__
     self.__parse()
   File "/home/p/soft/kicad/kicad-library-utils/schlib/schlib.py", line 
328, in __parse
     self.components.append(Component(component_data, comments, 
self.documentation))
   File "/home/p/soft/kicad/kicad-library-utils/schlib/schlib.py", line 
220, in __init__
     x = checksum_data.encode('utf-8')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 
28022: ordinal not in range(128)

jkriege2:
Hi!

ad 1) ... I also dodn't find one ... if you want: make one and add another PR... otherwise: I think for now the socket is OK and if you want to use another footprint, teh filter(s) should find it, if it gets available.
ad 2)So you find any viable alternative (e.g. a socket, directly soldered ... or whatever, as long as it is PLCC and has the correct number of pins (that's why we add the NC-pins so the pin-number can  be matched against the footprints library)
ad 3): yes I used the up-to-date tools ... with the -p option ...???

Best,
JAN

piotr-wiszowaty:
Symbol selection with either -p or -c does not help. Maybe the operating 
system has something
to do with it (mine is GNU/Linux)...

jkriege2:
Am 08.11.2016 um 20:27 schrieb Piotr Wiszowaty:
> Symbol selection with either -p or -c does not help. Maybe the operating
> system has something
> to do with it (mine is GNU/Linux)...
>

Strange ... then let me know if I should run it again for you ...

JAN

piotr-wiszowaty:
The patch I attached resolves the problem. I wonder - did it pass 
through the github's mail system?

jkriege2:
^Hi!

no ... no patch ... what about pushing it to the utils repo?

best,
JAN

piotr-wiszowaty:
Hmm... Fork+PR you mean?

jkriege2:
yep

PR #788, opened 2016-11-08
Add ADuM 1410/1411/1412 to analog_devices. lib

El3ctr0nic:
Initial symbol look voting, propositions


El3ctr0nic:
adum_vote

Pls vote, maybe some other ideas? Regards.

jkriege2:
Hi!

I would stick to the left variant, as it's closer to the datasheet ... 

Note: You're still missing the datahseet ... and for some reasons this PR shows the `ADE7758` #787 as addition ??? Could you maybe pull the master again into your branch?

Best,
JAN

El3ctr0nic:
With lib. I should use?
analog_devices or interfaces?
First symbol is closer to datasheet but is bigger and not good "readable" on schematic zoom out.
And next issue : symbol is outstanding from current digital isolator symbols.
In analgo_devices.lib is another symbol that need to be redraw/ moved to interfaces.lib
Should I care of it on this stage?

jkriege2:
Hi!

I would go for analog_devices.lib ...
About the other symbol: Maybe we finish this one first and then modify the other one in a separate PR?

JAN

El3ctr0nic:
adum

jkriege2:
Hi!

please remove the pin count form eth FPFilter . The pin count filter matches these ... otherwise it looks good so far.

cu
JAN

jkriege2:
Nice work!
Thanks,
JAN

PR #789, opened 2016-11-09
Added BD1020HVF Analog Temperature Sensor

crackerizer:


jkriege2:
Hi!

thanks for the contrib! Could you please change some minor things?
1. please hide the NC-pin
2. please add a footprint and footprint filter
3. A suggestion: What about adding something to the symbol that indicates this is a temperature sensor? e.g. like the thermometer in the  MCP9700 symbols ...

est,
JAN

crackerizer:
1,3 are done. I have added footprint filter to the symbol but there is still no footprint (HVSOF5) in the module library.

jkriege2:
Do the two addition PRs to TO_SOT-lib contain the footprints? If so, I'll review them this evening ... then we can finish this PR with a complete symbol ;-)

JAN

jkriege2:
Please have a look at my comment there: https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/23 ... Otherwise this symbol looks good to me. So if you would already set the footprint field then I can merge and the footprint is available, as soon, as the issues there are resolved.

Best,
JAN

crackerizer:
I have set the footprint to TO_SOT_Packages_SMD:HVSOF5.

PR #792, opened 2016-11-11
xilinx: add XC95144XL-TQ100

piotr-wiszowaty:


jkriege2:
Hi!

Only one slight typo: 
There is an extra h in the web-adress `hhttp` .

Best,
JAN

piotr-wiszowaty:
Fixed.

jkriege2:
Thanks!

JAN

PR #793, opened 2016-11-11
added Intel 8255A Programmable Peripheral Interface to Intel library

nickysn:


nickysn:
Argh, just saw there's a another pull request for adding an 8255, contributed by someone else. :( I didn't know about it, when I made and submitted this. Feel free to add whichever you think is better.

jkriege2:
Hmmm ... I think I'll merge your symbol, as it is much cleaner than the other PR ... But could you please still change the following:
1. Maybe use this ds link (it's direct, not with an HTMl-wrapper): `http://aturing.umcs.maine.edu/~meadow/courses/cos335/Intel8255A.pdf`
2. Would you be willing to also add the 82C55 variants? http://jap.hu/electronic/8255.pdf ... maybe even in the PLCC-package? 

Thanks already,
JAN

nickysn:
Done. Although, I don't know what value should I use for the "Footprint" field for PLCC-44 and for the footprint filters, since I've never done or used one of these before.

jkriege2:
Hi!

thanks already:
- I think for the package you should use `Sockets:PLCC44` and then as FPFilter: `*PLCC?44*` ... that's the only PLC44 that I found (for now ;-)
- Could you please also add the NC pins 1,12,23,34 (marked as NC and invisible) of the PLCC-44 package? We try to make the pins in the symbol match the pins in the package, so we can use the pin-number filter when assigning a matching package together with the FPFilters.

Otherwise this looks great!

JAN

nickysn:
ok, done :)

jkriege2:
Thanks. Please also move the NC pins onto the 100mil grid (as the other pins). Also: Could you remove te device name (8255, 82C55, ...) from the description field?

Then I can merge. 

JAN

nickysn:
done

jkriege2:
:+1: Thanks!

Nice work,

JAN

PR #796, opened 2016-11-20
Added comparator and opamp

kajusK:
Added MCP6562 comparator and TSV912 opamp


jkriege2:
Hi!

nice work ... except some minor things:
1. Your datasheet links are missing the http:// prefix. Without that KiCAD cannot open the DS (at least on Windows)
2. Since the devices have two/three package options, you should not set the footprint field, or make device variants, one for each package
3. The package are all with extra exposed pads (1EP) at the bottom ... but according to the DS that are the wrong packages (except for the TSV912 in DFN package) ... please correct
4. For the TSV912 in DFN-package, you should add a second symbol which has a pin 9 for the EP
5. TSV912/MCP6562: Pins 1 and 7 should have a name assigned, or use `~` for NO_NAME


Best,
JAN

kajusK:
Hi Jan,
thanks for pointing out the issues:
1) didn't know that, fixed
2) I thought there should be some default value, removed
3) oops, thanks, fixed
4) lets avoid the DFN version in this pull request :-)
5) It they are already named ~

Thanks,
Jakub

jkriege2:
Thanks. Now they look fine ... I opened an issue https://github.com/KiCad/kicad-library-utils/issues/66 for the ~ as pin names, as that was reported by our check scripts and I didn't double-ckeck immediately.

One minor thing: As the symbols now do not cover all possible packages, please name the packages in the description ... 

Best,
JAN

kajusK:
Good point, added.

jkriege2:
please add the package at the end of the description ... that seems common in most symbols in the lib ... also I would use MSOP instead of MiniSOP ... I think that's more common and relates to the exact package name!

JAN

kajusK:
That was something I was not sure about. Fixed

jkriege2:
THanks for the quick responses and changes ;-)
JAN

PR #798, opened 2016-11-20
Added SN74LVC1G80

flaminggoat:
Single Positive-Edge-Triggered D-Type Flip-Flop
image


keruseykaryu:
Have you seen the `74xgxx.lib`?

flaminggoat:
I had not,  I think I must have added that lib to my kicad path after making this one. Not really a fan of the hidden pins though. I'll close this anyhow.

PR #801, opened 2016-11-21
xilinx: add XC95144XL-TQ144

piotr-wiszowaty:


jkriege2:
nice work!
JAN

piotr-wiszowaty:
Thank you.

PR #803, opened 2016-11-23
Added INA138/168 current shunt monitor

SchrodingersGat:
http://www.ti.com/lit/ds/symlink/ina138.pdf


jkriege2:
FPFilter should be SOT?23* with an asterisk? ... Although that may lead to mixup with SOT-232 as that also matches ... Any idea, how we can deal with that?

cu
JAN

jkriege2:
Otherwise it looks good ... and I would merge!

SchrodingersGat:
I have adjusted the FP filtre :)

PR #805, opened 2016-11-23
Add initial symbol library table to template folder.

stambaughw:
This is the initial table which contains all of the symbol libraries
currently in the library folder.


template/sym-lib-table
```diff
@@ -0,0 +1,85 @@
+(sym_lib_table
+  (lib (name 74xgxx)(type Legacy)(uri ${KICAD_SYS_SYMBOL_DIR}/74xgxx.lib)(options "")(descr "Legacy 74xgxx symbol library."))
```
keruseykaryu: Why not keep the current format of KiCad variable name: `KISYSSYM` and `KIPRJSYM` ?

stambaughw:
On 11/23/2016 11:55 AM, Kerusey Karyu wrote:
> *@keruseykaryu* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In template/sym-lib-table
> <https://github.com/KiCad/kicad-library/pull/805#pullrequestreview-9903854>:
> 
>> @@ -0,0 +1,85 @@
> +(sym_lib_table
> +  (lib (name 74xgxx)(type Legacy)(uri ${KICAD_SYS_SYMBOL_DIR}/74xgxx.lib)(options "")(descr "Legacy 74xgxx symbol library."))
> 
> Why not keep the current format of KiCad variable name: |KISYSSYM| and
> |KIPRJSYM| ?

I thought about it but KISYSSYM isn't very readable.  I never really
liked KISYSMOD but that was not my decision.  Honestly, I don't have a
strong preference but I tend to lean towards readability when it comes
to environment variables.

cpavlina:
I don't like KISYSMOD either, but don't you think it's more important to continue using variables of that format so they're consistent and memorable? Fixing something in one little part of kicad but not the rest leads to horribly confusing spaghetti design. If you don't like KISYSMOD or KISYSSYM, maybe we should look at totally renaming all of them in the future - but for now, let's stay consistent so we're not driving everybody bonkers.

PR #806, opened 2016-11-23
Added SW_Push_LCD

openelectron:
New part added.
Several manufacturers make these type of switches. I think they should be in the switches library rather than LCD's but you could argue either way.

http://www.e3-keys.com/files/SA%20Technical%20Datasheet%20v2.0.pdf
http://www.ledswitches.co.uk/wp-content/uploads/2014/06/pdf/Q5-100-14.pdf


jkriege2:
Hi!

thanks for the contribution. I reviewed it and there are some issues:
1. We have a set of conventions for this lib and I would ask you to stick to them: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention There are also scripts to check them: https://github.com/KiCad/kicad-library-utils
2. To conform to KLC: Please make the line width of the outline 0.01" thick and fill it with the background color
3. I would move the SW-pins to the switch symbol and all the inputs to the left (kind of standard in the lib) ... then the whole thing is clearer:
2016-11-23 18_35_55-part library editor d__kicad_kicad-library-review_library_switches lib
4. These devices seem pretty special and the pinout etc. will be heavily manufacturer-dependent. Therefore I would rename the device to the actual name from a manufacturer and add different symbols for different manufacturers/types, e.g. `SW_E3_SA3216` ...
5. Finally: Could you also provide the matching footprints? That would be a great addition.

Best,
JAN

openelectron:
@jkriege2  I have included most of your suggestions.

I cant find the rule about the background colour or line thickness in the KLC, or in the test scripts hence the symbol passed the checks. I do agree it looks better and have changed it. How do we add that rule to the KLC and subsequently the check scripts?

All other suggestions incorporated, I have added alias entries for the other parts that fit here.

I will be creating the footprints as soon as I can.

Thanks.

jkriege2:
Hi!

looks good now ... but what is missing are descriptions, keywords, DS-links for the aliases ... they are not copied automatically, so you'll have to set them manually.

cu
JAN

openelectron:
@jkriege2 
oops, I didn't realise they weren't automatically inherited from the parent.
all done.

jkriege2:
nice work,
Thanks!

PR #809, opened 2016-11-30
added symbols for LED with PAD, several LEDs in series in a single package and several LEDs in series in a single package with pad

jkriege2:
These symbols do in part match the footprints int his PR: https://github.com/KiCad/LEDs.pretty/pull/21

Also added new `leds.lib` for special LEDs, such as the CREE-LEDs that Is tarted with (ffotprints, see PR above).

led1
led2
led3
led4
led5


SchrodingersGat:
Great! Thanks for the contribution.

PR #811, opened 2016-12-01
Align grid with BeagleBone pins

dpifke:
This change sets the grid origin to the first pin on P9.  This allows
us to easily place components with the same pin spacing and alignment.


ashtonchase:
Thanks for your contribution. I'm merging now.

PR #813, opened 2016-12-04
Added ZXGD300XE6 gate drivers

kajusK:
Hi,
I added a mosfet gate drivers, I hope the transistors library was the correct one to use...

driver


kajusK:
Edit: Both sink and source set as output prevents connecting them together (ERC fails), changed to open collector/emitter accordingly.

jkriege2:
Hi!

thanks for the contribution. Could you please change some minor things:
1. The Footprint-filter should be `SOT?23-*` ... the pin number is filtered by the pin-number filter, so it does not have to be in teh FPFilter.
2. lease set the width of the box-outline to 0.01" (see https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention ... there are also scripts to check these rules: https://github.com/KiCad/kicad-library-utils)
3. Please move the part to a (new) library `diodes_incorporated.lib` or `zetex.lib` ... in transistors.lib we have only single transistors and this device seems to be more complex ;-) ... COmparable gate drivers from IR are found in `ir.lib` ... or am I mistaken here?

Otherwise it looks good to me!

Thanks,
JAN

kajusK:
Fixed, hope it's ok now.

jkriege2:
nice work!

Thanks,
JAN

PR #816, opened 2016-12-04
Added generic male and female connectors to conn library. 

Marsfan:
There seemed to be a lack of Generic Male/Female connectors in the conn library. I needed to work on something using Banana jacks, and I could not find a simple schematic for connectors. These fit IEEE/ANSI Standards. 

I just learned how to use kicad last week, Please let me know if anything looks wrong.


jkriege2:
Hi!
...and thanks for the contribution.

First of all. We have some conventions for parts in this libarary (FYI). You can find it here:
Also we have some scripts that check some of these rules:
  https://github.com/KiCad/kicad-library-utils

So now for your symbols. Could you please change the following:
1. We try to make the lib conform mostly to IEC-60617. So the connector symbols for male/female should look more like theis:
male_female
2. Please rename the symbols to CONN_01x01_MALE and CONN_01x01_FEMALE (then they fit the rest of the conn.lib)
3. the REFDES should be P instead of J (see e.g. https://en.wikipedia.org/wiki/Reference_designator)

best,
JAN

Marsfan:
Thanks Jan. I will work on these right now. Where can I download IEC-60617?

jkriege2:
http://pcad-libs.embedders.org/rules/ref_617.pdf
http://www.tinline.ch/tinweb/download/category/31-autocad-ecscad.html?download=103%3Aiec-60617-katalog

Marsfan:
Fixed it up. Let me know if any other tweaking is needed.

Also, where are SPST and SPDT switches? They are in kicad windows download, but I cannot find them in the repo.

jkriege2:
Switches should be in `switches.lib` (see https://github.com/KiCad/kicad-library/blob/master/library/switches.lib)

Marsfan:
Thanks, for some reason that .lib is not loaded for me. Oh well, time for some debugging.

jkriege2:
Hi,
nice work ... what would you think of moving the labels a bit, e.g.:
2016-12-05 22_47_55-part library editor d__kicad_kicad-library-review_library_conn lib
... or consistent with CONN_01x01:
2016-12-05 22_49_04-part library editor d__kicad_kicad-library-review_library_conn lib

jkriege2:
@diggit @pointhi @SchrodingersGat What do you think?

Marsfan:
Ok. And I realized that the pin does not really fit the grid. Makes wiring a pain. I will fix that.

SchrodingersGat:
It appears that the descriptions for all the CONN_xx connectors have been deleted from the .dcm file, you would want to rectify this before this can be merged.

Regarding the need for specific male/female connectors, we already have generic CONN and Terminal_Block symbols:

eg

I would be open to adding a third style if there is a real need for it, but I'm not quite sure what is being asked here. Is it only a single connector symbol that you are asking for? e.g. `Jack_1x01` or something like that? Or an entire series?

SchrodingersGat:
This is the same issue we discussed in https://github.com/KiCad/kicad-library/pull/719

Marsfan:
These are generic jacks, I have not planned a series. I was intending to use them for things like these
banana jack.


As for the strange disappearance of connectors, it seems that for some reason my computer deleted everything in conn.dcm. Add this with the disappearance of some libraries, and I am VERY confused. There is a backup, I will try to restore it. Might take me some time, now that the week has started up again.

Marsfan:
Fixed the accidental deletion of the connector description. There was some sort of git merge error.

jkriege2:
Hi!

two things:
1. now the new sy,bols are missing the description ;-)
2. I would be in favour of having these symbols and I would propose we make a small series from them, i.e. devices 1x01, 1x02, 1x03 (I think e.g. for the 4mm-stuff these varieties are available. I think they can really come in handy for such cases as proposed above + we have a standard-conforming symbol, which other people might want to use for other stuff.

Best,
JAN

Marsfan:
Here are some descriptions and I added up to 01x04. I hope this looks good. Thank you so much for being kind about giving me tips.

jkriege2:
Hi!

please align the pins on a 100mil grid.

Best,
JAN

Marsfan:
Here, this should do it.

Marsfan:
@jkriege2 can you take a look at this now?

jkriege2:
Hi!

to me it looks good up to one small thing:
coul you switch on pin numbers for the symbols? For the x01 that's maybe not necessary, but this way the symbols stay consistent with the CON_XXxYY:

2016-12-19 07_45_55-eeschema kicad-library-review _ d__kicad_kicad-library-review-project

Best,
JAN

PS: Sorry for the long delay in review ... I'm wuite busy at the moment ...

Marsfan:
@jkriege2 it will be ready in a few day.

Marsfan:
@jkriege2 Added it. Sorry for the delay, it totally slipped my mind until now.

jkriege2:
looks good!

nice work, thanks, JAN

PR #823, opened 2016-12-06
Add IRS2092, IRS20957, and IRS21867S

evanshultz:
They all pass KLC without any issues.

Datasheets:
IRS2092: http://www.infineon.com/dgdl/irs2092.pdf?fileId=5546d462533600a401535675f1be2790
IRS20957S: http://www.infineon.com/dgdl/irs20957spbf.pdf?fileId=5546d462533600a401535676143e2799
IRS21867S: http://www.infineon.com/dgdl/irs21867spbf.pdf?fileId=5546d462533600a4015356770e8327eb

Here are symbol images:
image


jkriege2:
Hi!

could you maybe change the following:
1. hide the NC-pins
2. could you reorder the pins, so that inputs are generally on the left, outputs on the right and power-supply at top/bottom?
3. remove commas from keywords

Best,
JAN

evanshultz:
Thanks for the feedback.

1. Done.
2. This is generally already done with the default pin ordering.
3. Done. There are commas in other parts in this lib.

evanshultz:
I fixed IRS20957 since it's only available in SO16.

For IRS2092, should the DIP and SO package be aliases?

jkriege2:
Hi!

I would propose something, which is nearer to the logic symbol (incl. grouping ...), e.g. like this:
2016-12-07 20_01_38-part library editor d__kicad_kicad-library-review_library_ir lib

2016-12-07 20_04_28-part library editor d__kicad_kicad-library-review_library_ir lib
You could even draw some sketch of the interior setup, as e.g.:
exfunction

Also you should change the different pin types:
1. VAA,VCC,VB,VSS,COM,VS should be power/voltage input
2. all other inputs should be simple inputs

Best,
JAN

evanshultz:
Hello again, Jan, and thanks for your continued review.

I looked at some other IC pins in the KiCad library and found a few things:
1. Floating supplies (like VB and VCC and VAA pins here) appear to be Passive. I assume so there's no need for a big PWR_FLAG symbol on those nets as they're often separated from a true power supply by 10ohm or so. Many opamps have the power pins set to Input, though. Knowing that the KLC libraries are often a bit non-conforming, I'm not 100% sure what's best for these pins.
2. True inputs like IN of IRS20957 should be Input type.
3. True outputs, like VREF and HO should be Output type.
4. Other pins should be Passive because they'll have resistors, capacitors, transistors, etc. connected.

I modified these three parts correspondingly, but I'm not 100% sure if this is all correct. You have an excellent point that not all of my pins were of the proper type, but if the above is wrong then I need to make more changes.

Thanks again!

evanshultz:
@jkriege2 How about now? Anything I can do to speed along merging this PR?

jkriege2:
HI!
I'm a bit busy at the moment ... hope I can have a look on the weekend ;-)

Best,
JAN

jkriege2:
Hi!

about the floating power pins I'm also not so sure ... I would make them POWER-pins ... ut then also see this thread: https://github.com/KiCad/kicad-library/pull/850 ... the topics there are somewhat comparable. So I would recommend to have the pins as POWER_INPUT ... i the end it does not harm, or?

A second thing: Please group the pins logically, as proposed above (see KLC 3.4 https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention). At least put the main power supply pins VCC to the top and GND/COM to the bottom. For IRS21867S it's mostly OK (up to VCC/COM), but for the others: please reorder, see e.g. this device from the lib (although the power pins are different than recommended ;-)
2016-12-19 07_54_54-part library editor d__kicad_kicad-library-review_library_ir lib


Best,
JAN

jkriege2:
Here' another examples of how these devices could look:
- http://forum.cxem.net/uploads/monthly_09_2014/post-49172-0-02792300-1410537911.jpg

evanshultz:
Hi Jan, thanks for taking a look!

As long as the giant PWR_FLAG isn't needed on the input power pins I'm fine with everything. They're now changed to be POWER_INPUT type.

With regards to the pin locations, I'll look at it more later. Personally, having text written up is tremendously ugly to me and less readable than horizontal text. Let me see what I can do.

jkriege2:
Hi!

I'm not really sure, what you mean by the PWR_FLAG?

Best,
JAN

jkriege2:
Would this PR help?
https://github.com/KiCad/kicad-library/pull/855

jkriege2:
Are you still working on that?
Best,
JAN

evanshultz:
Fancy you should ask today! I was working on this last night and will send a proposed symbol update to this and the UCC289x parts this evening.

evanshultz:
Here's a shot at the symbols (old ones at top):
image

I left room for the VCC and bootstrap caps on the right side, and the pins are all arranged to make it look nice. On the left side, I grouped the floating input pins at the top section and the COM-referenced at the bottom, again leaving room for a cap on the CSD pin. Since there are 3 different voltage domains here, and the capacitors will route nicely between the pins as shown, I think it makes most sense to avoid top and bottom in favor of this arrangement. Thoughts?

In addition, I updated the FP list on both parts and added a default footprint for IRS20957S since it only comes in one package.

Fails KLC because of VREF direction which is mentioned above and text location but that's not critical

jkriege2:
Symbols look fine, but please use the standard FPFilters, e.g. `SOIC*3.9x9.9mm*Pitch1.27mm*` (i.e. geometric info and pin number removed.

Also IRS21867S seems to be available in SOIC-8 only, so please set the correct FPFilter (see above) and the footprint field.

Best,
JAN

evanshultz:
Done. I updated the other footprints that didn't have this style of FPFilters, as well.

IRS2104: http://www.infineon.com/dgdl/ir2104.pdf?fileId=5546d462533600a4015355c7c1c31671
IR2110/2113: http://www.infineon.com/dgdl/ir2110.pdf?fileId=5546d462533600a4015355c80333167e
IRS2181: http://www.infineon.com/dgdl/irs2181.pdf?fileId=5546d462533600a401535676c12b27d3
IRS2186: http://www.infineon.com/dgdl/irs2186pbf.pdf?fileId=5546d462533600a40153567716c427ed

Isn't including the `3.9x9.9mm` part just the same as including pin number? Won't either one basically push the user into including just a single part if they're filtering by keywords in CvPcb? So adding the pin number filter won't further filter the results? I guess what I'm asking is that, if the default CvPcb filters are used (keywords and pin number), won't `SOIC*Pitch1.27mm*` give the same footprint results?

jkriege2:
Nice work!

Thanks,
JAN

PR #826, opened 2016-12-07
Added MCP6567 comparator

kajusK:
Added MCP6567 comparator

2016-12-07-180711_1366x768_scrot


jkriege2:
nice work!

Thanks,
JAN

PR #827, opened 2016-12-07
Adds a system that checks library rules automatically

Marsfan:
This pull uses Travis-CI ([travis-ci.org](travis-ci.org)), a free continuous integration service, to run checklib.py from the kicad-library-utils repo. It pulls that file straight from the repo, so it will always remain up to date. 

The script is currently executing in verbose mode, so that an explanation of each error will appear. 

Currently there are tons of parts that violate rules 3.1, 3.6, and 3.9, so the output log is large. Fixing those will also mean it only shows an error if an addition is in violation. I will probably start working on fixing those for this reason.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 before_install:
--git clone https://github.com/KiCad/kicad-library-utils
--mv kicad-library-utils /usr/local/share/kicad-library-utils
+- git clone https://github.com/KiCad/kicad-library-utils
+- mv kicad-library-utils /usr/local/share/kicad-library-utils
 
 script:
-python /usr/local/share/kicad-library-utils/schlib/checklib.py ./library
+- python /usr/local/share/kicad-library-utils/schlib/checklib.py ./library
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,5 @@ before_install:
 - sudo mv kicad-library-utils /usr/local/share/kicad-library-utils
 
 script:
-- ls .
 - sudo bash ./library-check.sh
 - python /usr/local/share/kicad-library-utils/schlib/checklib.py -v ./library/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,5 @@ before_install:
 - sudo mv kicad-library-utils /usr/local/share/kicad-library-utils
 
 script:
-- ./library-check.sh
+- sudo ./library-check.sh
 - python /usr/local/share/kicad-library-utils/schlib/checklib.py -v ./library/*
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Hey @Marsfan this is a great idea, I believe that Jon @reportingsjr was working on this a while ago, not sure how far that progressed.

Ideally, when there is a new PR it will only be checked against the components that have been either added or updated.

Have a look at the [comparelibs](https://github.com/KiCad/kicad-library-utils/blob/master/schlib/comparelibs.py) script which was written for exactly this purpose. 

If we can compare only the altered files this will reduce the noise that users have to wade through to see the errors relevant to them.

Can you provide an example (maybe a screenshot?) of your script working? Does it give the output message straight into the PR conversation?

Marsfan:
@SchrodingersGat [Here ](https://travis-ci.org/Marsfan/kicad-library)is a link from when I was testing it. 

I was planning to leave it that way, so everyone could see what needs fixing and handle it. It adds pressure to fellow developers, but I will look at changing it.

SchrodingersGat:
Looks great, but I think that a user being presented with all this info (rather than only what they have added) is very overwhelming.

Would it be very much effort to look at the diffed files, work out which components have changed (using the comparelibs script) and then only output data for those? 

If we can get this working it will be a fantastic help to library management.

Marsfan:
I will look, but I am not an expert. Give me a bit of time.

SchrodingersGat:
Neither am I but it is worth investigating. I might have some time to take a look at this after work :)

Marsfan:
Probably have it clone kicad repo and the forked repo, run a comparison, then run the final result against the two. 

Also, an admin for this repo will need to enable travis-ci for this repo once it is working.

Marsfan:
I can do it, just need some time.

Marsfan:
@SchrodingersGat I taking a look at this, why don't we run fix-pins.py and fix-text-sizes.py to fix everything, and leave the code as is.

Marsfan:
I am still going to check this though

Marsfan:
@SchrodingersGat The checklib script does not seem to work, when I pass a library that has differences I knew exist, it does not return anything. Right now the code will check everything, but that is too verbose. Can you get someone to check the checklib.py script?

Marsfan:
@SchrodingersGat Turns out I need to call sudo :(

Marsfan:
@SchrodingersGat I discovered that a command line option automatically calls check. I do not want to clog up the line with so many commits, so I am closing this pull and will open a new one when I finalize everything.

PR #829, opened 2016-12-10
Add IR2106 high and low side driver

jhartell:
IR2106 datasheet: http://www.infineon.com/dgdl/ir2106.pdf?fileId=5546d462533600a4015355c7cfc51673


SchrodingersGat:
Nice work :)

PR #830, opened 2016-12-10
added the PIC18(L)F4458 and PIC18(L)F4553 microcontrollers

nickysn:


jkriege2:
Hi!

thanks for the contribution. Please change this:
1. No leading * required for Footprint FIlters
2. Shouldn't `OSC2/CLKO/RA6` be an OUTPUT? See DS of F4553, p. 5
3. in PIC18(L)F4458-I/ML: Why is there a pin 50?


Otherwise it looks good to me.

Best,
JAN

nickysn:
1. done
2. yes, but page 15 of the datasheet states that RA6 can also be an input
3. done; this was copied from the PIC18(L)F4455, which also had this bug. I also noticed it is present in the PIC18(L)F4450, so I fixed it there as well.

jkriege2:
Hi!

and first of all: Seasons Greetings :-)))

Please have a final look at the FPFilters. They should be (but I spotted a few that differ):
  - `QFN*1EP*8x8mm*Pitch0.65mm*`
  - `TQFP*10x10mm*Pitch0.8mm*`
  - `DIP*W15.24mm*`

Best and thanks,
JAN

jkriege2:
... oh and please add a pin for the bottom pad (pin 45) of the QFN-footprints!

nickysn:
ok, done

jkriege2:
plase make PAD-pin same length as other pins.

Otherwise: looks good and I'll merge when fixed!

Best and thanks already,
JAN

nickysn:
done :)

jkriege2:
nice works
thanks,
JAN

PR #515, opened 2016-05-03
Add TCA9534 to interface

penoud:

PR #831, opened 2016-12-10
Add TCA9534 to interface #515

penoud:


jkriege2:
Hi!


I've seen, you removed an `$ENDCMP` from the dcm ... Please correct thaz:
https://github.com/KiCad/kicad-library/pull/831/files

cu
JAN

PR #832, opened 2016-12-10
rfcom: Add IQRF transceivers (DC)TR-52D and (DC)TR-72D

Roman3349:
Footprint: https://github.com/KiCad/Connectors.pretty/pull/47
Datasheets:
 - (DC)TR-52D: http://iqrf.org/weben/downloads.php?id=213
 - (DC)TR-72D: http://iqrf.org/weben/downloads.php?id=337

iqrf

Signed-off-by: Roman3349 <ondracek.roman@centrum.cz>


jkriege2:
Hi.
thanks for the contrib. Pleas change these things:
1. pin length should be 0.1"
2. as far as possible: VCC/...: top, GND/...:bottom, maybe C1/C2: left, C5-C8: right as in DS, page 4
3. Maybe add some symbol inside that signifies that this is a transceiver, I did something like that for ADCs/DACs, I think that makes schematics clearer, e.g.:
default
Does not have to be big ...
4. I've never seen a pin 0 ... maybe make it pin 1 and increase numbers for others?

Best,
JAN

Roman3349:
@jkriege2 Fixed. Pin 0 (now pin 9) is metal part of a SIM connector (KiCad/Connectors.pretty#47).

jkriege2:
hhmmmm ... what's not so good is that the long text of pin5 looks ugly ... I would propose: Either make the symbol a bit wider, or move the pins back to the right ... What do you think?
JAN

Roman3349:
@jkriege2 Is it better now?

tr-52d

jkriege2:
looks good to me now ... please move the reference point/anchor to the center of the symbol. then it should be done and I'll merge ;-)

Best,
JAN

Roman3349:
@jkriege2 Fixed. :)

jkriege2:
Great work,
Thanks,
JAN

PR #833, opened 2016-12-10
add Ads1243 and Ads7282

penoud:
pass the control with the python script.


jkriege2:
Hi!
your PR deletes several  pre-existing devices. Please that up, before we proceed with the review. Also you changed a 3D model ... why?

Best,
JAN 

PS: Maybe it would help to open a new, clean branch that does not have the additional cludder ... (e.g. https://github.com/KiCad/kicad-library/pull/833/commits/eea17997eede180a853e6d40c3e08f1eb1cf0c8b)

PR #834, opened 2016-12-11
Add PIC18F4480, PIC18LF4480, PIC18F4580 and PIC18LF4580

FabianInostroza:


SchrodingersGat:
Hi @FabianInostroza thanks for the contribution. I would suggest that this part should be named without the parentheses, as this makes it hard to search for and filter the components. The F/LF parts are also not drop-in replacements for each other so they should be separated.

e.g. I would name this:

*PIC18F4580-I/P*

with the following aliases
**PIC18LF4580-I/P**
**PIC18F4480-I/P**
**PIC18LF4480-I/P**

with each alias having a description that matches the F / FP variant.

The parts already in this library do not conform to this standard but I feel it it would be a big improvement ( the other parts will have to be changed at a later stage )

FabianInostroza:
I agree, I initially added the parts following the convention already used.

I updated all definitions in the library.

SchrodingersGat:
Oh, fantastic work :) I'll have a further check over these later on but this looks great!

FabianInostroza:
This was done partially automated, I had to remove some duplicated before running the scripts.
You may check the scripts instead of the library file.

Here are the scripts:
https://gist.github.com/FabianInostroza/9aaf0942c405579cdb3029f199df0a3d

SchrodingersGat:
You have labelled RB7 as "50" when it should be "40"

FabianInostroza:
Good catch. This error was also on the symbol of PIC18F458. I made the PIC18F4580 symbol copying that but didn't notice that error. Fixed

SchrodingersGat:
@FabianInostroza  thanks for the fix.

As we are addressing this library at large now, can I make one other request?

Some parts have the part-name prepended to the description

e.g.
```
$CMP PIC18LF14K50-E/SS
D PIC18LF14K50,16K Flash, 768 SRAM, 256 EEPROM, USB, nanoWatt XLP, SSOP20
```

It is now KLC policy to *not* have the part name included in the description - are you able to fix this so that all the parts do not have the part name in the description field? :)

FabianInostroza:
Ok, I also will update the other microchip libraries

SchrodingersGat:
HI @FabianInostroza it appears that there are now some conflicts as another PR has been merged in the interim.

Can you please fix the conflicts (there may be some stray (L) parts in there now) and I'll get this merged ASAP :)

Thanks for the great work thus far.

FabianInostroza:
Ok, it's done

SchrodingersGat:
Fantastic :)

One more thing!

If you check this library using the [kicad-library-utils](https://github.com/KiCad/kicad-library-utils) scripts, you'll see that it raises issues with the footprint fields for (most of) the symbols.

This is because the footprint should be set to an explicit footprint value, whereas currently they are set to generic text such as "TQFP-44".

The footprint should be specified as `Library:Footprint` and link to an actual footprint (use the "Select" button in the footprint field to assist here!)

For example:

## PIC18F45K80-I/PT
`TQFP44` (generic, not linked to a library) should be set to `Housings_QFP:TQFP-44_10x10mm_Pitch0.8mm`

**Note:** You should also set the footprint field to be invisible e.g.

untitled

If you can fix up the footprints so that this library passes the KLC checks, that would be fantastic.

If you have not used the scripts before, run the command:

`python checklib.py microchip_pic18mcu.lib`

(Note that a couple of the parts are flagging a `Rule 3.9` error which means they have off-grid pins)

Cheers

FabianInostroza:
Ok, but first merge this pull request. When I find time will make the library pass the checks

SchrodingersGat:
Fair enough. Thanks for your hard work.

PR #836, opened 2016-12-12
Add UCC2891/2/3/4/7

evanshultz:
Datasheet: http://www.ti.com/lit/ds/symlink/ucc2893.pdf

Passes KLC without issue.

Symbol image:
image

Here's the pin table for UCC2891 (same pin types for all but pin 16 is named differently on 2/4):
image


jkriege2:
Hi!

thanks for the contrib. I ran our scripts. Please fix:
```
checking component: UCC2891
    pin VREF (4): w (POWER OUTPUT), expected: W (POWER INPUT)
    pin CS (7): P (PASSIVE), expected: I (INPUT)
    pin VIN (16): P (PASSIVE), expected: W (POWER INPUT)
checking component: UCC2892
    pin VREF (4): I (INPUT), expected: W (POWER INPUT)
    pin GND (6): P (PASSIVE), expected: W (POWER INPUT)
```

- Also: Please group pins logically, where applicable (inputs: left, outputs: right, VCC...: top, GND...: bottom)
- FPFilters hould be `SOIC*3.9x9.9mm*Pitch1.27mm*` and `TSSOP*4.4x5mm*Pitch0.65mm*`

Best & thanks,
JAN

evanshultz:
Oops. Thanks for catching my mistakes.
- VREF is the output of an internal voltage regulator, so POWER INPUT seems wrong to me.
- VIN is stood off from a high voltage by high-value resistors. Eeschema manual states "Power pins are automatically connected to the other power input pins with the same name". So is POWER INPUT correct?

I made the other changes.

BTW, what tool are you using? checklib.py, even in verbose mode, doesn't show this info.

Shall I also create [UCC2897](http://www.ti.com/lit/ds/symlink/ucc2897.pdf), a superset part in the same family?

evanshultz:
Hi Jan,

Please check out the FPLIST. I assume you meant for the asterisks to be there? But if so, won't that allow more package choices which may not be appropriate? I updated the patch with just two options, SOIC and TSSOP. Please review.

Thank you!

jkriege2:
Hi!

I meant it as I put it above ... The idea is:
1. There is a filter for the number of pins. 
2. The info in the FPFilter + the pin-count should match only the matching packages ...

About the tool: I'm using checklib.py ... but with `-vvv` ;-) ... give it a shot that should increase verbosity even further ...

In `UCC2892` VREF is now an INPUT, but POWER_OUTPUT or OUTPUT should be correct (as you stated above). VIN is OK for me now.

About UCC2897: Go ahead if you have time ,-)

Best,
JAN

evanshultz:
Fixed VREF pin on UCC2892 and added UCC2897. Symbol:
image

Pin table:
image

I know the pin grouping violates KLC, but otherwise is it good? Thanks so much for your time and instruction. :)

jkriege2:
looks good to me now ... any reason why you want to retain the pin grouping as in the package?

Best,
JAN

evanshultz:
IMO, a black-box part has a datasheet so an engineer can find out what's inside. For manufacturing/service folks, having the pinout this way has proven to be an advantage because it makes it easier to check/test the product since there's a strong correlation between the schematic symbol and the physical package. But it's different perspectives. I tend to put more burden on the engineer and less on manufacturing/service in the interest of expediency where it really matters.

And I also detest vertical text and KiCad doesn't allow the pin text to be rotated (unless I'm daft).

But all that being said, KiCad has rules and and I must follow them, even if the topic is a matter of taste simply to establish a standard. I wanted to get everything else straightened out before seeing about a pin placement that follows KLC. Now that everything else is done, unless it's accepted to keep this pinout, I'll start experimenting with different pin placements in the interest of getting this PR merged.

SchrodingersGat:
I would also vote for adjusting pin placement to match KLC suggestions. It greatly improves schematic readability to have pins grouped by function rather than numbering.

Thanks, Oliver

evanshultz:
With #823 having been merged, I've modified the symbols here in a similar fashion and the FPLIST looks good to me. Thanks for the review!

Old on top and new on bottom:
image

evanshultz:
Voltage input on top, ground on bottom, outputs on left along with CS which is in a convenient position for the source of the main FET, and inputs are on the left with ground-referenced pins at the bottom for simple connection to ground.

jkriege2:
Hi1

nice work, I spotted only one minor thing:
- UCC2897 has still the pin-number in the QFN FPFilter

best,
JAN

jkriege2:
Great!

Thanks,
JAN

PR #837, opened 2016-12-13
Add current sense transformer

evanshultz:
This is a current sense transformer symbol for a part that is rather common for currents at 10-20A. See the following datasheets:
1. http://products.pulseelex.com/files/product_files/PC2005_47.pdf
2. http://products.pulseelex.com/files/product_files/SPM2007_58.pdf
3. https://static1.squarespace.com/static/56ff90f3d51cd4118179ec49/t/571da42a45bf21e6a9e47e8b/1461560366359/CT01.pdf
4. http://www.datatronics.com/pdf/ct319.pdf

Took a bit of KiCad-fu with a text editor to get the drawing elements onto Units A and B properly, but it passes all KLC checks.

The name is fairly ugly. Any suggestions here are welcome.

Symbol images:
image

Pin table:
image

This symbol matches with my PR for the footprint at https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/30.


jkriege2:
why two separate symbols? 

JAN

evanshultz:
Where current is sensed on the schematic may be quite far from where the measured current is used on the schematic. Having separate symbols allows for much cleaner schematic design in this circumstance.

This is more of an issue with larger and more complex schematics, but I use separate symbols exclusively because it dramatically improves readability (there is an element of subjectivity here). I can find lots of examples online of very simple circuits where having a single symbol is satisfactory. But I believe that in many professional/proprietary schematics, which aren't so easy to find online, separate symbols are preferred.

For example, take a look at how ugly the schematic on page 27 of the [UCC2893 datasheet](http://www.ti.com/lit/ds/symlink/ucc2893.pdf) looks with the current sense transformer traces traipsing all over the schematic page.

evanshultz:
As opposed to the discussion over a coupled inductor ([#890](https://github.com/KiCad/kicad-library/issues/890)) where separate units seemed to be accepted as equal with a single combined one, this PR has clearly different symbols for each unit so confusion is unlikely to happen.

Will this get merged in its current state? Would both a unified and separated symbols need to be made for this current sense transformer? I can do that.

SchrodingersGat:
@evanshultz having both a split and unified option seems like a good idea

evanshultz:
Hi folks! I've added a new split symbol matching the style of EMI_Filter_LL from device.lib to gather some feedback before finishing this and making a unified version:
image

**Naturally, only a single split and unified symbol will be pulled. This is just for feedback.**

Questions:
- General comments? (Image updated above.) Should it use the bigger arcs of these smaller ones?
- The symbol name is ugly. Suggestions?
- Is `transf.lib` the right library for this part? `device.lib` with all the other inductors?
- The pinout is for the footprint I created and mentioned in the first comment. Should this symbol then be specific for only that component as well, instead of being generic?

jkriege2:
Hmmm ... for reference an overview of the different symbols:
2017-01-26 22_26_21-eeschema kicad-library-review _ d__kicad_kicad-library-review-project

1. I would maybe vote for the new version ... but please move the hor. lines a bit closer together ... I think that would look better?
2. I would put this as a generic symbol (maybe with pins A: 1-2 and B: 3-4) and then add specialized symbols for the special trafos from the datasheets ... as these types of trafos often have very different pinouts:
- http://www.coilcraft.com/pdfs/sen60tr.pdf
- http://www.coilcraft.com/pdfs/sen60.pdf
- http://www.coilcraft.com/pdfs/senhi.pdf
- http://www.murata-ps.com/data/magnetics/kmp_5600.pdf#toolbar=1&navpanes=0
- http://www.coilcraft.com/pdfs/cst7030.pdf
- http://www.coilcraft.com/pdfs/senhitr.pdf
3. Name is uggly ... but I would leave it as long, but change the order (for sake of sorting): `Transformer_Current_Sense` and `Transformer_Current_Sense_Split`
4. I would put the generic symbol into `device.lib` and the special ones into `transf.lib`.

Best,
JAN

PS: We also have to clean up a bit in `transf.lib` ;-) ... just seeing, all the generic trafos should be unified and move to `device.lib`.

evanshultz:
Thanks for the review and suggestions.

I just pushed an update, with both symbols of the specific 8.4mmX7.2mm size. I'm using the symbol version with bigger arcs because that matches the default inductor symbol, as far as I can see.

The NC pins are ugly, but if they're away from the part body I'm afraid they'll get accidentally connected.

@jkriege2 If this is good, I'll push the generic symbol and DCM updates to close this PR.

evanshultz:
@jkriege2 @SchrodingersGat Would you have a chance to take a look at this PR? Thank you!

SchrodingersGat:
1. The .dcm file updates to not match the .lib file updates
1. These are supposed to be generic symbols - why do they have dimensions?
1. As they are generic (not associated with a particular footprint or part-number) should they instead go in device.lib
1. Should we rename transf.lib to transformers.lib

jkriege2:
@SchrodingersGat @evanshultz What's the status of this PR?

evanshultz:
@jkriege2 I'll update in the next couple days.

jkriege2:
OK, I have this on my watchlist ;-)

evanshultz:
Thanks for keeping an eye on my, Jan! :)

Going back to @SchrodingersGat's questions:
1. Yes. I'm soliciting feedback on symbols before DCM updates. See above.
2 & 3. Right. I want to decide on the symbols, then have the generic one and a specific one.
4. I think it should be spelled out.

In all cases, the discussion on Jan 26 is the last real progress on the symbols. The first decision to close this is on the symbol style. Jan likes the smaller arcs (from `EMI_filter_LL`) while I'm kinda in the middle but a little bit closer to the older style with bigger arcs. Can we get some additional feedback to help decide which symbol style to use? Once that's done, I believe it's straightforward to finish off this PR.

@jkriege2 @SchrodingersGat @poeschlr

jkriege2:
Looking at the symbols, we should also decide on the pin-out ... For the actual parts they are clear (i.e. as defined by the footprint), but for the general parts, should we use tha same pin-out as the EMI_Filter_LL? I think that makes sense ... especially as the choice of pinout is arbitrary for them anyways.

About the style I'm really not sure (also while having a second look after months now ;-) ... Maybe we should orient on the `L_Core_Iron`-device ... somehow I feel that the transformer loops are too large, but maybe the filter loops are too small ... and `L_Core_Iron` is somewhere in between ... This is a bit strange of an argument, but when thinking of transformers I usually think HUGE (i.e. for power-source applications), while for filters I think of small ... current-sensors then would be inthe middle of the two???

Does that make any sense? Does it help?

Best,
JAN

SchrodingersGat:

PR #838, opened 2016-12-13
Add spark gap

evanshultz:
Spark gap created in accordance with the IEEE 315-1975 standard using two unfilled triangles and an "E" ref des prefix. Passes all KLC tests.

Symbol image:
image


SchrodingersGat:
Nice work @evanshultz - I would suggest that this should go in the ESD_protection.lib library rather than generic devices.

jkriege2:
Hhmmm ... I think this is generic enough for device.lib ... in ESD_protection.lib we have mainly specific devices (with part number and all) ...

cu
JAN

evanshultz:
Thank you @SchrodingersGat and @jkriege2.

I'll move it to ESD_protection if that's the consensus, but it is generic which is why I went with device.lib for the PR. Please bring in the folks that should consider this topic so we can discuss.

SchrodingersGat:
device.lib should be fine :) @evanshultz  can you please add some extra tags to assist with searching e.g. "esd electrostatic" etc

evanshultz:
Done! Anything else?

SchrodingersGat:
Looks great :)

PR #840, opened 2016-12-13
Add dual BJT symbols

evanshultz:
These symbols are suitable for 6-pin footprints like SOT363 containing a pair of BJTs (dual NPN, dual PNP, and NPN + PNP). Here are some example part datasheets:
NPN/NPN (BC846BS): http://www.nxp.com/documents/data_sheet/BC846BS.pdf
PNP/PNP (BC856BS): http://www.nxp.com/documents/data_sheet/BC856BS.pdf
NPN/PNP (BC846BPN): http://www.nxp.com/documents/data_sheet/BC846BPN.pdf

All symbols pass KLC.

Symbol images:
image


jkriege2:
I would make them as concrete BC846BS... devices in `transistors.lib`, as the pinout seems to be quite special to me ... and then we would need all sorts of variants with mixed pins ...

JAN

evanshultz:
In my experience this is a common package for dual transistors. I gave examples from NXP above and here's [BC846BDW ](http://www.onsemi.com/pub_link/Collateral/BC846BDW1T1-D.PDF)from On Semi, [FFB3904 ](https://www.fairchildsemi.com/datasheets/FM/FMB3904.pdf)from Fairchild (note FMB3904 has a different pinout), and [DMC20402 ](http://www.semicon.panasonic.co.jp/ds4/DMC20402_J.pdf)from Panasonic.

Here are some packages from the above that fit this symbol:
-SOT363
-SC70-6
-SC74-6

Certainly there are many variations and packages of transistors which aren't compatible, and if there's a better way to make the parts (either as generic or specific) or classify them I'm open to changing my PR. I do see this type of part used frequently in my sliver of the electronics world and as you can see above there are multiple packages and types and vendors who support these symbols.

jkriege2:
Hhmmmm ... thinking about this I would rather have the real transistors + the generic symbols (I think I wouldn't search for the generic parts, but first search for the actual devices (or examples thereof), but that's how I work ... 

What I would propose:
1. We keep the generic symbols, but they should be renamed in the style of the other generic transistors, i.e. with a suffix that denotes the pin-order, e.g. `Q_DUAL_NPN_NPN_E1B1C2E2B2C1`
2. add some of the transistors you named above as specific symbols to `transistors.lib` with the same symbol (i.e. a copy) ... I think we can do all this in one PR ;-)

Best,
JAN

PS: @diggit @SchrodingersGat @pointhi @CarlPoirier What do you think?

evanshultz:
Makes sense to me. Thanks for the feedback. I'll go ahead and rename these generic symbols as mentioned above.

For the specific devices, I'll get to them later and then the PR can be reviewed by the other folks too.

evanshultz:
Continuing on the path set above, here are the specific transistors I'm planning to add to `transistors.lib`. Any problems? And can these all stay in the same PR?

- [BC846BS]
- [BC846BDW1] 
- [BC847BS]
- [BC847BDW1] 
- [BC856BS] 
- [BC856BDW1]
- [BC857BS] 
- [BC857BDW1]
- [BC846BPN] 
- [BC846BPDW1] 
- [BC847BPN] 
- [BC847BPDW1]
- [PMBT2222AYS]
- [MMDT2222A]
- [MBT2222ADW1T1] 
- [FFB2222A]
- [PMBT3904YS]
- [MMDT3904]
- [MBT3904DW1]
- [FFB3904]
- [PMBT3906YS]
- [MMDT3906]
- [MBT3906DW1] 
- [FFB3906]
- [PMBT3946YPN]
- [MMDT3946]
- [MBT3946DW1T1]
- [FFB3946]
- [MMDT5401] 
- [MMDT5551] 
- [FFB5551]

jkriege2:
Hi!

symbols look good to me now ... What about adding `, 6-pin package, e.g. SOT-23-6` or something like that to the description? When done, I'll merge!

About `transistors.lib`: Go ahead ... If you do a single PR (there should be 3 symbols at max, right? They are all in SOT-23-6 so most can be ALIASes ;-) I'll review+merge that.

Best,
JAN

evanshultz:
I appended the generic ", 6-pin package" to the description since there are multiple packages that could be used.

evanshultz:
Added the specific parts into `transistors.lib`. There are a number of packages, but you're correct that there's only 3 symbols and the rest are aliases. This lib has loads of KLC violations, but none I see related to the dual transistors.

I also added SC* footprints to the FPLIST for all symbols since you can see that as a valid footprint prefix I noted above.

Anything else, or is this ready to merge?

jkriege2:
Hi!

nice work already ... Could you still please change this:
0. `device.lib` looks good to me!
1. add datasheet links to the actual transistors
2. all the devices that I found are in SOT-363/SC70-6 package ... so could you simply set that package to e.g. `TO_SOT_Packages_SMD:SC-70-6` and set the FPFilter to e.g. `SC?70*` ... BTW: I found either the code SC-70-6 or SOT-363 for the transistors ... and this document (http://www.onsemi.com/pub_link/Collateral/419B-02.PDF) gives the same drawing for both ... is there any difference? (this also concerns your PR https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/26)

Best,
JAN

evanshultz:
Interesting. The SOT363 given by [NXP](http://www.nxp.com/documents/outline_drawing/SOT363.pdf) is quite a bit different. Long thin pins are replaced by square ones. [Central Semi](https://www.centralsemi.com/PDFs/case/SOT-363PD.PDF). [Vishay](http://www.vishay.com/docs/84015/sot363.pdf). [Diodes Inc](http://www.diodes.com/_files/packages/SOT363.pdf).

Here's one from [Linear Tech](http://cds.linear.com/docs/en/packaging/SOT_6_05-08-1638.pdf) that doesn't match either of those two. [Fairchild ](http://www.icbank.com/icbank_data/semi_package/sc706_dim.pdf)is different. So is [Vishay](http://www.vishay.com/docs/72602/72602.pdf).

Are they all compatible? Probably, but they're not the same footprint. The component package is virtually identical in all cases.

So it seems to me this is manufacturer-specific. What to do? I can make any changes that's best for KiCad.

evanshultz:
And datasheets added.

jkriege2:
Hi!

I'll answer to the SOT/SC-questions in the thread there ... When we cleared that up, I would suggest to select ANY matching package (maybe depending on how the manufactrer calls the package in the transistor datasheet and set it here as package + set the FPFilter to match SOT?363* and SC?70*.

OK with you?

cu
JAN

evanshultz:
It appears to me that SC70-6 is one common package, and SC88-6/SOT363 is another common package with two names. Yes?

So we could make FPLIST contain the following:
 SC?70*
 SC?88*
 SOT?363*

And I like your idea to, by default, select the package name the manufacturer gives on the datasheet.

Is that good?

jkriege2:
perfect!

JAN

evanshultz:
FPLIST updated as discussed above for both specific and generic dual BJT devices. Ready for merge?

jkriege2:
Hi!

since all actual transistors seem to be in SOT-363 package, please set that package in the footprint-field (i.e. not in device.lib, only in transistors.lib). Rest looks fine to me.

Best,
JAN

evanshultz:
After an "oops" commit, the last patch should have it all. Thanks so much, Jan!

jkriege2:
Nice work, Thanks!

JAN

PR #839, opened 2016-12-13
Add NXP 4- and 5-unit PESD* ESD protection diodes

evanshultz:
Datasheets:
-http://www.nxp.com/documents/data_sheet/PESDXL4UF_G_W.pdf
-http://www.nxp.com/documents/data_sheet/PESDXL5UF_V_Y.pdf

Passes all KLC tests. Common anode on all symbols.

Symbol image:
image

If this should go in a different library, like ESD_protection, please let me know. I'm happy to update.

PR #841, opened 2016-12-13
Add NXP 4- and 5-unit PESD* ESD protection diodes in ESD_protection.lib

evanshultz:
Datasheets:
-http://www.nxp.com/documents/data_sheet/PESDXL4UF_G_W.pdf
-http://www.nxp.com/documents/data_sheet/PESDXL5UF_V_Y.pdf

Passes all KLC tests. Common anode on all symbols.

From PR #839.


SchrodingersGat:
Nice work

PR #842, opened 2016-12-13
xilinx: Add XC2C256-TQ144

piotr-wiszowaty:


jkriege2:
nice work, thanks!
JAN

PR #850, opened 2016-12-16
Update KiCad Library with TI TPS2561-Q1 current limited power switch.

BradWalker:


SchrodingersGat:
Hi @BradWalker thanks for the contribution. There are a few areas here that could be improved to match the library conventions.

Firstly, are you aware of the [Kicad Library Convention](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention)? This outlines the standards for developing symbols and footprints, and it's useful to have a general knowledge of these.

## KLC3.4 - Group Pins Logically
It is (generally) better to arrange pins via logic function rather than footprint pin position. e.g. keep GND pads at bottom of part, INPUTS on left, OUTPUTS on right

## KLC3.6 - Field text uses a common size of 50 mils
Ensure that all fields use a 50 mil text size

## KLC3.2 - Pin Size
The pin size you have chosen (0.2") should be reduced to 0.1" 

## Other notes
- You should add the GND pad (pin 11)
- GND should be specified as **Power Input** type
- Set the width of the box to 0.01" and set the fill type to "background"

e.g.

untitled

Thanks :)

BradWalker:
Will create a new pull request..

PR #851, opened 2016-12-17
Fixes for Tantalum Capacitor 3D models

SchrodingersGat:
* Updated to match revised footprints
* Added step / wrl file pairs
* Fixed rotation to math IPC (pin 1 on left)
* Removed .wings files

As requested on the [forums](https://forum.kicad.info/t/footprint-and-3d-mismatch/4644/2)

This PR provides new models for the Tantalum capacitor footprints, in accordance with the updated footprints. Models  were sources from https://github.com/easyw/kicad-3d-models-in-freecad and adjusted to match the footprints.

The new models are provided in STEP/WRL pairs, and the old .wings files are gone.

e.g.

tant


jkriege2:
Hi!

OK, they look very nice (thanks) ... Do you also plan to change the .pretty-lib? If so I would propose to make a PR for that and we merge them both together ...

cu
JAN

SchrodingersGat:
Jan,

Such a fix has been made (you merged it! ;) )

https://github.com/KiCad/Capacitors_Tantalum_SMD.pretty/pull/8

jkriege2:
OK ;-)

nice work BTW!

JAN

PR #852, opened 2016-12-17
JP patch

SchrodingersGat:
See https://lists.launchpad.net/kicad-lib-committers/msg00456.html


SchrodingersGat:
Oops, did that wrong...

PR #853, opened 2016-12-17
JP template patch

SchrodingersGat:
This PR applies two patches submitted to the kicad-lib-committers mailing list.

Reference: https://lists.launchpad.net/kicad-lib-committers/msg00456.html


SchrodingersGat:
Merging as no objections

PR #855, opened 2016-12-20
made PWR_FLAG smaller (comparable to other power symbols)

jkriege2:
The PWR_FLAG-symbol was way larger than the other power symbols ... I fixed this

2016-12-20 07_44_30-eeschema test_addlin _ d__kicad_test_addlin

Former symbols:
2016-12-20 08_10_29-eeschema test_addlin _ d__kicad_test_addlin


SchrodingersGat:
Nice addition Jan, thanks

PR #856, opened 2016-12-20
Add MC79L00 family in SO-8

evanshultz:
[MC79L00 family](http://www.ti.com/lit/ds/symlink/mc79l05a.pdf) based on existing MC78L05ACD symbol (and keeping the practice of pin non-consolidation). No KLC violations.

Symbol image:
image


jkriege2:
Hi!

nice work. Please change:
1. add NC-pins (invisible)
2. Please add datahseet links to all devices
3. FPFilter should be `SOIJ*5.3X5.3mm*Pitch1.27mm*`

Thanks,
JAN

evanshultz:
Thank you Jan!

1. Gotcha. Done. (I can make a PR for MC78L00ACD.)
2. This is a generic part, made by TI, Fairchild, ON Semi... Do I simply choose one vendor's datasheet? Seems unfair.
3. When I read KLC 3.9 I understand I should fully qualify the footprint path and name, like what I've done. There is no violation of KLC 3.9 in this symbol. It seems using the asterisks is a more flexible solution but if there's only a single "correct" footprint, why not fully qualify it? There must be a good reason so please help me understand. Thanks again!

jkriege2:
Hi!

ad 1) you can also put it in this PR

ad 2) Just choose any you like (maybe the one you like best, because it contains the most info) ... it's for reference, and we don't have to be fair ;-)

ad 3) Yes the KLC are not specific here. SO it's more best practice: Usually we have BOTH. This way the part has a default footprint, but the filters still apply, so you can easily choose any matching variant (e.g. `_HandSoldering` is available for some devices)

Best,
JAN

evanshultz:
Hi Jan.

1. Actually, there's so much wrong with MC78L05ACD I'd rather do it separately.
2. Done.
3. Is this correct?

What's next?

jkriege2:
ad 1) OK!
ad 3) perfect ...

One last minor thing: could you reduce the pin length to 0.1" that should be way enough ;-)

Best,
JAN

evanshultz:
I aligned the pins on grid for MC79L05ACD.

Then, I decided to look into KLC violations of the existing regul.lib. I updated 7805, which looks fine to me, and also MC78L05ACD, which I think is too narrow. What do you think, @jkriege2? If we find something that works, I might as well go ahead and clean up regul.lib.

jkriege2:
Hi!

..taking a closer look at reul.lib: It is a nightmare ;-) ... so any effort to get it (at least somewhat) KLC-conforming would be appreciated a lot. Things I spotted:
1. 7805:
  1. there are no pin-numbers assigned i.e. IN=1, OUT=3, GND=2
  2. DS-link is not to the WWW ... Maybe use this one: http://www.ti.com/lit/ds/symlink/lm7805.pdf
  3. You altered symbol looks much better :-) ... maybe reduce the text offset from 50mil now to 10-20mil, so the text is closer to the pin/border
  4. symbol outline should be 0.01" thick
  5. no FPFilter, I think we should have `TO?220*`, `SOT?223*`, `TO?263*`, `TO?3*`
2. 78L05:
  1. There should be (at least) two variants, one `78L05ACZ` for TO-92 and one `78L05ACM` for SO-8 (based on this DS: http://www.ti.com/lit/ds/symlink/lm78l05.pdf) with matching FPFilters and footprint-fields set
  2. set DS-linK: http://www.ti.com/lit/ds/symlink/lm78l05.pdf

Basically: `MC78L05ACD` seems a very good starting point! For the size 7805 with 10/20mil offset seems good for me. Here's a size comparison to other symbols in a simple circuit (I like the relative sizes ...):
2016-12-23 13_23_05-eeschema kicad-library-review _ d__kicad_kicad-library-review-project

But seeing all this I would propose to merge your initial changes for 79Lxx and do the other stuff in a second PR ... OK?

Best,
JAN

evanshultz:
Fine with me. Can you pick commit f5146692fc703e82d5008891b1e8321a94cb1609 to address the negative regulator and we'll deal with the existing library symbols at a later date? Thanks for the feedback, by the way. :)

evanshultz:
Done. I think we're good! :)

jkriege2:
OK. Thanks ... then we can do the workover of regul.lib in a second PR ;-) ... Should we maybe share the work? I'll have some time during the next days ...

Best,
JAN

PR #857, opened 2016-12-20
memory: add AS6C1616

piotr-wiszowaty:


jkriege2:
Hi!

nice work ... since there is only one package ... please set the package-field accordingly (FPFilter can stay as it is ;-)

best,
JAN

piotr-wiszowaty:
All right, but which package should I select (I can't find the right one 
in the library)?

jkriege2:
OK, I see the problem ... So two options (I would say):
1. create the package in a second PR and I can merge them both
2. leave package empty and set the FPFIlter to something that SHOULD match a future package, e.g. `TSOP*Pitch0.5mm*` and a second filter 'TSOP*12x18.4mm*Pitch0.5mm*'

Best,
JAN

piotr-wiszowaty:
I've PR-ed a 12x20 mm package - that's what they say in the datasheet, but
the package is in fact - as you wrote - 18.4 mm wide.
Should I close the PR and create a new one with renamed package?
Or maybe PR the package under two names?

jkriege2:
just rename the file/package inside the PR (ans possibly the PR).
Best,
JAN

piotr-wiszowaty:
Done (+ changes in the symbol).

jkriege2:
Hi!

and first of all: Seasons Greetings :-)))

Please remove the pin count from the FPFilters. Theres a separate filter to match that (that's also why we need the NC-pins! and remove the comma from the keywords (they should be separated by whitespaces)

Best,
JAN

piotr-wiszowaty:
Done. Merry Christmas!

jkriege2:
Nice work!
JAN

PR #861, opened 2016-12-24
Add ADM2587E/2582E isolated RS422/485 transciever

El3ctr0nic:


jkriege2:
Nice work! Thanks,
JAN

PR #864, opened 2016-12-27
Adding Bluetooth module Laird BL652

cepr:


jkriege2:
Hi!

first of all: thanks for the contrib. Generally we have a set of rules that symbols for this lib have to follow (KLC): https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention There are also some python scripts that can check some of the rules and even fix some bugs: https://github.com/KiCad/kicad-library-utils

Some general remarks on your symbols:
1. please ensure that description, keywords, datasheet link are set in all symbols
2. align pins on a 100mil grid. Also the usual pin-distance (within a block) should be 100mil.
3. outline of black-box symbols should be 0.01"
4. Please also add footprint filters, such as `SOIC*3.9x8.7mm*Pitch1.27mm*`


About the 3D model: What is the source for this model? Did you draw it yourself? If so, could you provide the soruces, if not: Can you provide info on whether we are allowed to host that model here?

Best,
JAN

cepr:
OK, I will close this pull request and create a new one.

PR #866, opened 2016-12-28
Adding MCP2050, PIC16F1829LIN and BL652

cepr:


jkriege2:
Hi!

OK, so I'll review all three here ;-)

## BL652:
1. please copy datasheet from datasheet field (we don't populate it) to the datasheet property (the the PDF-button in the lib-editor is available!

Rest looks good to me!

## MCP2050
1. datasheet as above
2. please move power-supply pins to the top/bottom border of the box


## PIC16F1829LIN
1. please move this device to the library microchip_pic16mcu where there are all the other PIC16-variants!
2. datasheet as above
3. remove digikey field
4. modify description to match other PIC16F MCUs (give flash size ...)


Otherwise: THanks for the work you put in :-)))

best,
JAN

jkriege2:
Hi!

a short note: When you update something, please also post some message here, because otherwise we don't get noted properly (or think, the commits aren't doen yet ;-)

JAN

jkriege2:
Now for the review:

## PIC16(L)F1829
- please add keywords

## PIC16F1829LIN
- remove IC name PIC16F1829LIN from description

## MCP2050
- please empty the footprint field as this version has two possible packages and add an FPFilter `DIP*W7.62mm*`
- add `, PDIP/SOIC` to the description
- you could also think about moving the left sid of the symbol a bit in to make it smaller (just a suggestion ... maybe it looks better?)

## BL652
- set FPFilter to `Laird*BL652*`

Best,
JAN

cepr:
Hi Jan,

I will add more comments in the future.

I just learned about aliases so I changed the name of the modules to be more accurate and allow multiple packages. I also fixed some footprint filters and description and took your review into account.

It's ready for a new review, thank you in advance!

## PIC16(L)F1829

* Added the QFN package
* Added aliases for all variants
* Bigger symbol to accommodate the longer names
* Fixed description, footprint filter

screenshot from 2017-01-08 15-30-03

## PIC16F1829LIN

* Renamed to PIC16F1829LIN-E/SS to match the same naming convention than other microcontrollers

## MCP2050

* Added the QFN package
* Added aliases for all variants
* Bigger symbol to accommodate the longer names
* Fixed description, footprint filter

screenshot from 2017-01-08 15-25-26

Best,
-Cedric

jkriege2:
Hi!

OK, this is going along nicely ;-)

Some comments on the status:
1. if you add one symbol per package, the explicit package should be put into the footprint field for each symbol. Then this footprint is used by default for the symbol!
2. Please remove the pin-number form the FPFilter, e.g. alter `DIP-14*W7.62mm*` to `DIP*W7.62mm*`
3. please make the style of `PIC16F1829LIN-E/SS` to match that of `PIC16(L)F1829-I/ML`...
4. `MCP2050-330-E/MQ` is missing pins 6,9,10,11,16,19,20 (add them as invisble NC)
5. add a pin21 "PAD" for the bottom pad of the QFN-package

Rest looks OK!

Best,
JAN

cepr:
Hi Jan,

> if you add one symbol per package, the explicit package should be put into the footprint field for each symbol. Then this footprint is used by default for the symbol!

One package can still have multiple footprints (e.g. `DIP-14_W7.62mm` or `DIP-14_W7.62mm_LongPads` for MCP2050-330-E/P), which one should I use as the default one?

> Please remove the pin-number form the FPFilter, e.g. alter `DIP-14*W7.62mm*` to `DIP*W7.62mm*`

How does Kicad selects the footprint with the right number of pins?
 
> MCP2050-330-E/MQ is missing pins 6,9,10,11,16,19,20 (add them as invisble NC)

For my understanding, what is the purpose of adding the NC pins?

cheers

cepr:
Done

## MCP2050 update

* adding missing NC pads
* adding default footprint for the QFN version

 ## PIC16F1829 update

* updating style of PIC16F1829LIN to match the other PIC16F1829
* adding default footprints
* updating footprint filters to remove the pin count

jkriege2:
Hi1

Please split the -E/P and -E/SL-versions of MCP2050 into separate symbols so you can assign an appropriate default footprint (take the standard version, as default, the `_LargePads` are specializations) to each. Also note that there ist (I think) stil a `-14` in the DIP filter.

Otherwise it looks good to me ... thanks for all the work an patience you put into this :-)

best,
JAN

cepr:
done

jkriege2:
nice work!

Thanks,
JAN

PR #867, opened 2016-12-29
Added MCP6L91X and MCP6566X

kajusK:
Added SOT-23-5 comparator and op amp.

@jkriege2 - it seems recent commit broke line endings (\n vs \n\r) - https://github.com/KiCad/kicad-library/commit/3aaf88778e4ef3e6f05e340719c9d3081e57c270. Kicad changed it back when I created the new parts, so I had to insert new lines manually to avoid changing lines with incorrect line ending.


jkriege2:
Hi !

nice work ... some changes please ,-)
1. FPFilter should be `SOT-23-*'
2. remove commas from keywords (space as separator!)
3. MCP6566: add FPFilter SC?70* and remove package, as there are two variants!

Otherwise: nice work!

best,
JAN

kajusK:
1. Didn't know that, maybe it should be mentioned in rules or somewhere
2. I missed that one in the KLC, check script didn't see that as a rule violation
3. Oops, fixed :)

jkriege2:
nice work!

Thanks,
JAN

PR #870, opened 2016-12-30
Adding TPS2042D - dual version of existing quad component

scott-t-wilson:
TPS2042D is the dual version of the TPS2044D Quad power-distribution switch.

This component is a copy of the TPS2044D adjusted for package/pinout.


jkriege2:
Hi!

nice work, thanks! Please change two things:
1. switch footprint field to invisible
2. (suggestion): maybe move GND-pin to the bottom border of the symbol ... usually power pins are on top (VCC...) or bottom (GND...)

Best,
JAN

scott-t-wilson:
I'm happy to move GND to the bottom, it does make sense. I copied the existing quad part TSP2044, I can move the two new parts, or all three. Let me know if you have a preference.

SchrodingersGat:
Hi @scott-t-wilson  if you are willing to fix the quad part too (if that is what you are asking?) then that would be great :)

SchrodingersGat:
Great work, thanks :)

PR #872, opened 2016-12-30
Added LPC11E12FBD48 componet

michalkona:
Added LPC11E12FBD48 component with aliases LPC11E13FBD48 and LPC11E14FBD48 to nxp_armmcu library.


jkriege2:
Hi!

nice work, thanks. Please change these things, before we can merge this PR:
1. provide datasheet-link+keywords for the two aliases (description is there)
2. make the two NC pins invisible

Rest looks good!
best,
JAN

PS: We have some python scripts that can check new symbols and point you to some errors (as e.g. the two above, or missing pins): https://github.com/KiCad/kicad-library-utils

michalkona:
Sorry!!!, I know the kicad-library-utils,  but I did not use it because I thought that keeping rules is piece of cake for me.

jkriege2:
Looks mostly good to me, one minor hange please: could you move all the pins on the right by 100mil more to the right, so the VDD does not collide with the pin label?
2017-01-08 13_42_28-part library editor d__kicad_kicad-library-review_library_nxp_armmcu lib

best,
JAN

jkriege2:
nice work.
Thanks,
JAN

PR #874, opened 2016-12-31
Add Vishay VO615 optocoupler family

evanshultz:
When looking to add this opto, I found that there's no generic opto with no base connection in a 4-pin symbol/footprint. Similar to LTV-356T but just 4 pins. [Vishay VO615A](http://www.vishay.com/docs/81753/vo615a.pdf) is the specific part I wanted where a generic opto wasn't available (at least not that I could find).

I'm submitting the specific part in this PR. I can add a generic part as well, or remove this specific part and just submit a generic one. Let me know.

Symbol and pin types:
image

Please pay particular attention to the footprint and FPLIST to make sure this is correct. I've taken into account the latest updates for SMD DIP footprints from @jkriege2.

Thanks!


jkriege2:
I think the specific part is OK ... Maybe a suggestion: Change the LED to the style we use now in device.lib:
2016-12-31 16_08_43-library browser d__kicad_kicad-library-review_library_device lib

Rest is good1

Thanks,
JAN

jkriege2:
ahhh ... and FPOFilter should be DIP*W7.62mm*SMD*

evanshultz:
Would it make sense to add a generic also? Or just this?

Also, I appreciate all of your help and instruction. This PR went fairly smoothly because you've spent time to teach me the past few weeks. Happy New Year!

Photodiode style:
Was there only part of the symbol you were thinking should be updated? Since the size of the diode symbols are different I assume you're talking about only the style and not the size. Just making it an unfilled triangle with a line through it?

FPLIST:
The rendering on GitHub messed up the FPLIST proposal, which for clarity you wrote `DIP*W7.62mm*SMD`. Is there a reason to restrict the size and not restrict the pin count? There are a number of packages available for this part and all seem equally valid. It seems odd to filter out the wider footprints by default and not restrict the pin count like I've seen elsewhere.

Also, is there a guide to FPLIST syntax? It seems to match a regular expression with regards to `*` and `?`, but I otherwise haven't seen any documentation.

evanshultz:
@jkriege2 How about that diode style? Is this what you were thinking about?
image

jkriege2:
Yes the PNg looks good. I'll complete this review in the next days!

Best,
JAN

jkriege2:
Hi!

please also add the FPFilters
- `DIP*W7.62mm*` 
- `DIP*W7.62mm*SMD*`
- `DIP*W9.53mm*SMD*`
- `DIP*W11.48mm*SMD*`
as the datasheet also says there is a simple DIP version. ... and remove `DIP?4*SMD*`
Also empty the footprint field.

Otherwise it looks fine!

best,
JAN

evanshultz:
Thanks Jan! I also added a 10.16mm lead spacing DIP package, which is in the datasheet, and I'll create a separate PR to merge that footprint into the library shortly.

jkriege2:
OK, great work,
Thanks,
JAN

PR #876, opened 2017-01-01
Added LTC2508-32

milosz88:
Linear Technology LTC2508-32
32-Bit Oversampling ADC with Configurable Digital Filter


SchrodingersGat:
Hi @milosz88 I would firstly direct you to the [Kicad Library Convention](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) which provides some guidelines to creating symbols.

The [Kicad Library Utils](https://github.com/KiCad/kicad-library-utils) are Python scripts which automate KLC checks to make it easy to see where you are violating KLC requirements.

For example when I run this component through KLC checks:

```
checking component: LTC2508-32
  Violating Rule 3.1
    pin: RDLA (1), posx 650, posy 400
    pin: RDLB (2), posx 650, posy 300
    pin: VDD (3), posx -250, posy 850
    pin: GND (4), posx -150, posy -850
    pin: IN+ (5), posx -650, posy 400
    pin: IN- (6), posx -650, posy 300
    pin: GND (7), posx -50, posy -850
    pin: REF (8), posx 0, posy 850
    pin: REF (9), posx 100, posy 850
    pin: REF (10), posx 200, posy 850
    pin: SDOB (20), posx 650, posy -400
    pin: SEL0 (11), posx -650, posy -200
    pin: BUSY (21), posx 650, posy 100
    pin: SEL1 (12), posx -650, posy -300
    pin: OVDD (22), posx -150, posy 850
    pin: MCLK (13), posx -650, posy 100
    pin: GND (23), posx 150, posy -850
    pin: SYNC (14), posx -650, posy 0
    pin: GND (24), posx 250, posy -850
    pin: DRL (15), posx 650, posy 200
    pin: EPAD (25), posx -250, posy -850
    pin: GND (16), posx 50, posy -850
    pin: SDOA (17), posx 650, posy -200
    pin: SCKA (18), posx 650, posy -100
    pin: SCKB (19), posx 650, posy -300
  Violating Rule 3.6
    field: U at posx -400 posy 700 size 60
    field: LTC2508-32 at posx 550 posy 700 size 60
    field: Housings_DFN_QFN:DFN-24-1EP_4x7mm_Pitch0.5mm at posx 150 posy -950 size 60
    field: http://cds.linear.com/docs/en/datasheet/250832fa.pdf at posx 200 posy -1050 size 60
  Violating Rule 3.8
    documentation OK
    checking alias: LTC2508CDKD-32
    checking alias: LTC2508IDKD-32
```

Please address the KLC errors as above :)

Further, as you have multiple ALIASes of the same part, it is important that the description of each ALIAS is different, so that they can be differentiated in the library.

If I am reading the datasheet correctly, there is no such part as `LTC2508-32`, it is either `LTC2508CDKD-32` or `LTC2508IDKD-32` where the only difference is temperature range?

If this is the case, please remove the `LTC2508-32` part and add the temperature range information for the two other parts.

Cheers

milosz88:
Hi @SchrodingersGat 

1. Violations are fixed now (sorry for such a basic mistake...)
2. Aliases are fixed but...

It looks pretty unusual for me to add so detailed information like temperature range in a main description. I searched for similiar cases in KiCad library and found nothing. What is general idea to deal with aliases in these cases? Maybe additional field in parameter list?

Thanks,
Milosz

SchrodingersGat:
Normally, this functional information should not be included in the description, you are correct.

However, the *temperature rating* is the *only* differentiating factor between the ALIAS parts - if the difference is not important, why have an ALIAS?

As an example, please have a look at the LTC4440* parts in *Power_Management.lib* - these show examples of multiple parts which are differentiated only by temperature rating.

Please also copy the temperature notation as for the LTC4440 - and ensure that within the description, items are comma-separated.

Does this make sense? I hope I have cleared things up.

Cheers

milosz88:
Yes, it fully makes sense.

SchrodingersGat:
Please run the KLC script with the `--enable-extra` flag - this will show up two more things:

1. Pin type mismatch (GND pins should be set to `POWER INPUT` type
1. Line width - Change the symbol box line width to 0.01" (10 mil)

SchrodingersGat:
Also please remove the datasheet field in the **symbol** - the datasheet should be stored in the `.dcm` file (which it is) but **not** in the .lib file. 

This is because someone else may make an ALIAS of this part which is from a different manufacturer and thus has a different datasheet. The datasheet should be associated with the part (.dcm) and not the symbol (.lib)

Cheers

SchrodingersGat:
Fantastic, thanks :)

PR #877, opened 2017-01-01
Added Trinamic TLC262

matyro:
Added a trinamic library including the TLC262 step motor driver


SchrodingersGat:
Thanks for the contribution. This part would be better located in *motor_drivers.lib* - can you please mvoe it across?

matyro:
More cleanup done 
tmc262

jkriege2:
Hi!

please remove the datasheet field and save the datasheet in the symbol properties (i.e. in DCM file) only.

Also the power-pins should be at the top (VCC,...) or bottom (GND...) of the symbol and there should be a separate pin for the bottom PAD of the package (pin-number 33).

Best,
JAN

jkriege2:
PS: We are currently clearing up the KLC ... please have a (preliminary) look here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC (and here: https://github.com/KiCad/kicad-library/issues/879 for the discussion!)

matyro:
Symbol update
5V is not realy a power input/output but a pin to put some capacitors for stability.
SG_TST is a feedback output pin.

Right side is not counter clockwise but sorted after the ic naming (A before B)
Should i move the PWR pins to counter clockwise order?

jkriege2:
looks good so far.

1. please set the styles of the outputpins from non-logic to Line (looks a bit weird to me with that cross)... or is there a reason to have them as "non-logic"?
2. Also maybe move the three power-inputs and the four GND pins close together (100mil distance) and have the 5VOut with 200mil distance as kind of "grouping"?
3. Please remove the inverted circle from pin 15 and set its text to `~ENN~`. That gives a clearer symbol (with overline over ENN to symbolize the inversion!)
2017-01-08 13_38_58-part library editor d__kicad_kicad-library-review_library_motor_drivers lib

4. Please reference the datasheet PDF directly, i.e. `http://www.trinamic.com/fileadmin/assets/Products/ICs_Documents/TMC262_datasheet.pdf` instead of a link to the webpage.

Best,
JAN

matyro:
I set the pins explicit to no logic because they can source, sink or sense up to 60V.

jkriege2:
Hhmmmm .. I don't think we use that feature anywhere else in the lib, but can also leave it like that ... would you fix the other points?

best,
JAN

matyro:
Sry took me way longer then expected

jkriege2:
please fix conflicts

matyro:
Not sure where this conflicts coming from or if the merge is correct but hopefully it is

jkriege2:
OK, two minor things to still change:
1. please set FPFilter `QFN*1EP*5x5mm*Pitch0.5mm*`
2. remove commas from keywords, just separate by spaces.

best,
JAN

jkriege2:
nice work!
JAN

Issue #879, opened 2017-01-02
KLC Addition - PWR and GND pins


SchrodingersGat:
Nominally we suggest PWR pins are located at the top of a symbol, and GND pins at the bottom. However this is not stated in the [KLC rules](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention).

I suggest that rule 3.5 be updated as follows:

###Old Rule 3.5
`Whenever possible, inputs are on the left and outputs are on the right.`

###New Rule 3.5
```
Wherever possible, pins should be arranged by function:
* Power pins should be placed at the top of the symbol
* Ground pins should be placed at the bottom of the symbol
* Inputs should be placed on the left of the symbol
* Outputs should be placed on the right of the symbol
```

CarlPoirier:
Sounds good, can't we put that in one short sentence?

`Whenever possible, power pins are placed at the top, ground at the bottom, inputs on the left and outputs on the right.

SchrodingersGat:
I would lean towards bullets for clarity, especially as many users are English as a second language. Thoughts?

CarlPoirier:
Then we better split most rules into bullets!

SchrodingersGat:
I can tackle this.

diggit:
I agree on bullets.

There are negative and positive power pins. Ground is power pin too. I'd mention control/controlled pins too.
Wee can add examples...

Wherever possible, pins should be arranged by function:
* **Positive power** pins should be placed at the **top** of the symbol
   (VCC, VIN, VDD,...)
* **Ground and negative power** pins should be placed at the **bottom** of the symbol
   (GND, VSS, ...)
* **Input or control** pins should be placed on the **left** of the symbol
   (OP-AMP +-, SPI, I2C, UART,...)
* **Output or controlled** pins should be placed on the **right** of the symbol
   (Vout, RS-485, Ethernet, ADC in, DAC out,...)

jkriege2:
Hi!

I only see obne (special case) problem with these rules: What happens if you try to design a 7905 or comparable neg. regulator ... is the GND pin now at the bottom or at the top?

Otherwise: That would be a great addition!

Best,
JAN

Best,
JAN

SchrodingersGat:
Great suggestions @diggit 

I have mocked up some changes here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC

Let me know what you think - @CarlPoirier if you're happy with this I'll transfer it to the KiCad page

[...]

PR #886, opened 2017-01-05
Added Microchip MCP1650

Frankkkkk:
My first commit. I checked with checklib and no warnings on this component. I hope it's OK !

Thanks


jkriege2:
Hi!

nice work ... some minor things to do:
1. pin length 100mil is enough
2. outline whould be 0.01" thick
3. I would rearrange the REFDES (top) and symbol name (bottom), i.e. mirrored ;-)
4. GND should be at bottom
5. please add the two NC pins as invisible NC-pins (run checklib with `--enable-extra`) as otherwise the pin-count filter does not work
6. move datasheet link to DCM-file (i.e. document properties) and clear datasheet field
7. I think this device fits better into dc-dc.lib ... please move it there 

... and have a look e.g. at MCP16301 for an example how your symbol could look:


Also note: We are currently discussing some clarifications of the KLC ... these should also really help with the symbol, as that is what we currently actually require ;-)
  - https://github.com/SchrodingersGat/kicad-library/wiki/KLC
  - https://github.com/KiCad/kicad-library/issues/891
  - https://github.com/KiCad/kicad-library/issues/879

Best and thanks for your contribution again!
Cheers,
JAN

Frankkkkk:
Hi,
I modified the thing (I left-out moving it into dc-dc. Will do that in a new "clean" branch once it's all ok).

There is still a EC03 warning, I don't know what it is.

What do you think ?

Thanks,
Frank

jkriege2:
Hi!

looks great to me ... only one thing would remain: Also if the footprint is fully specified, you need a FPFilter (so wre can match variants, e.g. `_HandSoldering` of the footprin. So please add the FPFilter `MSOP*3x3mm*Pitch0.65mm*`.

The EC03 is a bit cumbersome (IMHO), it tries to advice on the location and orientation of the labels. You can run the scripts with the option `--fix` to autofix several problems. Then you would get something like that:
2017-01-08 13_23_19-part library editor d__kicad_kicad-library-review_library_microchip lib

But either change it (if you like that style9 or leave it as it is right now. I would merge this now after you fixed the FPFilter, so could you move it to the dc-dc.lib? Also: Would you be willing to also contribute the other parts from the datasheet? That should be copy+minor mods ...

best,
JAN

Frankkkkk:
Here we go: on dc-dc.lib and MCP1650/51/52/53

PS: how do you know how FP filter regex to use ? Is it "manual" or is there a picker ?
PS2: how do you display the blue output types on the library editor ?

Thanks,
Frank

jkriege2:
Hi!

nice work ... I spotted one more bug: in MCP1653 you have the wrong package, it should be `Housings_SSOP:MSOP-10_3x3mm_Pitch0.5mm` and FPFilter `MSOP*3x3mm*Pitch0.5mm*`

Best,
JAN

PS: It matches all footprints that match ANY RegExp
PS2: In a current KiCAD nightly, click in the smybol editor here:
2017-01-08 16_02_56-part library editor d__kicad_kicad-library-review_library_dc-dc lib

Frankkkkk:
Whoops, forgot to update.
Here it is ;-)

jkriege2:
nice work!

Thanks,
JAN

PR #888, opened 2017-01-06
Add ADM2484 isolated RS422/RS485 transceiver

MHolzbrecher:


jkriege2:
Hi!

thanks for the contribution. we have a set of rules that parts for this lib need to follow: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention + some scripts that check them: https://github.com/KiCad/kicad-library-utils Also we are currently discussing a clarified version here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC

So about your symbol, please change the following before we can merge it:
1. add the two NC pins 7 and 10 as invisible NC-pins (otherwise the pin-count filter will not work correctly)
2. move symbol name and REFDES a bit close to the symbol outline (see e.g. ADM2582E)
3. please remove comma from keywords, they should be space-separated

Best,
JAN

MHolzbrecher:
Hi,

it's now modified.

jkriege2:
nice work!
Thanks,
JAN

PR #897, opened 2017-01-10
Add standard diodes

jkriege2:
- updated `diode.lib` to meet new diode style (empty triangle) in `device.lib`
- added several standard diodes


SchrodingersGat:
Great :)

PR #902, opened 2017-01-11
Modify capacitor 3dmodels

jkriege2:
this PR complements: https://github.com/KiCad/Capacitors_THT.pretty/pull/12

2017-01-11 18_34_38-3d viewer
2017-01-11 18_35_28-3d viewer


SchrodingersGat:
I'm happy with all the 3D models but unsure about the FP filters:

### Non-Polarised Caps
* `C_*` 

### Polarised Caps
* `CP_*`
* `C_*`
* `c_elec_*`
* `Tantalum_`

I *think* that this should cover all possibilities, at least in the libs:
* Capacitors_SMD
* Capacitors_Tantalum_SMD
* Capacitors_ThroughHole

Thoughts?

jkriege2:
working on this now ... I'll let you know, when I'm done!

SchrodingersGat:
I've merged the Tantalum PR but would like to see a small change to SMD PR. Thanks

SchrodingersGat:
Great, thanks Jan

SchrodingersGat:
@jkriege2 there is an issue here that I did not notice before:

1. Model scaling should be 1:1:1 (model size will need to be changed)
1. Model name should include height (we can ignore the KLC that states they must be the same name as the footprint)

poeschlr:
Also i noticed these models have the fcstd file added instead of a step file.
Would it be possible to add the step file aswell?
You could use the stepup tools by @easyw to create correctly scaled wrl files and a step file in the same operation. (The wrl export of these tools also creates better wrl models than the freecad internal exporter. It adds a crease angle setting and better materials)

jkriege2:
Hi!

I can do the changes ... maybe during the week, as I first have to look how the export with proper sizes (for 1:1:1-scaling works for me) ... I'm planing on pushing all my models accordingly. I will also combine that with a change to the footprints: BTW: How do we do it then? I push the 3D models to the kicad-library + the new 3D package repo and add a PR with the footprint changes (including added F.Fab-REFDES and possibly other KLC-fixes)?

Best,
JAN

easyw:
@jkriege2 as @poeschlr suggested, it is possible to use [kicad StepUp](https://sourceforge.net/projects/kicadstepup/) to export STEP and export and scale wrl from FC file...
The best to have a single object for the STEP file is to fuse (Part Union) the components of the Capacitors and export the fused object through StepUp button... After exporting the models, check if the STEP file is fine and exported as single object
Because of a small issue in FC recent releases I would suggest you to use FC 0.15 to make the export...
Feel free to contact me for further detail you may require...

jkriege2:
doing that by hand yould be quite tedious ... is there a python function that I can call to do the export? Or any other means of automatisation?

jkriege2:
also I started a little test:
https://github.com/KiCad/Valves.pretty/pull/3

Please heave a look and tell me whether the STEP-models are fine!

poeschlr:
> doing that by hand yould be quite tedious ... is there a python function that I can call to do the export?

you could look at the 3d model generator scripts. These use the same export functions as the stepup tools. maybe you can use them as a starting point.
As an example look at the export_one_part function of the export_conn_jst_eh scripts in the jst subdir. (the part of interest starts at about line 196) These scripts use the freecad internal step export and the vrml export scripts in the exportVRML subfolder.

easyw:
it is important to have a good exported model **to make an union of parts before exporting**...
Moreover, atm, FC015 is still the best option to export the model without having surplus products in step models
https://forum.freecadweb.org/viewtopic.php?t=20455
step-export-comparison

jkriege2:
@easyw Thanks for the advice ... Could you have a second look? I exorted now from FreeCAD 0.15!

easyw:
it is fine now! It is a single object in FC 0.15, 0.16 and 0.17 :smiley:
Only one thing... before the exporting, it is better to rename the 'Fusion' object to the name of the model to export as in this case
'VALVE-MINI_PENTODE_LINEAR'
In this way when loading the STEP file, its name will be the name of the file itself and not a generic Fusion label.

keep in touch 😄 
Maurice

PR #905, opened 2017-01-12
Rework diode 3dmodels

jkriege2:
this PR renames all diodes so they have a prefix `D_` ... this way they conform to the standard set by e.g. the Resitors, Capacitors, ...
also the FPFilters are adapted accordingly.

Note that this PR has to be merged together with these, so the footprints get renamed too:
- https://github.com/KiCad/Diodes_SMD.pretty/pull/26
- https://github.com/KiCad/Diodes_THT.pretty/pull/16


SchrodingersGat:
Thanks Jan

Issue #892, opened 2017-01-07
make 74xx library KLC compatible and extend with more devices


koder68:
The 74xx.lib library today is far from KLC compliant. It should be updated to follow the rules. Also there are a lot of devices that are not included. For some I have sen there are issues entered but not for those I needed.

I could volunteer to do at least some of this work but as far as I understand some of the changes needed will break existing schematic diagrams and I kind of seek approval to go ahead anyway.

- Changing pin lenghts.
- Changing origo
- Changing description to match these from wikipedia (https://en.wikipedia.org/wiki/List_of_7400_series_integrated_circuits)
- Moving label and reference to some common location (but which)
- Using keywords from kind of a controlled vocabulary
- Making the 74xx part root for all components and HC, HCT, L, S, LS, ALS and F variants just aliases.

It would be a lot of work and I guess some use the ttl_ieee library which I plan not to touch.

One question would be what to do with the 1-gate devices. There is a separate 74xgxx library. Perhaps those in the 74xx library should be moved there.

koder68:
After some checking i'm not overly enthusiastic about this .1 inch standard pin length. For pins with inverted input its seldom enough so need to go to .15 and that is not enough for pins with 2 digit numbers. .2 inch works but why have .1 inch on one side and .2 on the other. And why have .1 inch on one device and .2 on another.
All in all, I would like to make every pin for black box components .2 inch long. Besides, the checker does not seem to check pin length...

jkriege2:
Hi!

I think for the 74xx-libs, a lot more work has to be done (maybe a complete rewrite would be needed), because there is also the topic of the hidden power-pins, which many think should be removed, see: https://github.com/KiCad/kicad-library/issues/605

Best,
JAN

koder68:
I had seen that issue but obviously not read it in detail. I have now and will think about this for a while and then add to that discussion.

bobc:
Just an FYI, I am working on an updated 74xx.lib. I would  like to address several of the issues mentioned here.

Many of the logic devices no longer have a 'standard' part available, e.g. 7400,7402 and are only available in HC, HCT etc. However, I think it will be a lot tidier if the base part is always 74xx and a lot easier to search.

I have scraped the data from the wiki page into a text file with a simple script, so I can automatically add descriptions, datasheets. Suggestions for standardised keywords welcome!

SchrodingersGat:
Closing as the new logic work done by bobc has been merged now.

Issue #605, opened 2016-07-06
74xx and 74xgxx forced power pins


diggit:
All symbols from 74xx and 74xgxx libraries has power pins connected to GND/VCC and set invisible. This is OK in most cases, but I designed several devices with isolated parts and it was problem.
When there are 74\* chips on both sides of isolation barrier, you have to modify symbols to stay isolated.

Single part symbols are easy to modify (CNT, REG, MUX,...).
Multi part symbols are problem, there is not space, where you could place power pins (AND, OR, INV, NAND,...).

Does anybody else solved this problem?

SchrodingersGat:
I'd really like to see this fixed too. The way I see it there are three options:

a) Leave it as it is, people struggle along with invisible pins.
b) Change all the parts and make the Vcc / Gnd pins visible and give them a non-zero length
c) Create a duplicate of each part with visible pins (as in b) above) so that user have "options"

What do you think we should do @diggit ? I am happy to help with this, I can see it taking some time as there are a lot of parts.

At the same time we should try to improve the general quality of these symbols.

diggit:
I'd go b) way, but do you have any idea where to place power pins for multipart symbols like AND,OR,.. ?
- Maybe we can leave power pins where they are and set visible. I am surprised, that same pin name with power symbol is enough to connect it to same called power pin (net). Or is it something else? Is there any documentation about this behavior?

Once this is solved, we can discuss what improvements to do and start working on it.

jkriege2:
Hi!

yes, that is a huge pain for me too ... and I've seen this is e.g. also done for the NE555 (even worse). I would suggest a third option though, which I've seen in Target3001!:

d) add an additional part to the component, which contains only the power pins and which can be placed somewhere else on the sheet ... 
![pwrpins](https://cloud.githubusercontent.com/assets/11967523/19070516/f8be5cb2-8a2b-11e6-9663-d9ae23b0af3f.png)

Also this way, we can maybe write a script that removes the existing power pins and adds the new part to the component with defined distances and all ...

What do you think?

Best,
JAN

jkriege2:
Eagle also does it wioth option d):
http://dangerousprototypes.com/blog/2013/02/12/eagle-tip-gateswap-and-pinswap-tool/

jkriege2:
Just to clarify the problem I made a short test:
[...]

PR #906, opened 2017-01-14
Add 74HC240 part

cydanil:
Add the 74HC240 Octal buffer/line driver; 3-state; inverting device symbol, as described in the [ attached datasheet](https://github.com/KiCad/kicad-library/files/706025/74HC_HCT240.pdf), to the 74xx library.


cydanil:
I got a better hang at making parts, and found that this drawing violates the rules 3.1, 3.6, and 3.8.
I will make the necessary amends first.

jkriege2:
Hi1

thanks for the contribution. First of all: We have a set of rules (KLC) for symbols for this lib (see here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention and note we're discussing an improved version here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC and here: https://github.com/KiCad/kicad-library/issues/879

Also there exist Python-scripts to check (and in some cases fix) these rules: https://github.com/KiCad/kicad-library-utils

Now for your PR:
1. please split this pr for the 74xx.lib and the 6N137 parts ... then it's easier to review!
2. On the 74xx.lib, please note that there are severaldiscussion ongoing ... as these libs are not even near our KLCs ;-):
     - https://github.com/KiCad/kicad-library/issues/605
     - https://github.com/KiCad/kicad-library/issues/892
3. Concerning your 74HC240 version: Please make outline 0.01 thick and "fill with background" (yellow), make pins 0.2" long, VCC should be on top-edge, GND on bottom-edge and please add description to 74HCT240 (it#s not automatically copied in ALIASes), footprint field should be hidden be empty, as this is available (with the same pinput) in different packages. the FPFilters should contain filters that match all these packages: e.g. `DIP*W7.62mm*`, `TSSOP*4.4x6.5mm*Pitch0.65mm*`, `SSOP*5.3x7.2mm*Pitch0.65mm*`, `SOIC*7.5x12.8mm*Pitch1.27mm*` ... based on your datasheet
4 Concerning the 6N135-devices: Please also add description... to the ALIASes, we now uses LED-symbols as defined by IEC, i.e. non-filled triangle, see e.g. `VO615A`: 2017-01-15 18_37_49-part library editor d__kicad_kicad-library-review_library_opto lib, also please clear the footprint field (observe the different versions in the datasheet!) and use filters based on thise in `VO615A` that also are available in comparable packages, also: Why do you use a different symbol style for VO2630/VO2631?

Best,
JAN

cydanil:
Hi Jan,
Thank you for taking the time to get back to me, and pointing me on the right track.
I'll do the amend to the pull request, as you've suggested, and I'm currently working on the issues you've highlighted.

However, I've made the pins a bit longer (increased the length by 1.27mm) as suggested in rule 3.2, as the numbering on pin 1 and 19 was messy (overlapping the inverted graphic sign). 

It now looks so:
pin_numbering_issue

Should I make them even longer, or is that acceptable?

Regarding the optos, I draw them in different ways as I wasn't sure which is acceptable, likewise for the led component within. I'll amend these.

Thanks again,
Cyril

jkriege2:
I would leave the pin lenght as is ... this isn't really nice, but is done often ... and IMHO the too-long-pins look a bit ugly ;-)

Best,
JAN

cydanil:
Okay, I believe it should now be good!
I added the footprints to the filters, and fixed the drawings, and will do a separate pull request for the optocouplers.

Thanks,
Cyril

jkriege2:
Hi!

somehow you lost the background and line-width of outline (Please make outline 0.01 thick and "fill with background" (yellow)).

Also the FPFilters should not list the exact footprint, but should be:
- `DIP*W7.62mm*`
- `SOIC*7.5x12.8mm*Pitch1.27mm*`
- `TSSOP*4.4x6.5mm*Pitch0.65mm*`
- `SSOP*4.4x6.5mm*Pitch0.65mm*`

Best,
JAN

cydanil:
Thank you for your patience.
 I've amended the filters and outlines.

Cyril

jkriege2:
Nce work1
THANKS,
JAN

PR #907, opened 2017-01-14
add Ads1243 ADC 24bits 8channels

penoud:
I check the componnent with the tool named checklib.py


jkriege2:
Hi!

nice work. I spotted a bug though (you could run our chechlib.py-scripts with `--enable-extra`):
The GND pin should be 15 and not 9!

Also the footprint field should be set to `Housings_SSOP:TSSOP-20_4.4x6.5mm_Pitch0.65mm` and the FPFilter to `TSSOP*4.4x6.5mm*Pitch0.65mm*`

Best,
JAN

PS: We're currently discussing an improved version (more explanations/examples) of our KLCs: https://github.com/SchrodingersGat/kicad-library/wiki/KLC and the discussion here: https://github.com/KiCad/kicad-library/issues/879

penoud:
I updated as your requested. Sorry for GND pin misnumbering.

jkriege2:
??? for some reason you removed the .lib-file from this PR???

In the description TSSOP-20 is enough ;-) but the full housing-path should be set in the foorprint field!

best,
JAN

jkriege2:
??? for some reason you removed the .lib-file from this PR???

In the description TSSOP-20 is enough ;-) but the full housing-path should be set in the foorprint field!

best,
JAN

penoud:
Sorry, I did a wrong manipulation with Git. Now I understand what you mean with the FP field. I hope this is better.

jkriege2:
OK, could have explained better ;-)

... one thing is left: the footprint field should be hidden ... please still change that, then I can merge!

Best, JAN

penoud:
I changed the footprint to invisible field.

jkriege2:
Nice work!

Thanks,
JAN

PR #909, opened 2017-01-14
SIM card added

Frankkkkk:
A new sim card.

Still no footprint for it: don't really know how to make "good" ones + it really depends on the socket you want to use.

Passes the tests.

Thanks ;-)


jkriege2:
Hi!

nice work, please fill it "with background"  (yellow) ... and maybe change the outline to be a bit closer to an actual SIM-cad, e.g. (Polygon with 2 points on the rounded edges ... can be filled!):
2017-01-15 17_26_35-part library editor d__kicad_kicad-library-review_library_conn lib


A question: Is the pin-out with the two NC-pins common to all sockets?

For the socket: If you have time ... maybe provide a footprint for a SIM-socket available from a shop like e.g. conrad.de (germany): https://www.conrad.de/de/wr-crd-sim-kartensockel-pole-6-wuerth-elektronik-inhalt-1-st-1088834.html , http://www.produktinfo.conrad.com/datenblaetter/1000000-1099999/001088834-da-01-en-WR_CRD_SIM_KARTENSOCKEL_693010020611.pdf

Best,
JAN

Frankkkkk:
Hi,

- I've added a footprint (KiCad/Connectors.pretty#54). Not sure if it's the right proceeding ? It doesn't seem to pick it up in CvPcb. Also, only did the 2FF SMD version. If its okay and I've got time, will try to do the 3FF SMD version too.
- As you said, pins 4 and 8 are rarely used so I took them out. Nevertheless they still are taken into account in the pinout (i.e. I can't reorder them as 1,2,3,4,5,6). It corresponds on the Connector.
- The dimensions are now nearly the same as a real sim card, however I can't fill the background of the SIM card.. How can I convert it to a polygon so it fills the inside ?

Thanks for your time,
Cheers,
Frank

jkriege2:
If they are in the footprint, you'll need them in the symbol ... What are they (sometimes) used for? Would it make sense to have two symbols:
One with the two pins hidden and one with them visible and labeled properly?

... as you see i'M not an expert on SIM-cards ... ;-)

About filling: I think it is not possible in KiCAD if you use arcs ... they can only fill their section :-( ... You will have to make the outline as a single polygon, i.e. simply add 2-3 Points on the rounded corners, then it LOOKS mostly round and you can fill it.

I'll review the connector there ... it was the right way to do ;-)
What I just saw. You should set the FPFilter to `SIM*Card*Holder*` ... then also the underscores are matched (if e.g. someone uses dashes in another SIM-card-Holder). If it doesn't show up: Is the new holder in a search path of KiCAD? Maybe restart?

best,
JAN

Frankkkkk:
After a more extensive search, the two "unsused" pins are sometimes used for chip detection (i.e. is the chip present). But they seem to be rarely used today (newer sim cards seem to only use 6).

Nevertheless, the standarized pinout for 6 pins seems to be 1,2,3,5,6,7; so I think we should keep it that way. Of course, the connector only offers these 6 pins. Maybe I'll make an 8 pin version if I use it in the future ;-).

I've changed a bit the look. I think that it looks better now:

2017-01-16-140205_436x318_scrot

Thanks,
Cheers

jkriege2:
Nice Work!

JAN

PR #912, opened 2017-01-15
add TCA9544A i2c 4 channels hub

penoud:


jkriege2:
Hi!

nice work!

Could you change three minor things:
1. Since this symbol has only one footprint, please set the (invisible!) footprint field to `Housings_SSOP:TSSOP-20_4.4x6.5mm_Pitch0.65mm`
2. the FPFilter should be `TSSOP*4.4x6.5mm*Pitch0.65mm*`
3. Could you please make a clean PR without all the clutter of commits that are not connected to the TCA9544A?
Thanks,
JAN

jkriege2:
did you do something to the line-endings (I think Git does that under some circumstances), as they are all changed for me now ... please revert that change!

Best,
JAN

jkriege2:
PS: The line-endings should be `0x0D 0x0A` i.e. CR+LF

penoud:
Did the last change remove the problem?

jkriege2:
yes it did ... two more changes please:
1. hide the footprint field
2. could you remove the "Input low" style? I think Input+iverterd-text is enough ;-)

JAN

jkriege2:
please fix problems by merging in the global master!
JAN

penoud:
I hide the footprint, change pin style for int.

jkriege2:
Symbol looks good now .... but please limit this PR to the files `interface.lib/dcm` before I can merge!

best,
JAN

PR #913, opened 2017-01-15
add TCA9534 I2C SMBus Gpio expander 8 channels to interface

penoud:
checked with the python script


jkriege2:
Also here: nice work!

Please change these things:

1. clean up PR (remove unrelated commits)
2. maybe INT-pin clode to SDA/SCL?
3. remove footprint field (there are two  options, then the field should be empty)
4. replace FPFilters with these two:
  - `TSSOP*4.4x5mm*Pitch0.65mm*`
  - `SOIC*7.5x12.8mm*Pitch1.27mm*`

Best,
JAN

PR #914, opened 2017-01-15
update TCA9534 comments according recommandations of PR #913

penoud:

PR #915, opened 2017-01-15
add TCA9534 to interface

penoud:
According last comment of PR #914


jkriege2:
nice and clean. Please only set the pinstyle of INT to LINE ... and what do you think of this pin-layout:
2017-01-16 10_29_43-part library editor d__kicad_kicad-library-review_library_interface lib

then it is easy to have the INT-line together with the SDA/SCL and it would e.g. match the layout of your I2C 4-port-Hub! 


Best,
JAN

jkriege2:
INT is alright now, but you have again mangled up the line-endings ... maybe check how you git interferes with them, as I think git has a setting that might correct them when pulling or committing to a preset mode!

JAN

PR #916, opened 2017-01-15
Changed pin lengths for all diodes in diode.lib to 100 (down from 150)

SchrodingersGat:
There were some diodes in diodes.lib with longer pins (150 mils). This PR reduces those pins to 100mils


jkriege2:
nice work!

JAN

PR #919, opened 2017-01-16
Add MIC5219-3.3BM5

svheymann:


jkriege2:
Hi!

I think you lost the ADC128D818 when merging in the master!

Best,
JAN

svheymann:
Still struggling a little with that windows github interface (and also with KiCAD) :) I had used subversion and eagle until now so i have to get used to it... I hope this works now.
Concerning the ADC: For me it looks like it is still in the pull request?
I like to try out this workflow first, then i will add more components.

svheymann:
Ah, i see my mistake. I thought the pull request would be closed after sending it. I did not know that i can add more commits by committing to the master branch while the pull request is open. Next time i will create another branch if i want to add 2 different components at a time?

jkriege2:
Yes1

the way to go is to make a branch for each new component/group of components (Git is really all about branching ... much more than in SVN). Then you send a PR for this branch and edit in it, until it is merged. After we merge it, you can actually even delete you branch, as all commits will then be put into the master branch. The problem in this scheme is that you might have a branch that is behind the master, which can lead to merge conflicts. In that case you have to merge the master into that branch, so they are compatible again ... or rebase the branch.

Best,
JAN

PS: I'll have a look at your fixes later ,-)

jkriege2:
about the actual symbol: VERY NICE!

Please add the description/keywords/datasheet to each alias (they are not inherited automatically).

Best,
JAN

svheymann:
done.

jkriege2:
Nice Work!

Thanks,
JAN

PR #928, opened 2017-01-17
add TCA9534 to interface

penoud:
…e.autocrlf input' to resolve CR+LF trouble


SchrodingersGat:
You can keep adding commits to a single PR - you don't need to keep closing and opening a new one when you need to make a change :)

penoud:
I know but I'm not a git expert and my learning curve is slow. Then sometime when I messed up, I delete a branch and create a new one for trying to keep some clarity...

jkriege2:
nice work!

Thanks,
JAN

PR #517, opened 2016-05-03
Add 3d model for Grove connector (footprint currently in PR process)

penoud:
for more informations : https://github.com/KiCad/Connect.pretty/pull/38

PR #935, opened 2017-01-18
add Grove_1x04 3d package, based on PDF https://statics3.seeedstudio.…

penoud:
reopen previous PR  #517


jkriege2:
please change the directory to Connectors.3dshapes ... as noted already elsewhere ;-)

With which software did you build thisß If you used freeCAD, cold you also provide the raw file?

Best,
JAN

jkriege2:
aaahhh ... just saw you used OpenScad ... could you also add the script you used to this PR?

Best,
JAN

penoud:
Ok, updated.

jkriege2:
Thanks!

SchrodingersGat:
@jkriege2 I think that in the future we should be linking to generator scripts in the README.md file, rather than including the scripts themselves - the lib should really only be for the .kicad_mod scripts.

The *Generator_Scripts* repo (not yet created) is where these scripts should go.

PR #564, opened 2016-05-31
Fixed 7805 & 7905 to use pin number and IN/OUT instead of VI/VO labels

stanleyseow:

PR #316, opened 2015-10-07
Added Micrel MIC5205-*M5 regulator family

jhol:

PR #948, opened 2017-01-21
fixed several issues in regul.lib and references.lib (and moved some regulators to dc-dc.lib)

jkriege2:
- missing/incomplete docu
- corrected footprint-links and FPFilters
- bit of cleanup
- pin-lengths etz. not yet done


SchrodingersGat:
Hi Jan it is hard to see what you have changed here (because of the darn CRLF line endings, again).

Can you provide a list of what has been changed here, so I can check?

jkriege2:
Hi!

changes that I did:
- several fixes of footprints (using newer names, completed names with package-sizes, adding names, adding/fixing FPFilters ...)
- several KLC-fixed (e.g. text size, 100mil-grid, ...)
- added keywords to several refernces (where missing)
- moved from regul to dc-dc: LT3472, LTC3406B-1.2, MIC4690, PT6100
- moved from regul to reference: MAX872
- fixed many symbols in regul (KLC, text-sizes, footprints, NC-pins, ...)
- set datasheets (some were local links) in regul.lib
- removed unusable old symbols, e.g. 7805,7805MS,78L05,7905,79L05, which didn't even have numbered pins!
- added LT1117-XYZ variants to regul.lib
- added MAX603EPA, MAX603ESA, MAX604EPA (+variants)
- added series of µA78xyKC/µA78xyKT of regulators

Using a local diffs shows nicely what I changed ... unfortunately the DIFF of GitHUB is not very good :-(

Does this help?

best,
JAN

jkriege2:
More fixes of KLC errors:
- moved pins on grid
- changed symbols a bit to conform to new pin placement
- ensured that no two driving pins (OUT, POWOUT) are stacked => one pin has correct type, other pins are PASSIVE
- label positions
- label text size

Specific changes below!

## dc-dc.lib
- PT6100: stacked Vout+PAS+PAS-pins should be OK (prevents ERC errors as far as I understand that)

## refernces.lib
- removed `/NOPB` from symbol names
- VIN pin-type changed to POWER_INPUT

## regul.lib
- several symbols have that NO DESCRIPTION FOUND problem, but the docu is actually there ... @SchrodingersGat can you have a look into this?
- removed `_MountingTab` variants, as those do not really make sense ...
- made 3-pin regulator symbols smaller and unified sizes


NOTE; This effort is ongoing ... so please don't merge yet ;-)

JAN

jkriege2:
Additional changes:

## refernces.lib
- Vout is now always POWER_OUTPUT
- removed symbol names from descriptions

jkriege2:
## regul.lib
- unified symbols (i.e. using ALIASes where possible at some places
- made 3-terminal symbol look alike (new&smaller)
- added package name to description

jkriege2:
@SchrodingersGat @poeschlr @pointhi 
I#m still working on this and it seems to become a large overhaul of the complete regul.lib-library ... If you have time, could you have a look at the current status (most of the devices are done, mainly some LT- and all MAX-device still need attendance ...

What I would like to know:
- How do you like the new style with comparable symbol sizes and a bt smaller than before?
- I tried to use the new `_PACKAGETYPE`-nomenclature, where it makes sense and sorted out a lot of doubled and tripled devices. Also I summarized some where that is possible into ALIASes of other symbols. Does that work for you?
- Any other obvious problems that you see / ideas that come to your minds?

Best,
JAN

jkriege2:
PS: Any idea how we deal with all the KLC-issues, like GND on top for 79XY, or power-pins on the left instead of the top? I think those make sense for the regul.lib. Should we just overlook them, or add exceptions to the rules for regul.lib (or some other scheme of matching regulators, e.g. a keyword-search in description+tags)?

jkriege2:
One last thing: Any idea for a good symbol for the simple devices in refences.lib? Especially when they have a lot of NC-pins?

jkriege2:
Hi @SchrodingersGat @poeschlr @diggit !

within the last weeks I worked thorugh the `regul.lib` to move towards KLC-compliance (as good as possible). Here are the results:
2017-05-25 11_34_40-eeschema test_addlin _ d__kicad_test_addlin

Basically I rewrote the whole lib, based on the symbols that were already there. Also I added several common regulators complementing the families that were there (mostly as ALIASes). Here's a more detailed list of things that I changed:
1. all KLC-issues have been solved (symbol style, pin-placement, packages, missing datasheets/text/description ...). Two Notes on that: As these are regulators I did not enforce the powerIN-on-top rule, but stuck the the usual symbols (see image) + I used power-pad-stacks as discussed there (https://github.com/KiCad/kicad-library-utils/pull/116) to make the symbols more sleek (especially the SO8 with e..g. 4 GND-pins). SO please ignore those errors and use the extension above.
2. I reworked the symbol names to include the package, where appropriate
3. I tried to combine the symbols as far as possible, using ALIASes (i.e. same pinout+pin-names+package => ALIAS). This should make the lib more handleable
4. The best overview over the added families (e.g. AB78Xyz/AB79Xyz-variants are given in the commit-comments)
5. I moved everything from `regul.lib` that does not fit there (e.g. references to `references.lib` and switching/DC-DC-converters to `dc-dc.lib`), but didn't give to much care to these symbols. I think gthose are best cleaned up together with the full set of symbols in those libs in a second PR ... maybe I have some time for that in the coming weeks ... but this will take time. So please do not have a too close look at those libs in this PR and ignore most errors, seeing them as a mere move!
6. I added some symbols that other users also requested. These include symbols from these PRs:
   - https://github.com/KiCad/kicad-library/pull/1238
   - https://github.com/KiCad/kicad-library/pull/1169
   - issues from https://github.com/KiCad/kicad-library/pull/564 should be fixed with this PR
   - https://github.com/KiCad/kicad-library/pull/316

Best,
JAN

jkriege2:
PS: I know this is a large workover, but I think without such efforts we will never get the libs into a good shape ... so thanks already for the work you put into the review ;-)

poeschlr:
Yes this is a very large overhaul. I admire your patience.

The symbol (or alias) for LM137K has a different naming sheme than similar parts. I think it should be called LM137_TO3 (Similar to LM117_TO3)

LM317_3PinPackages vs LM337_SOT223 (The 2 pin package includes the sot223 plus a few more. Do these not exist for the 337?)

LT1129-5.0_TO220_TO263 LT1129-5_package to match the ADJ Version and the other packages of the 5V version.

Is the NCP1117DTxxG_MountingTab version not needed anymore? (I assume the idea is to have pin 2 doubled in the footprint instead of having pin 4 in the symbol.)

In the NCP1117 the DT and ST suffixes stand for the package. Should they be called NCP117-voltage_package to match the other types?

Why is TPS769XX still there? for all other types you removed the general and replaced it by aliases for each output voltage.

I will take a closer look at this tomorrow. It is to late now to really check the symbols them selfes.

jkriege2:
Hi!

Thanks for the review!


> The symbol (or alias) for LM137K has a different naming sheme than similar parts. I think it should be called LM137_TO3 (Similar to LM117_TO3)
**renamed** 

> LM317_3PinPackages vs LM337_SOT223 (The 2 pin package includes the sot223 plus a few more. Do these not exist for the 337?)
**added versions with appropriate datasheets (in the DS I looked at first, these were not available!)**

> LT1129-5.0_TO220_TO263 LT1129-5_package to match the ADJ Version and the other packages of the 5V version.
**What do you mean?**

> In the NCP1117 the DT and ST suffixes stand for the package. Should they be called NCP117-voltage_package to match the other types?
**fixed this family (separated out ADJ + renamed + fixed TO263->TO252 ... one should really read the DS ;-)  )**

> Why is TPS769XX still there? for all other types you removed the general and replaced it by aliases for each output voltage.
**fixed this series ... all aliases are there now, general version removed, fixed DS-link**

> I will take a closer look at this tomorrow. It is to late now to really check the symbols them selfes.
**would be great!**

Best,
JAN

jkriege2:
THere is still an error for non-allowed character `µ` in `µA7805` ... I think this should be ignored for this PR, as the part really has that name!

JAN

poeschlr:
> LT1129-5.0_TO220_TO263 LT1129-5_package to match the ADJ Version and the other packages of the 5V version.
What do you mean?

Sorry i seem to have written nonsense here.
The 5V version for TO220_TO263 has the 5 V written as 5.0 whereas the other 5V versions don't have the  ".0"

poeschlr:
About the special character: Are we sure this will work on all supported operating systems? I would use "u" instead of "µ" just to be sure.

jkriege2:
About the µ: Both files should be encoded in UTF-8 (`regul.lib` states this explicitly), so this should work on all systems ... Do you use Linux, Windows, or MacOS? Does it work for you?

@SchrodingersGat What about you?

Best,
JAN

jkriege2:
`LT1129-5_XYZ` have been renamed

poeschlr:
About the µ:
I forgot for a moment that it is inside a file. (For some reason i thought this would be a filename.)
In this case it should really work everywhere.

evanshultz:
One note on the "µ". There are a large number of "equivalent" parts (there are actually different dies, but in 99% of circuits it doesn't matter), and this is only one vendor's way of naming this generic 5V regulator. "LM7805" would be functionally identical. Or "7805". Or "MC7805". Or "KA7805".

For KiCad, perhaps simply "7805" is the most universal nomenclature?

Further, note that even TI uses a simple "u" or "U" in [the datasheet](http://www.ti.com/lit/ds/slvs056p/slvs056p.pdf). See pages 1 and 19 for examples.

jkriege2:
renamed `µA78xy` to `uA78xy`

JAN

poeschlr:
For reviewing i started a google document. (Lists all "Symbols" such that i can take notes and now what i need to look at.) It should be editable by everyone with this link so if someone else wants to help with the review process, simply use this document to show if a symbol is ok or if there is a problem.

I think i will report here from time to time if i find a problem.

https://docs.google.com/spreadsheets/d/13XIsR3w7fAD7a5huCkMgNqtapnr18BSz_LbDSMmILIM/edit?usp=sharing

poeschlr:
Ok this will take me a while. (I have only about 25% done until now.)

jkriege2:
Hi @poeschlr I had a first quick glance at your doc (first of all: Thanks for all the work :-))) ... one comment though:
You mentioned for several symbols (like AP1117) that the ADJ and the fixed versions should be ALIASES of only one basic symbol. I explicitly split those at several places, because they the ADJ/GND-pin has different names and pin-types:
2017-06-06 08_22_57-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib
2017-06-06 08_22_43-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib

So I feel that should not be changed!

Best,
JAN

jkriege2:
About the naming with voltages: For some ICs I used the part-number given in the DS (e.g. TLV70012), for others the more general version LT1117-5.0 ... I'm not so sure whether this should be unified, at least not in those cases where the DS states the special numbers prominently and in some cases adds letter-codes behind them [if they are not only for packaging] ...

What do you think? @SchrodingersGat @pointhi Input on this?
I'm happy to change naming, if we agree on something ;-)

Best,
JAN

PS: I will look over all the other things you spotted within the next few days (might take some time, as it's quite busy at the moment ...)

poeschlr:
Hm for some reason i did not notice that the adj pin is of type input.
(This means i did not check the symbols well enough.)

poeschlr:
Checking the regul lib is done.
See google document for details.

jkriege2:
@SchrodingersGat @poeschlr @pointhi 

So this is coming forth slowly ... but I'm still stuck on one question: How should we put the output-voltage into the symbol names:

- Some manufacturer add numbers with decimal dot, e.g. `LT1117-5.0` (Linear Technologies)
- Some manufacturer add numbers without decimal dot, e.g. `AP131-18` (Diodes.com), where it's only from the description/DS clear, whether this is 18V or 1.8V
- Some manufacturers include the voltage directly into the device name, e.g. `TLV70030` for a 3.0V regulator, or `TPS71533`
- Some devices do not have a number part for the device name, e.g. `LF33`, `LF50`, ...
- The adjustable version often has simply no extension ... I called them in many placed with added `-ADJ`, e.g. `LT1117-ADJ_SOT223`.
- @poeschlr s proposal is to put the generic part of the device name (e.g. LT1117) and then add the voltage including decimal dot (+optionally at the back the packages), e.g. `LT1117-5.0_SOT223` for most devices. This also fits more or less with some manufacturers (e.g. LT) that actually call their devices this way. For other devices this scheme does not really fit, e.e. "everybody knows" what a `uA7805_TO220` does  and it would be rather strange to call it `uA78-5.0_TO220` I fell

So ... any idea of how to solve this problem? 
Personally I would use whatever name the manufacturer uses in their datasheet ... one could then even think about removing the `-ADJ`, where it is not used by the manufacturer ... Having a general scheme would be nice on the one hand, but it makes the connection to the actual part number far less obvious ... + the descriptions contain the output voltage in a standardized form, so e.g. filtering for `3.3V` will already give you all applicable regulators:
2017-06-07 08_00_08-choose component 699 items loaded


Best,
JAN

suzizecat:
In my opinion, the manufacturer part name should be used, because when I search a component, it's often after looking it up on a distributor website and i expect it to be found by its manufacturer part-number. 

A clear version of voltage may be added in component description, since it's this field purpose...

jkriege2:
- I think I fixed all things from your spreadsheet ... comments are there in dark-green!
- Generally I'm trying to always use the manufacturer part-name (-ADJ is an exception)
- added LDK130-family requested there: https://github.com/KiCad/kicad-library/pull/1297

Best,
JAN

SchrodingersGat:
@jkriege2 my recommendation is:

1. Use the manufacturer P/N as the datasheet specifies
1. Add the output voltage to the description
1. Then user can search either by P/N or voltage

For generic devices, @poeschlr suggestion is good - GenericName-<voltage>-<pkg>

Essentially what you have shown in your comment above is good.

jkriege2:
OK @SchrodingersGat THen I think the naming in this PR should be OK ... basically there are no generic devices ;-)

JAN

SchrodingersGat:
Great! @poeschlr as you have done an extensive review of this PR, please merge when you and @jkriege2 feel it is ready. Thanks for the great work Jan

jkriege2:
@poeschlr can this be merged?

Best,
JAN

poeschlr:
I rechecked the symbols. (I still have to check a few symbols. Everything where the first column in my google doc is still white to be precise. I will do this later today.)

MCP1703A-xx02E_SOT89 and MCP1703A-xx02_SOT89 are the same except for the temperature range. I think the aliases for the MCP1703A-xx02E_SOT89 versions can be removed.

LP2987 and LP2988 are still missing the 2.8, 3.2 and 3.8V aliases (all package variants)

MC78L18 and MC78L24: You wrote that you could not find them in the datasheet either. But the symbols are still here?

LP3963-xx and LP3966-xx according to the datasheet linked in the symbol is not available in to-252 but it is available in to-263 (It is also available in to-220 but this is already correctly represented in the filter.)

And a few datasheet links are still broken:
- MIC5205-xx: The datasheet link links to the microchip search page. Should it link to: http://ww1.microchip.com/downloads/en/devicedoc/mic5205.pdf
- APE8865-xx: The datsheet link is dead. Suggestion: http://www.a-power.com.tw/files/AP_ProductData/PD_File/APE8865.pdf or better http://www.tme.eu/fr/Document/ced3461ed31ea70a3c416fb648e0cde7/APE8865-3.pdf
- LM317_3PinPackage: datasheet link dead. Suggestion: http://www.ti.com/lit/ds/symlink/lm317.pdf
- LM79Lxx: datasheet dead. Suggestion: http://www.ti.com/lit/ds/symlink/lm79l.pdf (Otherwise symbol is ok. I did check it against this datasheet.)
- MC79Lxx_SO8 still have a broken datasheet link. Should be http://www.ti.com/lit/ds/symlink/lm79l.pdf

jkriege2:
Thanks,

will fix those during the weekend.

best,
JAN

jkriege2:
Hi!

- I fixed all datasheet links you named
- I found a DS for MC78L18 and MC78L24 (http://pdf.datasheetcatalog.com/datasheet/vishay/mc78l00.pdf)
- LP2988 is available in 2.8 and 3.8 so I added those LP2987 is available only on 3.0,3.3,5.0 (see list of available devices at end of DS), so I didn't add those
- LP3963-xx and LP3966-xx: fixed
- MCP1703A: extended temp. Range version removed
- added several missing MCP1703A and MCP1700 package variants

Let me know if you find more ;-)

Best,
JAN

poeschlr:
I finally found time to finish my review. We are really close to finish this i think. I only found one real problem (malformed datasheet link that causes an error message.) Everything else would be ok to merge even without fixing it.

LT1584-xx: I would add  TO?263* to the footprint filter

From the stuff you fixed from my last comment:
LM79Lxx: Space in front of Datasheet link. (I get an error message when i press the pdf symbol in my kicad 4.0.2)

MC78L18 and MC78L24 datasheet link broken (I get a short message about something being forbidden then i get a search window datasheetCatalog.com with two results.) But this might be close enough for now. (might it be possible that these are discontinued? I can't seem to find a place where i can order them.)

jkriege2:
Hi!

- according to DS, only LT1585 and LT1587 are available in TO-263/DDPAK
- additional SPACE: corrected
- MC78L18/MC78L24 I couldn't find any other DS link ... so I think that should do for now ... if they are not available any more (obsolete) only few people will use these symbols anyways.

So can this be merged :-))))))))))))))

Best,
JAN

poeschlr:
Yes looks good now.

Thanks for taking care of this lib and for your patience.

PR #951, opened 2017-01-22
Fix ADUM4160 pin "PIN" location

aurel32:
The "PIN" pin control the upstream pull-up, but is located on the
downstream side, and is relative to the downstream "GND" pins. Therefore
move it to the right side of the isolator.


jkriege2:
thanks!

PR #952, opened 2017-01-23
Add NXP lib, Add PCA9685

svheymann:


jkriege2:
Almost perfect. I only spotted two little mistakes:
1. Pin A4 should be an INPUT (not an output)
2. Pin SDA should be BIDIRECTIONAL

Best,
JAN

svheymann:
done.

jkriege2:
THanks!
JAN

PR #780, opened 2016-11-01
Improve and add transistors

jkriege2:
improved several transistors (so they match the symbols in `device.lib`) and added several missing devices:

## MOSFETs: 
* [ ] TP0610T  VP0610T 
* [ ] BS107 BS108 
* [ ] BS170 
* [ ] BS250 
* [ ] 2N7000 
* [ ] 2N7002 
* [ ] BF244 BF245 
* [ ] TP0610L VP0610L
* [ ]  IRF540N 
* [ ] IRLZ44N 
* [ ] IRF4905 
* [ ] IRF3205 
* [ ] BUZ11 
* [ ] IRLZ34N 
* [ ] IRLIZ44N 
* [ ] IRF9504N
* [ ] IRF7404 

## transistors: 
* [ ] BC337 BC338 
* [ ] BC327 BC328 
* [ ] BC807 BC808
* [ ] BC807W BC808W 
* [ ] BC817 BC818
* [ ] BC817W BC818W 
* [ ] BC546 BC547 BC548 BC549 BC550 
* [ ] BC556 BC557 BC558 BC559 BC560 
* [ ] BC847 BC848 BC849 BC850
* [ ] BC847W BC848W BC849W BC850W 
* [ ] BC857 BC 858 BC859 BC860
* [ ] BC857W BC858W BC859W BC860W 
* [ ] 2N3904 2N3905 
* [ ] 2N3906 
* [ ] MMBT3904 
* [ ] MMBT3906 
* [ ] PZT3904 
* [ ] PZT3906 
* [ ] 2N2219 
* [ ] 2N2646 2N2647 
* [ ] 2N3055 
* [ ] BC107 BC108 BC109 
* [ ] BC140 BC141 
* [ ] BCP53 
* [ ] BCP56 
* [ ] BCX56 
* [ ] BF199 
* [ ] MJ2955 
* [ ] TIP2955 
* [ ] TIP3055 
* [ ] TIP41 TIP41A TIP41B TIP41C 
* [ ] TIP42 TIP42A TIP42B TIP42C 
## Darlingtons: 
* [ ] BC517 
* [ ] BC518 
* [ ] BDW93 BDW93A BDW93B BDW93C 
* [ ] BDW94 BDW94A BDW94B BDW94C
## added library `triac_thyristor.lib` with TRIACs/Thyristors: 
* [ ] TIC106 TIC116 TIC126
* [ ] BT169B,D,G
* [ ] TIC206 TIC216 TIC226
* [ ] BT138-600/800
[...]

PR #953, opened 2017-01-24
Add triac lib, add Z01 series triac

svheymann:
I am not sure if i should name the components starting with Triac_ or not?
The SOT-223 package has 4 pins, but the triac has the same signal on pin 2 and 4. How do i build the symbol for this? My guess is that the way i did it is not correct because it looks very strange ;)


jkriege2:
Hi!

please rename your new lib to `triac_thyristor.lib` ... as I have already a PR running that adds such a lib: https://github.com/KiCad/kicad-library/pull/780

About the naming: no prefix required!

I'll have a closer look this evening,
JAN

svheymann:
done.

jkriege2:
I think for the 2-pin problem there are two solutions:

1. have one pin invisible and overlay the two. Then connecting to one will connect to both
2017-01-24 20_44_00-part library editor d__kicad_kicad-library-review_library_triac_thyristor lib

2. Try something like that (looks better than your solution, IMHO):
2017-01-24 20_45_56-part library editor d__kicad_kicad-library-review_library_triac_thyristor lib

Otherwise: Looks very good ;-) Onely one thing to change:
Please remove the device name from the description and rather add some basic properties of the device, e.g. `1A RMS, 600V Off-State Voltage, Triac, SOT-223`

Best,
JAN

svheymann:
done.

jkriege2:
Looks good. Some final changes:
1. Please add commas to the description, e.g. `4Q Triac, 1A RMS, 600V VDRM, 3mA Igt, 7mA Ih, SOT-223`
2. remove type ID from keywords ... that's matched by the symbol name already ;-)

I'll merge when that's changed!

Best,
JAN

svheymann:
done.

jkriege2:
Please also remove the Z01 from the keywords!
JAN

svheymann:
done.

jkriege2:
Thanks for the work!

JAN

PR #960, opened 2017-01-30
Add Hahn HS-400XX symbols

svheymann:
I have a problem with this lib: As soon as i save it with KiCad, a lot of changes happen like deletions of some spaces etc. I have done these changes in a text editor so i dont commit the other changes. Maybe someone has a look into this?


jkriege2:
Hi!

looks good: Just one question: Do you have one of these? Because the DS you link and the catalog (http://www.tme.eu/de/Document/fcd5cccf56bcc8ba4968d5c20c954514/Catalog%202016-E.pdf) show different pinouts:
- DS: +:7, -:9 (as I interpret the cap at the end)
- catalog: +:9, -:7

Do you knwo which one is right?
best,
JAN

jkriege2:
Oh and some minor corrections please:
1. The pins should be on the 100mil grid (in your case best move them 50mil up/down
2. outline should be 0.01" thick

Best,
JAN

svheymann:
I used the catalog for drawing the symbol. And yes, i have some of those and surprise: the catalog is wrong... I corrected the symbol.

jkriege2:
nice work! Thanks,
JAN

PR #961, opened 2017-01-30
added example images for KLC wiki page

jkriege2:


SchrodingersGat:
Jan this is a good idea. Can you change the name of the subdir `example_images` to `klc_images`?

SchrodingersGat:
Great, thanks Jan. Hopefully we can get the final touches on the new KLC up and running soon.

PR #962, opened 2017-01-31
new component on ESD-protection, memory and power lib. 

jeromebeguin:


jkriege2:
please split this PR by lib ... then it's much easier to Review!!!

Bst,
JAN

jeromebeguin:
How should I do this ? The only way I've found is to create a branch per lib. Is there a simpler way to do it ?

jkriege2:
THat's the way to go ... usually for every PR I create a new branch from a master, which is up-to-date with the kicad-master. Then it's easy to handle several PRs ... and when you split them by lib merging one usually does not create conflicts between them ;-)

PR #963, opened 2017-01-31
Add Microchip MCP1826S

jjuyon:
Hi,
I add the LDO Microchip MCP1826S in regul library

Best,
Julien


jkriege2:
Hi!

some changes please:
1. this will only work with SOT-223 as the other packages have only 3 pins as the tab is usually tied to pin2. An option might also be deriving a new footprint from `SOT-223` which has the top tap as pin2
2. make VIN a power input and Vout a power output, GND is fine
3. move pins onto a 100mil raster

best,
JAN

PS: Please note that we are currently discussing a better explained set of KLCs:
- https://github.com/SchrodingersGat/kicad-library/wiki/KLC
- https://github.com/KiCad/kicad-library/issues/879

jjuyon:
Hi,
I made changes:
1. I remove the pin 4 in symbol and as you suggest, I created a new footprint from SOT-223 with tab connected to pin 2 (https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/34).
2. Vin and Vout directions fixed
3. What do you mean by "100mil raster"? They were already 100mil long. Is that not enough?

Best,
JULIEN

jkriege2:
ad 3) The pin Ends should lie on Point on a 100mil raster, i.e. 0mil, 100mil, -200mil, ... not on 50mil or 150mil!

JAN

jjuyon:
Ok it's done

Julien

jkriege2:
please use these FPFilters:
- `SOT?223*TabPin2*`
- `TO?220*`
- `TO?263*`

The rest should be filtered by pincount filter!

JAN

jjuyon:
It's done but it seems not filtered by pincount filter!

Best
Julien

jkriege2:
Works for me ... just the SOT-variant is not yet in my lib ...
2017-02-01 19_09_13-eeschema kicad-library-review _ d__kicad_kicad-library-review-project

PR #964, opened 2017-01-31
adding LM73, TLC5973, TUSB4041I to texas lib

jeromebeguin:


jkriege2:
Hi!

1. Please move LM73/LM73-1 to `sensors.lib`
2. on LM73, reduce pin length to 100mil
3. on LM73 think about using maybe the same symbol as here 
2017-01-31 21_40_23-library browser d__kicad_kicad-library-review_library_sensors lib
4. FPFilter for LM75 should be `SOT-23*`
5. reduce pin length of TLC5973 to 100mil
6. TUSB4041: pakcage should be called VQFN
7. What are all those RSVD pins? Please give them the correct name and unhide them

best,
JAN

jkriege2:
What about the RSVD  pins on the TUSB4041?

TLC5973: please fix outline and GND-pin:
2017-02-01 18_46_00-part library editor d__kicad_kicad-library-review_library_texas lib

jeromebeguin:
For TUSB4041I, the RSVB pins are pins reserved and should not be connected. This is what indicated in the datasheet.

evanshultz:
@jeromebeguin You will still need to add the RSVB pins because they exist on the device. Let me give you two reasons:
- While TI says not to use those pins, they do have a purpose. They're probably for testing or programming an internal memory (product customization during manufacturing?). So while TI doesn't want just anybody using them, _someone_ may know what to do and could use those pins.
- Because those pins are physically on the device, they should be included so that the actual number of pins will be captured by KiCad. CvPcb uses the pin number to find appropriate footprints for a symbol, and if you leave off the 20 RSVD pins your 64-pin part will look to CvPcb like a 44-pin part and the matching footprints will not be displayed using the "pin number" filter. This is the most important reason to include these pins.

It can be hard to figure out why these steps are needed, but I hope this reason is clear. And we really value your time and effort to add new parts to KiCad. Thank you!

jeromebeguin:
@evanshultz I perfectly understand why I need to put these pins on the device. This why they are already on the schematics. But they are hidden for a light representation, but they are there. Should I really put them visible ?

jkriege2:
No, if they are as you explained, it's fine. I just thought I didn't see them in the datasheet, but saw pns with actual names. I will recheck this evening. If I was mistaken, then your Explanation is fine and we can go on towards merging. 

best,
JAN

jeromebeguin:
Yeah I know the datasheet wasn't the good one on the first released. Sorry for that.

jkriege2:
Ok!

two final remarks: 
1. Please use this DS: http://www.ti.com/lit/ds/symlink/tusb4041i.pdf as it also has a recommendet footprint
2. The package should really be `Housings_QFP:HTQFP-64_1EP_10x10mm_Pitch0.5mm_ThermalPad` (with an H infront of TQFP), see datasheet. and would you be willing to create that footprint, as shown on p. 41 on the DS above

best,
JAN

jkriege2:
Nice work!

thanks,
JAN

PR #966, opened 2017-01-31
adding 24AA02 to memory lib

jeromebeguin:


jkriege2:
please:
1. move VCC to top of symbol
2. hide NC
3. group SDA and SCL on the same side (left?)

best,
JAN

jkriege2:
nice work!
Thanks,
JAN

PR #967, opened 2017-01-31
adding IP3319CX6, TPD2EUSB30, TPD4EUSB30, TPD8F003 to ESD_protection lib

jeromebeguin:


jkriege2:
Hi!

- `IP3319CX6`:
  - all pins should be passive. 
  - maybe provide the internal schematic? E.g. looking something along this:
2017-01-31 21_58_34-part library editor d__kicad_kicad-library-review_library_esd_protection lib
  - We don't have a Gamaya lib for footprints
  - Could you provide the footprint also?
- `TPD2EUSB30`
  - add internal schematic
  - pins should all be passive
  - could you proide the footprint?
- `TPD4EUSB30`
  - along the lines above
- `TPD8F003`
  - provide one internal schematic (like e.g. here: 
2017-01-31 22_03_03-addlin png - windows-fotoanzeige
  - move GND to the bottom
  - pins except GND should be PASSIVE

Best,
JAN

jeromebeguin:
- IP3319CX6
The internal schematics is maybe a bit too complex for the size of device (7 diode and a coil filter). I have made the footprint, just don't know in which lib to put it: WLCSP-6

- TPD2EUSB30 & TPD4EUSB30
Also have the footprint available. I will push it as soon as the PR32 on TO_SOT_Package_SMD is merged.

jkriege2:
A short note on packages: If you plan to draw the small SOT 3-pin package, name could be sometihng like `DRT-3` or even `R-PDSO-N3` (that's how texas calls it) ... would you the also provide the derive 4,5,6-pin packages?

best,
JAN

jeromebeguin:
As you wish, but texas always have strange name for their packages, so I'm not sur it will make them generic, or we should maybe create another one with texas name.

jkriege2:
Mmhhh ... do you know whether any other Company also uses that component and how they call it ... or if there is even a "Standard" SOT-XYZ number? If so, use that .... otherwise I would use the strange Designator of Texas and not try to imply a Standard which onley exists in KiCAD

JAN

jeromebeguin:
Haha ok. I got it.

jkriege2:
hi!

The FPFilter `DRT*1x0.8mm*Pitch0.7mm` does not make too much sense if you plan to name the footprint `TO_SOT_Packages_SMD:DRT-3` as it will never match! With the footprint name you gave the FPFilter should be simply `DRT*`

JAN

jeromebeguin:
I had issue with some conflict so I have erased all the directory. The new PR is 1002.

PR #968, opened 2017-01-31
adding Micro_SD_Card_Det to conn lib.

jeromebeguin:
Micro SD card socket with detection pins.


jkriege2:
nice work, minor changes please:
1. make FPFilter `microSD?Card?Receptable*`
2. change REFDES to `J`
3. empty footprint-field, this is a generic symbol and might match to other receptables too. If not you should rename the symbol to something like the footprint name, as it's not generic then.

best,
JAN

jkriege2:
Please add documentation for 'DM3AT'
JAN

jkriege2:
Could you also add a name-prefix to the DM3AT Symbol? e.g. `Micro_SD_Card_Det_Hirose_DM3AT`.
JAN

jkriege2:
looks good so far. Please fix conflicts and I'll have a final look!
JAN

jeromebeguin:
issue with conflict. I've resolved it by erasing all the directory. The new PR is 1001.

PR #969, opened 2017-01-31
adding BAT54A to diode lib

jeromebeguin:


jkriege2:
Please make the symbol look like the generic symbol from `device.lib`:
2017-01-31 21_27_41-part library editor d__kicad_kicad-library-review_library_device lib

jeromebeguin:
I've made the De Morgan symbol already isn't that enough ?

jkriege2:
We usually don't use the De Morgan ... only for logic symbols and there for the DeMorgan equivalent!

Please don't use deMorgan here and Change the Symbol as requested. THanks!
JAN

jkriege2:
OK!

one last change:
Please add the line that goes (horizontally) through the diodes' triangle (as in the image above)

JAN

jkriege2:
thanks,
JAN

PR #970, opened 2017-01-31
adding LSM9DS1 to sensors lib.

jeromebeguin:
I plan to creat the LGA-24L footprint. But don't know in which library I
should put it.


jkriege2:
Hi1

1. maybe incorporate `accelerometer gyroscope magnetometer` into the description also?
2. maybe 100mil pin length is enough?
3. CAP/C1 should be passive
4. please go ahead and submit the package and also add the full name here + a proper FPFilter.

best,
JAN

jkriege2:
- LM73: please add `, SOT23-6` to the description
- LSM9DS1 : Please fix footprint/FPFilter as noted above1

Thanks,
JAN

jeromebeguin:
Could you tell me in which footprint library should I put the LGA-24 footprint ?

jkriege2:
@SchrodingersGat I think we would need three more .pretty-repos:
- `Housings_LGA.pretty` for land-grid array packages (see above) https://en.wikipedia.org/wiki/Land_grid_array
- `Housing_PGA.pretty` for pin-grid arrays https://en.wikipedia.org/wiki/Pin_grid_array
- `Housing_BGA.pretty` for ball-grid arrays https://en.wikipedia.org/wiki/Ball_grid_array

Could you create them? If I find some time I can move+rework some of the footprints from `SMD_Packages.pretty` that should be put into one of these.

@jeromebeguin When @SchrodingersGat created the LG-rpo, submit there and remind me of this PR, when we discussed the new footprint there!

Best,
JAN

PS: Please add a comma `,` in front of the SOT-23-6 in the description of LM73. Thanks

jeromebeguin:
I have adding the LGA footprint.

SchrodingersGat:
@jeromebeguin the footprint should be submitted to the https://github.com/KiCad/Housings_LGA.pretty library.

jeromebeguin:
It's the case.

SchrodingersGat:
Ah, I see you already have :)

SchrodingersGat:
What is the difference between the LM73 and the LM73-1? There are no differences stated in the descriptions for the two symbols.

jeromebeguin:
the I2C address

SchrodingersGat:
Great, this looks OK to me :)

jeromebeguin:
Great. thank you.

PR #972, opened 2017-02-02
Update fp-lib-table.for-pretty

t3hcatpaw:
add missing '_SMD' and '_THT' to inductor lib names to fix crash when launching Cvpcb (win/nightly r7567)


SchrodingersGat:
Argh, well that's embarrassing. Thanks @t3hcatpaw

PR #978, opened 2017-02-06
Added new pretty libs to lib-table

SchrodingersGat:
This PR adds the new Housings_xxx pretty libs to the lib-table templates

Also cleans up descriptions for other Housings_xxx libs


jkriege2:
Hi!

is SIP 
1. "Single-inline Package" 
2. or: "System In Package"

Some footprints in that lib match 1-only (classic SIP9-package) some 2 (STK672?) and some 1+2 ... and some are SIL but with zig-zag-pin-ordering (SLA704XM) ...

What should we do?

Best,
JAN

SchrodingersGat:
In context, it probably is Single-Inline Package, you're right.

SchrodingersGat:
I have updated to match. Thanks for pointing this out!

SchrodingersGat:
Thanks Jan

PR #980, opened 2017-02-06
modules.lib add Nucleo64

El3ctr0nic:
Add Nucleo64 FR411RE

footprint almost done :)


jkriege2:
Hi, sorry that this PR was lying around for some time ;-) ... I will review it shortly. From a first look: Will you also submit the footprint (to https://github.com/KiCad/Modules.pretty )? If so, we should clear up the pin numbering scheme therein (ST uses two independently numbered 2x37 pinheaders, but two pin 1s.... in a single footprint does not work in KiCAD)! As a name you proposed `MODULES:ST_MORPHO_CONNECTOR_64` in this PR: Please don't use all uppercase-letter. Also the .pretty is called `Modules` with only M uppercase!

Also please link a datasheet that also contains the pinout, e.g. http://www.st.com/content/ccc/resource/technical/document/user_manual/98/2e/fa/4b/e0/82/43/b7/DM00105823.pdf/files/DM00105823.pdf/jcr:content/translations/en.DM00105823.pdf

Best, more review soon,
JAN

jkriege2:
PS: When you make the footprint, could you also provide a variant `_MountedFromLS`, which is on the bottom of the board, i.e. the nucleo is mounted there and the board is kind of a shield on top of the nucleo. I think that might be handy, what do you think?

El3ctr0nic:
Yep ST uses diffrent numeration. it will be hard to get perfect match. I use this lib for fast prottyping, draw, it may need need some tweaks. All upper case is my miss match ;/

Pin numbering is only my decide. Maybe someone have better idea?

I use full pin numberig for PIM module. Arduino pin compatybile is limitation for this module.

jkriege2:
OK,
- so generally the symbol looks fine to me. 
- The only thing I found is that there are GND-pins 77-80, but the morpho connector only has 2*38=76 pins ... please remove them.
- The pin-numbering seems fine to me (would have proposed that way, but only looked at it now ;-) ).
- please set the datasheet-link as noted above
- Finally: please modify the footprint field and the FPFilter and we can merge this ... then we can work on a footprint. the name seems OK for me!

JAN

El3ctr0nic:
Pins 77-80 are ST-LINK GND.

Remove them? They are internal part of NUCLEO PCB

jkriege2:
Are Pins 77-80 on the footprint? I thought there are only the two 38-pin connectors?

El3ctr0nic:
They are not part of morpho but they are part of nucleo board :)
st_morpho

jkriege2:
OK, then levae them ... I'll review later today or tomorrow! 

NOte if these pins are part of the board, the footprint should have a board-specific name, as in `MODULES:ST_MORPHO_CONNECTOR_64` I would only expect the NUCLEO-pins.

JAN

jkriege2:
Hi!

I googled a bit more on this. The GND pins belong to the Programmer-attachment board STLink, correct? So they will also only be there, if the programmer is attached to the board, but as it looks that can be broken of. Correct? Maybe we should have that in the symbol and footprint-names somehow.

maybe `Modules:ST_Morpho_Connector_64_With_STLink` and `Modules:ST_Morpho_Connector_64`?

best,
JAN

El3ctr0nic:
Hi,
Modules:ST_Morpho_Connector_64_With_STLink look the best for me :)
I have drawed Nucleo 144, connector name is ZIO so name will be universal for this case too

Regards
Marcin

El3ctr0nic:
Symbol fixed :). Pls check.

Next step is footprint. I dont know what to do with LS mounting ;/
We really need it? It's more of complex than arduino. Top hat is waste potential of this board.  Personally i never used top hat.
I don't know - thinking of loud. Pls feedback :)

jkriege2:
Hi!

sorry took me a bit to come back to this ... 
1. Could you please change the FPFilter to `ST?Morpho?Connector?64?With?STLink*` This way it's more "resistent" to the separating character used.


Otherwise it's fine and I will merge!
JAN

PS: Very nice work 👍 :-)

El3ctr0nic:
Should be ok now :)

jkriege2:
Nice Work!

Thanks,
JAN

PR #931, opened 2017-01-18
Global fixes for field text

SchrodingersGat:
* Footprint and datasheet fields set to `Invisible`
* Changed all field size to `50mils`
* Changed any RefDes labeled `IC?` to `U?`

(This PR is a clean-up of https://github.com/KiCad/kicad-library/pull/917)

PR #982, opened 2017-02-07
Fixed reference type from IC -> U (multiple libs)

SchrodingersGat:
https://github.com/KiCad/kicad-library/pull/931 changed the value of REF field from "IC" to "U" for many components but the DEF line was not updated.

This PR fixes this oversight.


jkriege2:
thanks,
JAN

Issue #984, opened 2017-02-08
Case sensitivity - Windows


MrMartinZockt:
There are two libraries called NXP. One in uppercase and one in lowercase.
But Windows has a case insensitive filesystem. So it is not possible to clone the library properly. One file is always missing.

SchrodingersGat:
Thanks for pointing this out @MrMartinZockt . I have deleted NXP.lib/.dcm (upper case). I will merge the components that were in the upper-case lib into the lower-case one.

SchrodingersGat:
Fix is here - https://github.com/KiCad/kicad-library/pull/985

jkriege2:
Done!
Thanks for pointing this out!

MrMartinZockt:
I found another one

[modules/packages3d/unused_3d.3dshapes/device.douille_4mm(black).wrl](https://github.com/MrMartinZockt/kicad-library/blob/master/modules/packages3d/unused_3d.3dshapes/device.douille_4mm(black).wrl)
[modules/packages3d/unused_3d.3dshapes/device.douille_4mm(black).wings](https://github.com/MrMartinZockt/kicad-library/blob/master/modules/packages3d/unused_3d.3dshapes/device.douille_4mm(black).wings)

and 

[modules/packages3d/unused_3d.3dshapes/Device.douille_4mm(black).wrl](https://github.com/MrMartinZockt/kicad-library/blob/master/modules/packages3d/unused_3d.3dshapes/Device.douille_4mm(black).wrl)
[modules/packages3d/unused_3d.3dshapes/Device.douille_4mm(black).wings](https://github.com/MrMartinZockt/kicad-library/blob/master/modules/packages3d/unused_3d.3dshapes/Device.douille_4mm(black).wings)

SchrodingersGat:
Good find! These have been fixed also.

MrMartinZockt:
Thanks!
Everything is working fine now. :+1:

PR #985, opened 2017-02-08
Merged components from old NXP.lib (uppercase) into nxp.lib (lowercase)

SchrodingersGat:
This PR merges the NXP.lib and nxp.lib - there were two libraries that were causing issues as reported: https://github.com/KiCad/kicad-library/issues/984


jkriege2:
Thanks!
JAN

PR #987, opened 2017-02-09
Silabs CP2102N 24 pin QFN version

bradfa:
The first commit switches the silabs.lib and silabs.dcm files back to Unix line endings as Unix line endings are automatically used with KiCad version 4.0.5 on Debian that I'm using and because the other .lib and .dcm files I checked all used Unix line endings.  There are no part changes, this is just a "whitespace" change.


SchrodingersGat:
You'll need to fix conflicts now that the line-ending patch has been made.

bradfa:
Closing in favor of #989.  Thanks!

PR #992, opened 2017-02-11
Make the 'pin header' connectors easier to find for (new) users

jwatte:
Some users expect to find the CONN_ symbols through the name "header" or "pin header."
Make sure to remove this snag from the on-boarding path to KiCad by adding that term to the description of the CONN_ components.
For example, Open Parts Library, SparkFun, Seeed, Jameco, and EAGLE all use variants of this name.


jkriege2:
thanks,
JAN

PR #993, opened 2017-02-11
Add footprint filters for IDC headers.

Schlumpf:
There are IDC header footprints in library already. This PR adds the filters for these footprints to pin header symbols.


jkriege2:
thanks for spotting!
JAN

PR #994, opened 2017-02-11
Add IRM-03-XS Symbols

svheymann:
The rest of the changes happend as i simply saved the file in kicad.


jkriege2:
Hi!

nice work, the only thing missing are all the NC-pins that are in the footprint, but not in the symbol. Please add them as hidden NC-pins 5,10,11,12,13,15,17,22,24,26

Best,
JAN

svheymann:
Is it ok to put all NC pins in one place or should i better make the symbol larger?

jkriege2:
I think its better to make them a larger symbol otherwise they might automatically get connected (which we don't really want)

JAN

svheymann:
done.

jkriege2:
Hmmm I would order the pins as follows:
2017-02-12 12_11_21-part library editor d__kicad_kicad-library-review_library_ac-dc lib

Otherwise it looks a bit strange in the schematic, when the NC-pins are invisible:
2017-02-12 12_13_01-eeschema kicad-library-review _ d__kicad_kicad-library-review-project
2017-02-12 12_14_07-eeschema kicad-library-review _ d__kicad_kicad-library-review-project

svheymann:
I tried some variants before but yours is definitly the best.

jkriege2:
nice work!
JAN

PR #995, opened 2017-02-11
Add DS2482-100

MHolzbrecher:


jkriege2:
very nice work!
Thanks,
JAN

PR #996, opened 2017-02-12
Add Digi XBee SMT module

svheymann:


SchrodingersGat:
This is great work. I have only one request:

Please change default field position like so:

capture

svheymann:
done. Is there some style rule where to place that fields? Sometimes the partname field is below the symbol, sometimes it is on the upper side but centered and sometimes in the upper left corner like this.

SchrodingersGat:
This is one area in which we need to standardize, there's not really a one-size-fits-all solution to this issue yet.

SchrodingersGat:
Thanks for your contribution.

PR #997, opened 2017-02-12
Add Maxim DS3231MZ RTC chip

aurel32:


jkriege2:
Hi!

nice work. some minor changes please:
1. FPFilter should be `SOIC*3.9x4.9mm*Pitch1.27mm*`
2. reduce pin-length to 100mil (enough for 1-digit pincounts

Otherwise looks good!
JAN

aurel32:
Thanks for the review. I have just done the changes you suggested.

jkriege2:
Hi1

nice. One small thing left to change: Please remove the datasheet from the datasheet field (that should be empty) F3 in teh .lib-file. the datasheet is already stored in the component properties (.cmp-file).

Then I will merge!

JAN

aurel32:
Thanks, I have just removed the link from the F3 field.

jkriege2:
thanks for all the work!
JAN

PR #1003, opened 2017-02-14
Travis ci integration

Marsfan:
This pull uses Travis-CI (travis-ci.org), a free continuous integration service, to run checklib.py from the kicad-library-utils repo. It pulls that file straight from the repo, so it will always remain up to date.

This is an improvement on #827. That had a lot of crappy commits I was using to test the code, so I created a new branch and pull request.

[Here ](https://travis-ci.org/Marsfan/kicad-library/builds/201556336)is an example of the code after it ran.

This will require someone with access to the KiCad github account to set up. You will need to go to travis-ci.org and authenticate with github, then go to profile settings and flip the switch for this kicad-library repo. After that, it should be automatic.


SchrodingersGat:
This looks great! A few questions:

1. Is this only running on components that are changed?
1. Is there a way to put the output (error msg) in the PR conversation rather than requiring user to click on a link to Travis CI page?

Marsfan:
@SchrodingersGat From what I can tell, it only runs on changed components. I do not know how to make it put the message into the PR. As far as I know, it is not possible. But it will display a red X on the PR page, and clicking on the X will link to the Travis Build page. 

Also, the commit says I deleted Device.douille_4mm(black).wrl  and Device.douille_4mm(black).wings. Do I need to edit the code to bring those back.

SchrodingersGat:
That file is an issue with case-sensitivity between windows and linux.

Can you try and clean up this commit so that the file is not touched?

Marsfan:
@SchrodingersGat Fixed it. Travis is running a test on my repo, so we should know withing then next 10 minutes if I broke anything.

Marsfan:
@SchrodingersGat something broke, Give me a bit of time to fix it.

Marsfan:
@SchrodingersGat And I forgot to bring back douille in the process, well here it is. Just remember to engage the Travis website when this is merged in.

SchrodingersGat:
Hi @Marsfan I've finally had some time to look at this.

If my understanding of travis is correct, this will do `git clone` once, but what happens the next time the script is called, and the comparison library in `/home/travis/build/kicad-library-old/` is out of date?

Should we add a `before_script` routine that ensures that the `kicad-library-old` is up to date with respect to kicad/master ?

Marsfan:
@SchrodingersGat As far as I know, Travis automatically deletes all of the data between script runs, to keep their servers clean. We should not have to worry about that.

SchrodingersGat:
oh ok.

Currently kicad-library repository is 1.26GB! If this is happening on every PR I think we may well run into bandwidth issues. Not to mention it will mean the tests take quite some time.

Is there any way that we can perform this test without re-downloading over a gigabyte of data each time?

Marsfan:
@SchrodingersGat. Nothing happens on the developer computer, it is all between the the github servers and the travis-ci servers. In all of my tests, the system (including cloning the repos) took about 1-2 minutes.

Marsfan:
@SchrodingersGat Yay! Just remember to get someone with admin right to enable travis.

SchrodingersGat:
https://github.com/KiCad/kicad-library/pull/1050

currently testing this :)

SchrodingersGat:
hmm ok so the output is as desired but kicad-lib-utils is borked

SchrodingersGat:
@Marsfan I have added some extra tricks to comparelibs.py

Essentially the entire lib now has a checksum applied. If the new checksum matches the old checksum, the entire library is skipped.

SchrodingersGat:
I have started another test at https://github.com/KiCad/kicad-library/pull/1051 also.

The travis tests are hanging (not started). I will leave it for a bit and see what happens. Perhaps Mr. Travis is just tired.

SchrodingersGat:
woo!

https://travis-ci.org/KiCad/kicad-library/builds/204859174
https://travis-ci.org/KiCad/kicad-library/builds/204854414

Thanks for the great work @Marsfan 

Once the 3D models get moved out of this repository it will be a lot quicker.

PR #1007, opened 2017-02-15
16F1459 DIP/SOIC Pin correction

rpfand:
Corrected Pin assignment for the DIP and SOIC variants of PIC16F1459,
RB7/TX.


jkriege2:
Thanks!
JAN

PR #1009, opened 2017-02-15
Fix pin name typo, duplicate Z0.

jophut:
Conflict: pin 3 and 5 have the same pin-name.
pin 5 is correct.
pin 3 is wrong, should be Z1
Source: See page 3, Figure 1 in http://assets.nexperia.com/documents/data-sheet/74HC_HCT4053.pdf


SchrodingersGat:
Thanks!

PR #1010, opened 2017-02-16
THT Model Directories

SchrodingersGat:
This PR renames the directories for various 3D model libs, from

`xxx_Throughhole.3dshapes`

to

`xxx_THT.3dshapes`

This is to match the renaming for many new repositories.


SchrodingersGat:
Some .kicad_mod files will need to have their references changes e.g. https://github.com/KiCad/Capacitors_THT.pretty/pull/13

jkriege2:
nice work!
JAN

PR #1011, opened 2017-02-16
New library for Microchip PIC24 family.

jophut:
New library for the Microchip PIC24 family with a PIC24FV32KA304-I/PT microcontroller.

[image](https://cloud.githubusercontent.com/assets/3954707/23025930/c7d49a96-f45f-11e6-83c6-85126d292b19.png)


jkriege2:
Hi!

thanks for the contribution!

First of all: We have a set of rules for symbols and footprints, called KLC. You can find them here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention There also exist a set of Python scripts (https://github.com/KiCad/kicad-library-utils) that check some of these rules and even correct some bugs. Finally we are working on an overhaul of these rules (mostly clarifications, adding what we learned of the pas months, better explanations, example images). The current version is here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC So please orient your PR on the rules in the latter document!

Now for a more detailed review: Please change these things on your symbol:
1. outline of the yellow box should be 0.01" thick
2. footprint should be `Housings_QFP:TQFP-44_10x10mm_Pitch0.8mm` and fpfilter accordingly `TQFP*10x10mm*Pitch0.8mm*`
3. pin length 200mil is way enough ;-)
4. move all pins onto a 100mil grid, these are not: 16,17,28,29,39,40
5. remove component name from description (this info is already present in the component name istelf ;-)

best,
JAN

jophut:
1. Done.
2. Footprint field updated, and fpfilter set to: TQFP*44*10x10*Pitch0.8mm*
note the included '44', or is this discouraged?
3. Done.
4. Done
5. Done

checklib.py found no violations.

best,
Jop

jkriege2:
Hi!

yes adding pin-count is discouraged. The number of pins in your symbol should match the number of pads (including possible bottoms pads) in the footprint, so we can use a footprint-count filter. This means that NC pins should be added to the symbol (although hidden). The FPFilter should then look e.g. like this `TQFP*10x10mm*Pitch0.8mm*`, i.e. including the units.

Best,
JAN

jophut:
Hi JAN,
Updated the fpfilter.
Maybe add a section for fpfilter to the KLC's?

Also accidentally put pin 12 and 13 on top of each other, fixed now.

Br, Jop

jkriege2:
I will check soon ... for now We're currently reworking/revamping the KLCs. See here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC
Hope we get that to the official page soon! If you have any comments, use that thread: https://github.com/KiCad/kicad-library/issues/879

jkriege2:
OK!

I final thing: Pin 18 should not be set as INVERTED, as that is already covered by the text (overline). Also if I understand it correctly the MCLR is only a special function and RA5 can be I/O, so the pin should be bidirectional as the other port pins!

When that's fixed, I think we can merge!

Thanks for the work you put in 👍 

JAN

jophut:
Agree, Fixed.

PR #1015, opened 2017-02-17
Added Analog Devices RF components

yo3iti:
Hello everybody,
Added AD640, AD641, AD8302, AD8306, AD8307, AD8309, AD8310, AD8311, AD8312, AD8313 - RF power sensors to Analog Devices library.


jkriege2:
Hi!

I just had a short look at your PR (close look to come ;-) ... Some first remarks:
1. Please remove all those `.DS_Store`-files from your PR ... don't know what they are
2. Why do you have symbols with just numbers as names, e.g. `5962-9095502M2A`. Are those the military variants? I'm not sure whether we really need them as separate symbols. Usually we try to cover at least the variants with different packages as separate symbols and then you can (optionally) add variants with e.g. different temp. ratings ... as ALIASes
3. If a symbol has only one package, set that package in the footprint field (format `LIBRARY:Footprint`), but still keep the FPFilter, so variants, as e.g. `_HandSoldering`, can be matched.
4. For e.g. `AD640BE` you give a list of packages, but in the DS that one is ONLY `20-Terminal LCC`

Finally: We have a set of rules for symbols and footprints, called KLC. You can find them here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention There also exist a set of Python scripts (https://github.com/KiCad/kicad-library-utils) that check some of these rules and even correct some bugs. Finally we are working on an overhaul of these rules (mostly clarifications, adding what we learned of the pas months, better explanations, example images). The current version is here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC So please orient your PR on the rules in the latter document!

best,
JAN

yo3iti:
Thank you, JAN. Will do.
BR

yo3iti:
Thank you, Jan.

This is my first and it was mostly a test.

.DS_Store come from my mac, I believe. I need to find a way to remove 
these from the final commit because these are automatically set in the 
folder with any activity.

Your answers largely clarify many questions I previously had.

Looking forward,

Regards,

Miron (YO3ITI)

jkriege2:
You can add the .DS_Store to you local  ignore list (not the one of the repo though). Also you don't need to close this PR: Simply push changes to this branch and they will appear here ... e.g. simply git-delete the .DS_Store and push ;-)

JAN

yo3iti:
Thank you, Jan, for your patience and guidance.

BR

M

jkriege2:
PS: One more thing: I noticed you pushed to the master-branch of your fork ... usually (especially when you have several parallel PRs) it's better to create a new branch from your master for every PR.

jkriege2:
PPS: Let me know if you have any other questions ;-) ... otherwise we will simply discuss your PRs in the respective threads.

yo3iti:
Ok. Good to know. Thank you. Will do. :)

BTW, is there a way to avoid working in parallel on same components with 
other colleagues ? I asked a question on the mailing list but no reply.

I also asked the question regarding the aliases: how to deal with 
symbols that have different names but same packages (e.g. the military 
ones in your first email were good examples). You answered to that so my 
understanding is that:

- we try to accommodate one component / footprint

- several names per footprint are taken as aliases.

How to you set the package in the footprint field (format 
|LIBRARY:Footprint|) ? Via a simple text entry ? I assume there is no 
separate field for the same (like Altium has, for example).


BR and thank you,

Miron

jkriege2:
Yes, basically you can do everything as an alias that only requires changes in description, keywords or datasheet. For all other changes you need a new footprint (with the current library format).

This should explain where/how to set the datasheet ...
2017-02-18 13_02_54-part library editor d__kicad_kicad-library-review_library_dc-dc lib

jkriege2:
There is no "blocking" mechanism. If you see someone else working on the very same symbol, just talk to them. If two people edit the same lib: It's usually no problem. Whenevere one of the two (or more) PRs is merged into the KiCAD-master, you simply pull the KiCAD-master into you branch and merge the two. In many simple cases, git handles the merge automatically and you only have to accept the changes.

Best,
JAN

yo3iti:
Thanks for all emails, support and patience.
BR
Miron

PR #1020, opened 2017-02-19
Added 3D model for 1W and 3W LED

MartinStej:
3D model created according to https://www.gme.cz/data/attachments/dsh.518-234.1.pdf
 Package name seems to be "1W 3W LED"


SchrodingersGat:
Nice model, thanks :)

PR #1021, opened 2017-02-19
Fix footprint fields

jkriege2:
This PR fixes several footprint fields from old standard to new (LIB:FP) and accounts for diverse new footprint names


SchrodingersGat:
Thanks Jan this is great

PR #1022, opened 2017-02-19
 updating all script generated Capacitors_SMD adding STEP and WRL parametric model

easyw:
models are generated by parametric scripts, WRL with KiCAD suggested materials

discussion related at

https://forum.kicad.info/t/contributing-to-update-of-vrml-librarys-to-step/5049/
capacitors_smd


SchrodingersGat:
I do not think that we need `_HandSoldering` variants of the models. Are there any differences in the models between the normal and hand-soldering versions?

easyw:
Because of the Kicad Library convention.
Rule 10.8 states that the 3d file should be named the same as the footprint, so I placed both

SchrodingersGat:
10.8 has been updated to state that such a violation is ok :)

easyw:
but the kicad_mod has still the _HandSoldering inside ... BTW the model size is 46KB

jkriege2:
Hi!

I reworked the footprints and the `_HandSoldering` now reference the "normal" 3D model .... so you can remove the doubled models:
https://github.com/KiCad/Capacitors_SMD.pretty/pull/13

Best,
JAN

easyw:
removed _HandSoldering 3d models (aligned to std footprint)

SchrodingersGat:
Thanks @easyw for the contribution

Shackmeister:
perhaps the same should be done for the SMD resistors

jkriege2:
Done!
https://github.com/KiCad/Resistors_SMD.pretty/pull/15

PR #1025, opened 2017-02-19
added Pololu A4988 and DRV8825 breakouts

jkriege2:
see https://github.com/KiCad/Modules.pretty/pull/4 for footprint

2017-02-19 22_36_32-part library editor d__kicad_kicad-library_library_modules lib
2017-02-19 22_36_51-part library editor d__kicad_kicad-library_library_modules lib


SchrodingersGat:
@jkriege2 I think you have the pinout incorrect for the A4988 - https://a.pololu-files.com/picture/0J3360.600x480.jpg?d94ef1356fab28463db67ff0619afadf

jkriege2:
No, the pinout is correct (just checked again ... also I have a cheaper China-version of that device here) ... The only thing I did was rotate 180° with respect to the image you sent: They do not really have a pin1-mark, so I took the square GND-pin in the bottom-right corner of the Pololu image as pin1 and rotated it top-left. then pin-ordering is as for any DIP-package!

Would you do it some other way?

JAN

SchrodingersGat:
I think that in the absence of a clear pin-1 indicator in their datasheets / drawngs, we should at least match the orientation of the picture they provide (e.g. ~ENABLE~ pin is 1)

jkriege2:
Hhmmm .. not so happy with that ... I would look at it the other way round: the rectangular pin is the only mark they have (omitting the second GND-pin which is also rectangular) and the footprint should somehow match that so people put it in correctly. That is currently given with the footprint you already merged. If we renumber, I think we would HAVE to do a specific footprint that has the correct pins rectangular, otherwise it'll be a mess ...

Sounds valid?

Best,
JAN

SchrodingersGat:
Are there any markings at all on the module (besides the square pin) to indicate pin 1?

I'm not opposed entirely to leaving it as it is, just want to avoid any potential confusion.

jkriege2:
On the China-version I didn't find any other and on the images on the web of the Pololu original neither ... that's why I decided to go for the rect pads

SchrodingersGat:
I suppose that the pins are names on the B.SilkS layer of the module itself. Probably ok with that.

PR #1030, opened 2017-02-21
Adding a few Si and InGaAs photodiodes

yhlien:
Adding a few Si and InGaAs photodiodes:

FDGA05
FDS010
FDS025
FDS100
FGA01

from Thorlabs.


jkriege2:
For the 3-pin packages, please add the NC-Pins to the symbols (as invisible NC-Pins).

Best,
JAN

yhlien:
Hi Jan,

I have updated the symbols with the third pin invisible.

Best regards,

YH

jkriege2:
Hi!

A few comments:
1. FDGA05: The package does not really match, as the datssheet states that the pins are positioned differently on this device (pin2 in center). Also pin3 should be visible as it is GND according to datasheet
2. FDS010: Also here GND should be visible
3. FDS025: Pin2 should be visible as CASE
4. FDS100: Pin2 should be visible as GND

Maybe you can come up with an extension of the symbol that signifies that the third pin connects to the can.

Best,
JAN

yhlien:
Hi Jan,

Many thanks for your comments.

Actually I am bit confused by the datasheets on the supplier website.

For example, the two compatible sockets for FDGA05 have identical pin locations but different from the pin locations of FDGA05. The datasheet does not include the enough information about pin locations.

The information about the unspecified third pin also seems to be inconsistent to me. Some diodes are noted as case but others as ground. I think the unspecified pin is usually connected to the case but the case might not be grounded. I have one or two photodiodes in my lab so I can doublecheck them but unfortunately I can do very little for others...

Hopely I can make a new footprint for FGA01 and FDS025 because the TO-46 case for these actually has a ball lens rather than a flat window.

Would you please advise me how I should proceed on these symbols? Should I try to finish the symbols as much as I can and wait for the feedback from other users or simply stop doing?

Best regards,

Yu-Hung

jkriege2:
I would assume that the third pin is n all cases connected to the CASE (then maybe call it CASE for all) and usually one Should connect it to GND (the diode itself should be isolated from the can). So modify the symbol to look something like that:
2017-02-22 06_53_40-part library editor d__kicad_kicad-library-review_library_device lib
2017-02-22 06_54_06-part library editor d__kicad_kicad-library-review_library_device lib

Then we can merge the symbols and see for the footprints (maybe already choose a footprint-name for the FDGA05). I would put that into https://github.com/KiCad/Opto-Devices.pretty as e.g. ` 	PhotoDiode_FDGA05`.

JAN

yhlien:
How about the symbol from ANSI/IEEE standard 315:
image

The circle enclosing the diode should just represent the case:

image

Meanwhile, I have some generic questions about the symbols. Firstly, is there any specific standard like ANSI/IEEE or IEC is preferred/adopted in KiCad community?

Secondly, I am curious why the horizontal diode symbols are orientied like cathode(left)-anode(right) but not the opposite? I understand the direction does not necessarily mean the flow of the signal but the triangle usually suggests some sort of direction. In that case, the orientation should not follow KLC, i.e. left to right?

Thirdly, is there any specific reason the origin of the simple diode symbol not setted at the center of the triangle but at the middle of cathode? My personal feeling is the symbol looks more symmetric as the center of triangle is the origin. One more advantage is the pins are symmetric respected to rotation.

I am sorry for so many questions. Any comment or suggestion is very appreciated.

Best regards,

Yu-Hung

jkriege2:
Hi!

we want to enforce IEC-symbols (see `device.lib`). The symbols here do not conform completely (IEC has non-filled Diode triangles with a hor. line thorugh it ... again see `device.lib` or the LEDs above). We will have to modernize this lib at some point.

About the symmetry: Anchro is at the Center between the Pins and it is in halfway through the triangle, but not in ist Center of mass ... but in x-direction everything is symmetric!

About the orientation: I thank that's not that important, but for most diodes with Special Diode-housings (e.g. D=-35...) we use pin1=K and pin2=A ... therefore pin1 is on the left.

Best,
JAN

yhlien:
Unfortunately I could not find the pdf for IEC standard and am quite confused by the information on Google. That is why I chose ANSI/IEEE for the first draft. I will be very happy to follow IEC standard.

I will redraw the diodes with unfilled triangle  and a horizontal through line. Meanwhile, I would like to confirm the bracket shape representing the case in your first drawing (LED) is also from IEC?

Best regards,

Yu-Hung

yhlien:
Hi Jan,

I quickly made the drawing:

image

Please let me know whether you like it and I would proceed to replace the ones I made.

Best regards,

Yu-Hung

jkriege2:
NICE work!

now please set the footprint-fields of the dveices with a special TO-46 package ... then we can merge this!

Best,
JAN

yhlien:
Hi Jan,

Many thanks for your recognisation.

I have committed a TO-46_3_Lens footprint to TO_SOT_Packages_THT.pretty.

Meanwhile, I also added another three Si PIN photodiodes from Hamamatsu. Please feel free to let me know if there is any issue existed.

Best,

Yu-Hung

jkriege2:
Hi!

I made a PR for these TO-46 variants ... We should wait util that's merged and then we can update the footprints here accordingly and merge!

see https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/19

cu
JAN

yhlien:
Hi Jan,

I am still waiting for the response from Thorlabs (the manufacturer) to clarify whether there is any mistake in their footprint drawing. I guess we can wait until everything is settled.

Best regards,

Yu-Hung

jkriege2:
Hi!

I can't find your PR to `TO_SOT_Packages_THT.pretty` ...?

cu
JAN

jkriege2:
+ did you push your Hammamatsu diodes?

SchrodingersGat:
https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/19 is merged

jkriege2:
@yhlien 

Hi! I tried to close and reopen this PR today to run our new KLC-check scripts ... but it seems you deleted your branch and I can't open this again???
Are you still on this one? And did you hear from Thorlabs?

best,
JAN

yhlien:
Hi Jan,

Sorry for replying late. I did not notice your message. I am not sure whether I touched my branch but I just resubmitted them.

I did hear from Thorlabs and can confirm the unconvential pin config is indeed used. I just started to learn FreeCAD and hopefully I can make the corresponding footprints and 3D models...

Best,

Yu-Hung

PR #1031, opened 2017-02-21
updating all script generated Housings_SOIC adding STEP and WRL

easyw:
models are generated by parametric scripts,
STEP file size optimized, WRL with KiCAD suggested
materials

discussion related at

https://forum.kicad.info/t/contributing-to-update-of-vrml-librarys-to-step/5049/
housings_soic 3dshapes


SchrodingersGat:
Fantastic, keep them coming!

One thing I would like to do (in the future) is add a file e.g. LICENSE.md in each subdir of the 3d models, stating where the models have originated.

I will not do that yet as I am currently trying to work on moving all 3D models to their own repository...

easyw:
License is a really nice thing to add... Many commercial users are afraid of GPL... for that reason I suggested at the forum to add similar to the gEDA license exception for KiCad 3D model libraries
http://wiki.geda-project.org/geda:license

SchrodingersGat:
Great, that's very helpful. I will let you know when I get a separate 3D repo going, I think the work you have been doing will be integral to that.

PR #1036, opened 2017-02-22
adding TSV911

jeromebeguin:


jkriege2:
the given FPFilter does not match the given footprint!

jkriege2:
as this only matches one specific variant, please rename to `TSV911ILT`

Best,
JAN

jkriege2:
Nice Work!

Thanks,
JAN

PR #1037, opened 2017-02-22
C128880MA & Flir_LEPTON

jeromebeguin:


SchrodingersGat:
Hi @jeromebeguin please change the field size of the `Footprint` and `Datasheet` fields and we should be good to go :)

SchrodingersGat:
Thanks :)

PR #737, opened 2016-10-04
added several missing linear ICs

jkriege2:
# linear.lib
## NEW/MODIFIED:
* [x] LM555N NE555 ICM7555 LMC555 TLC555 NA555 SE555 SA555
* [x] ICM7556 NA556 NE556 SA556 SE556
* [x] NE567 SE567
* [x] L272D 
* [x] L272M
* [x] LM321
* [ ] TL061 TL071 TL081
* [x] LT1492
* [x] LT1493
* [x] NE5534 SA5534 LM101 LM201 LM301
* [x] LMV321
* [x] INA128
* [x] INA129
* [x] INA326
* [ ] INA327
* [x] MAX4238ASA
* [x] MAX4238AUT
* [x] MAX4239ASA
* [x] MAX4239AUT
* [x] OPA333D OPA340D
* [x] OPA333DBV OPA340DBV
* [x] OPA2340
* [x] OPA2333D
* [x] OPA333DCK
* [x] OPA340P
* [x] OPA4340D
* [x] OPA4340DBQ
* [x] OP07 OP077 LT1363
* [x] LM4250
* [ ] MCP601P
* [ ] MCP601R
* [ ] MCP601SN
* [ ] MCP601ST
* [ ] MCP603
* [ ] MCP603P
* [ ] MCP603SN
* [ ] MCP603ST

## ALIASES:
* [ ] LF155 LF156 LF255 LF256 LF257 LF357 LF355 LF356
* [ ] MCP6004 TL084 TL064 LMV324 LMC6484 MCP604
* [ ] TL062 TL072 TL082 NE5532 LMV358 TS912 TLC272 MCP602

# mototr_drivers.lib:
## NEW:

* [ ] L293 
* [ ] L293D

# interface.lib:
## NEW:

* [x] MC1413 MC1413P SN75468 SN75469 ULN2002 ULN2002A ULN2003 ULN2003A ULN2004 ULN2004A
* [x] ULN2801A ULN2802A ULN2803A ULN2804A ULN2805A
* [ ] ICL7667

## ALIASES:
* [ ] MAX3232

# dc-dc.lib:
## NEW:
* [ ] LT1073
* [x] LM2734

## ALIASES: 
* [ ] LM2575-12BT LM2575-3.3BT LM2575-5.0BT LM2575BT-ADJ
* [ ] LM2575-12BU LM2575-3.3BU LM2575-5.0BU LM2575BU-ADJ

# audio.lib:
## NEW:
* [x] MSGEQ7

[...]

PR #1039, opened 2017-02-22
adding BC817DS & CSD17579Q3A

jeromebeguin:


jkriege2:
please fix conflicts!

jkriege2:
Hi!

please fix:
- where does the `BC818-40` come from (in the DCM)? There is no symbol for that ...
- transistors.dcm causes errors when loading in current nightlies!
- since we discussed a bit about hidden pins: You could use a standard MOSFET-symbol from device.lib for CSD17579Q3A and stack 2 hidden S-pins over the one visible one.

Best,
JAN

jeromebeguin:
I don't know what is BC818-40, I don't have it in my library. 
I thought device.lib was abandoned ?

jkriege2:
yes, but it's there in your PR ... maybe a remenant from a merge-gone-wrong? Please remove it, see whether there are any errors with current versions and correct this PR.

jeromebeguin:
Sorry I'don't have this in my lib. And I'm in this current branch. Don't know what happening. I cannot erase something I don't have. 
 image

jkriege2:
simply erase it in the DCM file ... it's a text-file that can be edited with any editor!

jeromebeguin:
Ok, I can do that.

jeromebeguin:
Here we are, could you merge ?

jkriege2:
Hi!

OK, so now for a review of the symbols
1. Please change the symbol for `BC817DS` to be like the other dual transistors that we have in this lib (e.g. `BC846BS`), i.e. make two subsymbols in the symbol!
2. Please fix the errors reported by Travis
3. For the `CSD17579Q3A` you might also consider having a symbol where the doubled/tripled pins are stacked (just a proposal)
4. What about having the ESD diode from the datasheet in the symbol for `RV3C002UN`? Althought that might make the symbol significantly larger ... (just a proposal)
5. The `XR20M1172` comes a bit as a surprise ... next time, please split the PRs by lib, which makes it easier to review the stuff! ... Now about the symbol: Please rename to `XR20M1172IL` as this postfix denominates the QFN32 package ... otherwise looks good to me.

best,
JAN

evanshultz:
A couple follow on questions/concerns:
3. See clause 4.3 of KLC for more details on pin stacking.
5. The "IL" suffix captures both temp range and package. KLC (clause 3.6) states that when packages for a different symbol, append the symbol with the simplified package name. In this case, it would be `XR20M1172_DFN`.

KLC can be found at: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention.

@jeromebeguin Thanks for the contribution!!

jeromebeguin:
I'm sorry but I don't have the time to make changes all the time. Why don't you asked me to transform directly the BC817DS and the CSD17579Q3A right at the beginning ? It's really hard to add new component to library. I'm taking the time to share because Kicad is free and I'm using it at work. But I could not spend half of my design time to build and correct several time the same component. So please accept the minimum to be useful and if somebody have the time to make it nice and pretty he can modify it as well later on. I'm happy to collaborate at this project, I think it worth. but the way to share is really long and not convenient for a work environment. Please take this in account. Travis helps a lot, this is a big improvement, but I'm sure you can simplify and accelerate the process. 

Best

Jérôme

jkriege2:
Hi!

if I remember correctly, there were already conflicts right when I had a look at it the first time and then I was unable to load your transistors.dcm due to the error I mentioned above ... So all the review process was a bit delayed UNTIL I could have a look at the symbols. Please consider two things for this lib:
1. We try to move towards a common style on most components, so we try to NOT acept stuff that does not fit so nicely ... If we would accept every symbol, we will end up having to find someone that makes everything nice, which will probably never happen ... so: Yes this is a bit of a burdon on the commiter, but will help the overall lib quality.
2. Please also consider: We are also only a small team of maintainers that have jobs of their own on the side therefore we also only contribute a bit of time ... and sometimes things do not go as smoothely as with someone ONLY committed to building this lib!

One way you can easily improve the review process is by providing some info when pushing your PR. usually what is helpful is:
- explain what you did (a bit more elaborated than in the title ;-), e.g. name the components you added/changed
- add screenshots of modified/new components
- give datasheet links

Here's an example of how this could look (and yes that PR of mine was WAYYYY too big ;-)
https://github.com/KiCad/kicad-library/pull/737

Best,
JAN

PS: I will have a look at the changes this evening ...

jeromebeguin:
Thank you for that feed back. I'll try to give you more info to make the review easier.

PR #1042, opened 2017-02-22
Adding LM6181 into library linear

yhlien:
Adding LM6181


jkriege2:
Hi!

this is a bit mixed up now ... please close this PR and
1. make one PR for LM6181 only (just create a branch from your master to do that)
2. add the Hammamatsu PDs to you other PD PR (https://github.com/KiCad/kicad-library/pull/1030) or add a third one for these (I would prefer the first ... then I can review them all at once ;-)

best,
JAN

PR #1045, opened 2017-02-23
Add H11L1 optocoupler

evanshultz:
Image:
image

The FPFilter is a bit hard here since it comes in multiple widths and both THT and SMT. I hope what I did will work, since it's quite generic.

Datasheets:
https://www.fairchildsemi.com/datasheets/H1/H11L1M.pdf
http://www.everlight.com/file/productfile/h11l1.pdf


jkriege2:
VCC+GND should not be hidden, ... please compare with other opto-couplers from this lib ...

evanshultz:
OK. I did find one with a logic output in this lib.

jkriege2:
Hi!

Please solve the FPFilter problem as e.g. in this OC from opto.lib:
2017-02-24 19_21_44-part library editor d__kicad_kicad-library-review_library_opto lib

Also pin3 is better positioned like that (IMHO) ... at least it's done that way in most other libs:
2017-02-24 19_23_03-part library editor d__kicad_kicad-library-review_library_opto lib

JAN

evanshultz:
Thank you, Jan. Done. Note that I added the 3 DIP SMD footprint shown on those datasheets.

jkriege2:
nice work!

Thanks,
JAN

PR #1046, opened 2017-02-23
Add HEF4093B to LIB

evanshultz:
Image:
image

Not sure if this is the right lib. If so, I'll add the DCM file update.

Kinda big and unwieldy symbol, but I'm not sure how else to make this package. Please advise if there are suggestions.

Using hidden pins as that issue hasn't been settled and this the current standard, as I understand it.

Datasheet:
http://www.nxp.com/documents/data_sheet/HEF4093B.pdf


SchrodingersGat:
As we haven't solved the hidden pins drama yet, please add .dcm file here

evanshultz:
Added DCM.

Regarding hidden pins, will update. Am I correct that we're using separate logic gates and then a single power gate? Is the usage of gates with visible power pins clear in KLC 2.0?

SchrodingersGat:
@evanshultz you are correct. Use multiple identical units for the logic, and a separate power unit with PWR and GND pins (visible!)

evanshultz:
It seems now that hidden power pins are disallowed, but they shouldn't be on a separate unit (because placement of the refdes and value are fixed on all units). Has KLC 2.0 settled enough to make this happen, or should I wait a bit to change this symbol?

jkriege2:
I think you can do a separate unit, as KiCAD (at least in current nightlies) auo-positions the labels ... so it's gonna be fine as soon as you add the symbol to a schematic!

Best,
JAN

evanshultz:
@SchrodingersGat @jkriege2 Updated this symbol with new style having a separate power unit. Image (yes, from KiCad 4.0.x):
image

What do you think? Is this the style we want for logic gates and opamps and such?

Questions:
1. Compared with an opamp, discrete transistor, etc. this symbol is huge! Any thoughts on logic gate size?
1. Also, I removed the Schmitt trigger symbol. Do we want to include this on relevant logic symbols?

poeschlr:
> Compared with an opamp, discrete transistor, etc. this symbol is huge! Any thoughts on logic gate size?

Yes they are really huge. If you use a 50 mil grid for the outline they might get smaller.
I made a quick example:
screenshot from 2017-03-31 11-53-54
Here the unit i designed exported as .sym (Uploaded as txt because .sym is not allowed.)
[HEF4093B_small.txt](https://github.com/KiCad/kicad-library/files/885200/HEF4093B_small.txt)

> Also, I removed the Schmitt trigger symbol. Do we want to include this on relevant logic symbols?

What does the norm say that defines these symbols?
The current ieee would add this marking. (I think it is the current version.)
ieee standard graphic symbols for logic functions

evanshultz:
@poeschlr Thanks! I didn't find anything in IEEE 315-1975 (and maybe I just missed it!), which is referenced by KLC, but I didn't look in newer docs. Here's my shrunken version for consideration (I want to play more):
image

If this looks better, I'll move onto the logic lib once I finish linear and use this as a symbol reference.

SchrodingersGat:
The newer smaller symbol is much better. I think perhaps there should be some more separation on the power-part, e.g. move GND and VCC pins further apart. It looks like there is one label that says "GNDVCC".

Perhaps also a rectangle box around the labels?

It's a pity that we cannot place the reference "per unit" but that's what we have to live with.

evanshultz:
To get this merged, at least so it's in the lib and can be captured by a scripting method like at https://github.com/KiCad/kicad-library/issues/605#issuecomment-313225824, what needs to be done?

@SchrodingersGat @jkriege2 @poeschlr 
I pushed the text closer to the pins (10mil away) to give more separation between "VCC" and "GND":
image

I think that obviates the need for a box, but that's me.

Any other comments or objections to this symbol?

poeschlr:
The symbol looks good to me. But merging is required.

poeschlr:
Closing because the symbol has been merged in another pull request.

Issue #986, opened 2017-02-08
Alternate symbol for dual diodes


evanshultz:
While the "opposed diode" symbol is nice in some cases, for others I find it unwieldy, For example, when using a common cathode rectifier on the output of a flyback supply. Having the diode symbols shown in parallel, next to each other, is more attractive to me.

Here are some example parts that, to me, better fit the symbol described above:
- [STPS40L15CT](http://www.st.com/content/ccc/resource/technical/document/datasheet/a3/ec/15/c3/8c/15/4b/98/CD00001105.pdf/files/CD00001105.pdf/jcr:content/translations/en.CD00001105.pdf)
- [MBR20200CT](http://www.onsemi.com/pub_link/Collateral/MBR20200CT-D.PDF)

Here is the symbol:
![image](https://cloud.githubusercontent.com/assets/1936989/22745182/912d1504-edd4-11e6-9dde-eb5882b77f2a.png)

And the symbol lib instance:
````
# D_Schottky_x2_KCom_AKA_Alt
#
DEF D_Schottky_x2_KCom_AKA_Alt D 0 40 Y N 1 F N
F0 "D" 0 250 50 H V C CNN
F1 "D_Schottky_x2_KCom_AKA_Alt" 0 -250 50 H V C CNN
F2 "" 100 -50 50 V I C CNN
F3 "" 100 -50 50 V I C CNN
$FPLIST
 TO-???*
 D_*
$ENDFPLIST
DRAW
C 0 100 10 0 1 0 F
P 2 0 1 0 -100 -100 -100 0 N
P 2 0 1 0 100 -100 100 0 N
P 2 0 1 0 100 -100 100 0 N
P 4 0 1 0 -100 0 -100 100 100 100 100 0 N
P 4 0 1 8 -50 -100 -150 -100 -100 0 -50 -100 N
P 4 0 1 8 150 -100 50 -100 100 0 150 -100 N
P 4 0 1 8 150 -100 50 -100 100 0 150 -100 N
P 6 0 1 8 -75 25 -50 25 -50 0 -150 0 -150 -25 -125 -25 N
P 6 0 1 8 125 25 150 25 150 0 50 0 50 -25 75 -25 N
P 6 0 1 8 125 25 150 25 150 0 50 0 50 -25 75 -25 N
X A1 1 -100 -200 100 U 50 50 1 1 P
X K 2 0 200 100 D 50 50 1 1 P
X A2 3 100 -200 100 U 50 50 1 1 I
[...]

PR #1047, opened 2017-02-23
Add BCV61/62 and generic dual transistor version

evanshultz:
Image:
image

Notes:
- I'm undecided on text placement. Probably to the right is better since wires will naturally go vertical?
- Is having the base connections this way acceptable? It's weirdly off-grid but looks like the datasheet.
- I did not add and entry into device.dcm since they're generic. I can if desired.
- The FPFilter may be too specific for the generic part. Please advise.

Datasheets:
http://www.nxp.com/documents/data_sheet/BCV61.pdf
http://www.nxp.com/documents/data_sheet/BCV62.pdf


jkriege2:
Hi!

a few things to change please:
1. add description to the new generic symbols in `device.lib` (i.e. put them also into `device.dcm`)
2. I'm not sure whether the FPFilter is needed for the generic symbol ... I would rather say no, as you don't really know in which package they might also come
3. Please correct line-endings to UNIX (LF-only).

Thanks,
JAN

evanshultz:
1. Done. Oops.
2. Yeah, I see that. But a 4 pin connector wouldn't be suitable. Perhaps we wait a couple days for feedback in Issue #986?
3. I've been seeing a lot of talk about line endings recently, and I use Windows. But what I don't see is what to do about it, just that it was a problem or that it was done. How to correct line endings?

evanshultz:
3. Got it. Notepad++ can convert automatically. Just a sec.

evanshultz:
Just above the comment box there is a conflicts section but I'm not seeing that it pertains to my PR. In fact, it's not clear to me what the conflict is. Is it line endings? When I copy part of the committed text to Notepad++ and view line endings, they're all only LF which is correct. Is it something else?

poeschlr:
> Just above the comment box there is a conflicts section but I'm not seeing that it pertains to my PR.
At the bottom of this conversation stuff i see this:
screenshot from 2017-02-25 18-43-01


> it's not clear to me what the conflict is.
someone else has changed the files since your branch deviated from the master branch.
(The last commit to the transistors.lib file was the changing of some footprint filters by @jkriege2)
To resolve it: merge in the changes done on this files.

jkriege2:
The conflicts mean that `transistors.lib` and `transistors.dcm` have changed in the master. This prevents a merge. So please pull/merge the master into you branch. That should solve the conflicts!

evanshultz:
I think that should do it, but the PR isn't being checked. How to trigger that? I will close and reopen.

evanshultz:
Hmm.... I pulled the current master for each `transistors` file from the repo, then re-added my two new parts, and still errors throughout the file. Any tips on what I'm doing wrong?

library/transistors.dcm
```diff
@@ -82,7 +82,7 @@ $ENDCMP
 #
 $CMP 2SC1945
 D 6A Ic, 80V Vce, Silicon 27MHz RF Power Transistors NPN, TO-220
-K RF Power Transistor NPN
```
poeschlr: Example for a changed line ending or removed whitespace

library/transistors.lib
```diff
@@ -634,8 +657,8 @@ ENDDEF
 #
 DEF BC807 Q 0 0 Y N 1 F N
 F0 "Q" 200 75 50 H V L CNN
-F1 "BC807" 200 0 50 H V L CNN
-F2 "TO_SOT_Packages_SMD:SOT-23" 200 -75 50 H I L CIN
```
poeschlr: Here is another example of a changed line ending

poeschlr:
The curse of the line endings it seems.
And there are also some whitespace characters removed.
(See my comments above)

poeschlr:
> Hmm.... I pulled the current master for each transistors file from the repo, then re-added my two new parts, and still errors throughout the file. Any tips on what I'm doing wrong?

To sync with master do the following.
git checkout master
git add remote upstream https://github.com/KiCad/kicad-library.git
git fetch upstream
git pull upstream/master
git checkout [your branch]
git merge master

check if it worked with git status
possibly you need to git add -u
git commit -m [a commit message]

jkriege2:
To make this simpler:
remove `transistors.*` from this PR and open a fresh new one from a new branch branched off from an up-to-date master and add the special transistors there ... sometimes it's easier to start all over than putting in a lot of work to fix a crooked branch ...

Best,
JAN

PS: If you do, just post the second PR here, so I can review them together to speed things up a bit ;-)!

PR #1057, opened 2017-02-24
Add FPFilter to potentiometer symbols

evanshultz:
Currently there is no footprint filter for pots. This PR will select footprints from the Potentiometers library with the format `Potentiometers:Pot*`.


evanshultz:
@SchrodingersGat Looking for a little help here... I wanted to add FPFilters as above, but it appears the symbols I modified violate KLC and thus wouldn't be valid for merging.

So in order to accomplish the original reason for this PR I must modify the symbols to meet KLC?

jkriege2:
Hi @evanshultz !

the KLC-checks are not very good for most simple symbols from device.lib. Many rules only apply to box-symbols (e.g. resistors, diodes ... have pins on a 50mil-raster and I think it's generally agreed that this is OK like that, because otherwise they would get way too large). Unfortunately there is no way to tell the scripts what set of rules to apply to which symbol ... so please: ignore those errors!

About the filter itself: All Pots (new symbols) currently use the prefix `Potentiometer_` So a filter `Potentiometer*` would be better I think ... Or are there any other footprints that could be used, but are not matched by the longer version? If we decide to drop the prefix, or modify it in future, this can be changed here then too.

JAN

evanshultz:
I was wondering about the using the full `Potentiometer*`. PR updated. Thanks for the review and comments!

jkriege2:
Thanks for spotting!
JAN

PR #1060, opened 2017-02-24
Add 20 pin legacy ARM JTAG+SWD header

gauravjuvekar:
Reference design taken from
http://infocenter.arm.com/help/topic/com.arm.doc.faqs/attached/13634/cortex_debug_connectors.pdf


poeschlr:
Could you post a screenshot of the symbol?
This would make the work of the maintainers easier.

gauravjuvekar:
@poeschlr  Sure

selection_089

BTW, the build is failing because of some other invalid .dcm files.

jkriege2:
Hi!

first of all: thanks for the contribution!

... and since this is (as far as I can see) your first PR to this lib: We have a set of rules for symbols and footprints, called KLC. You can find them here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention There also exist a set of Python scripts (https://github.com/KiCad/kicad-library-utils) that check some of these rules and even correct some bugs. Finally we are working on an overhaul of these rules (mostly clarifications, adding what we learned of the pas months, better explanations, example images). The current version is here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC So please orient your PR on the rules in the latter document!

Now for your symbol: 
1. Please orient the symbol on the style usef or other symbols in the lib. I think the USB3-connectors could be a good start:
2017-02-25 09_32_36-part library editor d__kicad_kicad-library-review_library_conn lib
2. As this is not specific to ATMEL (in contrast to the AVR-ISP-connector), please move this to `conn.lib`

Best,
JAN

gauravjuvekar:
@jkriege2 

I've based the symbol on AVR-JTAG-10 (which I now realize doesn't follow KLC). I'll move it to conn.

Can you tell me how to enable the blue pin type label as seen in your USB image?

jkriege2:
Hi!

I marked the button you're looking for:
2017-02-25 11_00_51-part library editor d__kicad_test_addlin_ _kicad-library_library_altera lib

poeschlr:
I'm not sure this feature exists in kicad stable.

gauravjuvekar:
@jkriege2 

part library editor -home-gaurav-kicad_sources-library-repos-kicad-library-library-conn lib_090

jkriege2:
Hi!

then either try a nightly, or have a look into Preferences | Component Editor Options. There's a checkbox "Show pin electircal type" ... if it's not there ... then use a nightly ;-)

cu
JAN

gauravjuvekar:
Heres the latest symbol with male pin indicators. Is there anything that I'm missing from KLC?

selection_091

jkriege2:
Hi!

I would move all I/O pins to the right ... and there maybe group by function (if required/possible)

Best,
JAN

gauravjuvekar:
@jkriege2 

KLC says that all inputs should be on the left and outputs on the right. Do you mean the bidirectional ~SRST~ when you say I/O?

gauravjuvekar:
selection_093

@jkriege2 Grouped the pins more logically and moved bidirectional pin to the right.

jkriege2:
We are currently working on the new KLCs and doing so we discussed stacked pins ... It was decided that they are acceptable if the stacked pins are NOT NC and have all the same name. This applies to yout GND-pins, so you can all move them to the same location. Just hide all of them but one.ABout the ordering:

I would put all outputs/inputs/bidis on the left-corner (as e.g. for the USB-connector) but make several groups there, separated by 200mil space, e.g.:
- TRST, SRST
- TDI, TDO, TCK, TMS (full JTAG-pins)
- other pins (if that grouping makes sense)

Best,
JAN

gauravjuvekar:
@jkriege2 

Not sure if this is what you meant by putting all pins to the left.
I have intentionally kept DBGACK/DBGRQ to the right as these are rarely used.

selection_095

jkriege2:
Hi!

I would move all IO-pins to the right, but with current grouping ... then the style matches e.g. the USB-connectors. Also please change the FPFilter to `IDC*Header*Straight*` and remove the second reference, as you give a definitive footprint in the footprint field (alternatively leave two entries in FPFilters and empty the footrpint field, iff there are several options).

Best,
JAN

gauravjuvekar:
@jkriege2 

See now
selection_103

I know that number of pins shouldn't be mentioned in FPFilter, however, I don't want the 1x20 Pin_Headers matched. Hence, I've explicitly added 2x20 in the FPFilter.

jkriege2:
Nice work1

Thanks,
JAN

PR #1061, opened 2017-02-25
BUGFIX: added missing DCMs for Altera and Siliconi libs + fixed all KLC issues in both

jkriege2:


jkriege2:
@SchrodingersGat please merge this, as it will fix some problems we currently have with Travis!
I will look into the transistors.lib later this weekend ... there seems to be a bit more severe problem there ...

Thanks,
JAN

SchrodingersGat:
Thanks for finding this Jan

PR #1063, opened 2017-02-25
Added 3d shapes for Panasonic EVQPL type push button

dingari:
Introducing 3d modules for Panasonic EVQPL type pushbutton  (both with push-plate and without).

http://industrial.panasonic.com/cdbs/www-data/pdf/ATK0000/ATK0000CE3.pdf

I'm making another pull request in the footprint repository with footprints for these.


poeschlr:
Is it on purpose that this 3d models are completely gray?
(It seems like you forgot to add material or you lost them by exporting.)
screenshot from 2017-02-25 17-08-41

poeschlr:
There seems to be a problem with your mesh. The fillets look weird.
screenshot from 2017-02-25 17-29-08

Did you create this models with a mechanical cad program? (freecad, catia, proe,...)
if so could you also add step files? (you could use kicad stepup to get correctly scaled wrl models.)

dingari:
Hi, these are step models I found online and converted to wrl using freecad.

poeschlr:
You found?
What is the license of these models?
What are the terms and conditions attached to these models?

Most online portals have very restricted licensing models. Nearly all of them state that any form of redistribution is not allowed. (Even most manufacturers have such restrictions attached to their models.)

If the models license is ok. Try to reexport them using the stepup tool by @easyw:
https://sourceforge.net/projects/kicadstepup/

dingari:
Hmm, that's something I overlooked, now that you mention it. I believe I got these from the Panasonic website some time ago. While their terms don't explicitly ban redistribution, I'd think it's safest to keep these out for the time being.

http://www3.panasonic.biz/ac/e_download/control/switch/light-touch/cad/3d_step/evq-plb_plf_plk_5pl-a08_step.zip?f_cd=4183354

jkriege2:
Then we can close this PR?

PR #1064, opened 2017-02-25
Adding Si photodiodes

yhlien:
Adding:
FDGA05
FDS010
FDS025
FDS100
FGA01
S5971
S5972
S5973


poeschlr:
Your pins are not aligned to a 100mil grid.
Also the pin lengths should be at least 100mil.
See the (new) KLC: https://github.com/SchrodingersGat/kicad-library/wiki/KLC Rule 3.1

Screenshot of the currently supplied symbols for reference.
screenshot from 2017-02-25 19-01-59

yhlien:
The screen dump of one revised symbol:

image

Please let me know whether it is working for you.

poeschlr:
The pin placement and pin lengths look ok now.
But as you might have noticed there is still a problem remaining.
The automated build system tells you us that the manufacturer field has a text size of 60mil instead of the 50mil specified in the KLC.
(Yes this is pedantic but it looks like the testscript stops at the first error it finds. This means such a minor error could mask a bigger problem.)

yhlien:
No worries at all.

I tried to use checklib.py to check the library before syncing it. Hopefully it can pass the system this time.

yhlien:
I am bit confused now. Did it fail again? But I could not see the errors?

poeschlr:
I can not see an error either. There seems to be a bug in the travis test system.

jkriege2:
About pin-lengths and grids: Although the KLC do not state that (yet) we do not enforce these rules for small symbols such as diodes, resistors, ... and even transistors. As you might notice most symbols from `device.lib` violate that rule. If we would enforce it, we would get way too large symbols.

So please revert back to the original pin-length and placement (in sync with the generic symbols from `device.lib`).

Best,
JAN

PS: As any component to this lib is also checked by a librarian, we will ignore the Travis/check-lib warnings in those cases ... also it's not really possible to mark a symbol, so one or the other set of rules is applied ... so ...

jkriege2:
Hi!

So have you heard from Thorlabs on the TO-46 packages?

One more thing: Please remove the Manufacturer fields ... we don't accept additional fields according to KLC!

Best,
JAN

yhlien:
Hi Jan,

I have not heard back from Thorlabs yet.

I will remove the manufacturer field.

Best regards,

Yu-Hung

jkriege2:
Hi!

we have our new Travis check scripts online now ... please have a look at the output
2017-03-11 15_19_49-update ftdi lib by mcvanili pull request 1097 kicad_kicad-library

For these diodes you can ignore KLC 4.1 as we do not consider it for small components like resistors, diodes, ...
But please correct KLC 4.9 violations, i.e. set the datasheet field invisible:
2017-04-07 14_40_55-part library editor d__kicad_kicad-library-review_library_atmel lib

Also I noted that you use these footprints:
- `TO-18-3_Window_20`
- `TO-18-3_Lens_15`
- `TO-18-3_Window_30`

Will you add these footprints? And if so what does the extension `_NUMBER` stand for?

Best,
JAN

jkriege2:
Please fix coonflict above by merging in the trunk!

THanks,
JAN

PS: Waht's the status of this PR?

jkriege2:
I had a quick look at the errors from Travis:
1. The errors for 50mil-grid can be ignore for THESE symbols
2. Please hide the datasheet-fields as shown above in the screenshots!

Have you heard from Thorlabs concerning the 3D models?

Best,
JAN

jkriege2:
What's the status of this (news on 3D models)?
I had a short look (again) and found a few minor things:
- please revert the changes to modules.dcm from this PR
- please ensure that the FPFilters match the footprint names!

Best,
JAN

yhlien:
Well, Thorlabs did not reply so I guess I will need to make my own copy to replace those 3D models.

yhlien:
By the way, do you guys agree on the naming of 3D packages for Hamamatsu photodiodes?

SchrodingersGat:

PR #1066, opened 2017-02-25
add L_Wuerth_WE-SL2

MHolzbrecher:


jkriege2:
Did you simply copy the `EMI_FILTER_LL` symbol from device.lib?

jkriege2:
As this is a specific coil, please set the FPFilter to `L*CommonMode*Wuerth*WE*SL2*`. The other option would be to make a series of variants of the original EMI-filter symbol with different pin-orders (as e.g. done for transistors in `device.lib`).

cu
JAN

MHolzbrecher:
Hi,
put it now in device.lib as EMI_Filter_CommonMode_1234. Counting is from top left to top right.

Michael

jkriege2:
@SchrodingersGat What do you think? Please have a look at this PR ...

Thanks,
JAN

SchrodingersGat:
I do not think that specific devices should be in `device.lib` - this is really for generic symbols. Perhaps a new lib `emi_filters.lib` or similar?

MHolzbrecher:
Hi,
added new lib Common_mode_filters.lib. I think this points more to a bare inductor than emi_filters. An EMI filter can consist out of different parts.

jkriege2:
Hmmm ... common mode filter for a separate lib seems quite specific ... why not simply inductors.lib? @SchrodingersGat  What do you think?

SchrodingersGat:
I can see the benefit to separating plain inductors from specific filter components. 

Perhaps `filters.lib` would be a better name? That way it can be used for other types of filters too.

jkriege2:
`filters.lib` seems fine by me ... Please also fix the issues reported by Travis and remove device.lib from this PR.

Thanks,
JAN

jkriege2:
... On the other hands. FILTERS is a function, but inductor is a class of device type (I would more search for device class than by function ... but that may be me only ;-) ... with that argument I would tend to the `inductors.lib` library ... @SchrodingersGat  What do you think?

MHolzbrecher:
Inductors.lib was my original pull request ....
Later on, when you decided what the name of the lib should be, I will start a new pull request.

PR #1073, opened 2017-02-28
KLC fixes for dc-dc.lib

SchrodingersGat:
- Fixed TL497
- Fixed TPS60500DGS

This is also another Travis test (thi s one should not fail!)


SchrodingersGat:
Woohoo! Travis passed :)

jkriege2:
nice!

bruneau77:
To draw first my schema I do not find the SN7404 component in the KICADE library, with its 14 legs, I find a reference, but it does not propose the whole box and places me the symbols of the inverters contained in the SN7404 ???? You can help me ?

jkriege2:
Apaprt from this thread beeing the wrong place to ask such question (I would suggest: https://forum.kicad.info/). There is a `74HC04`, which has the same pinout. Note that this symbol is split into 6 separate inverter symbols and the power-supply-pins are hidden. So add a GND and a VCC-powe-symbol to your schematic to connect them ... this is a bit non-obvious, but might change at some time in the future ;-)

Best,
JAN

Issue #1074, opened 2017-02-28
Duplicated MMBT3906?


yhlien:
Is there anybody experiencing a warning like "duplicated MMBT3906 symbol in transistor library"?

jkriege2:
did you update to the current kicad-library master? Is the problem still there (I tried to fix some problems in transistors.lib a few days ago)?

yhlien:
Yes, I did.  It is still there. I try to open the transistors.lib and find MMBT3906 is also an alias of BC807. I guess that is the origin of the issue...

jkriege2:
this should solve it: https://github.com/KiCad/kicad-library/pull/1075

yhlien:
Unfortunately, the same symptom for IRF540N...

PR #1075, opened 2017-02-28
removed doubled MMBT3906

jkriege2:
it's now there as alias of BC807, thanks @yhlien

see https://github.com/KiCad/kicad-library/issues/1074


yhlien:
glad to be of help. :)

evanshultz:
BC807 and MMBT3906 are both generic PNPs, even though they have notably different ratings, so I suppose it makes sense for them to be ALIASes. It will make the library smaller, at least.

Also note that MMBT3904 is both a unique part and an ALIAS of BC817. Probably will run into the same issue there.

jkriege2:
also fixed thanks @evanshultz

PR #1076, opened 2017-02-28
Fix name field

SchrodingersGat:
This PR fixes all violations of KLC 3.7 (value field must match symbol name)

These errors were causing issues with nightly builds - https://lists.launchpad.net/kicad-developers/msg28290.html


jkriege2:
What are all these tildes `~`in the names? Wasn't there something that parts with tilde are marked as power-supply symbol?
JAN

SchrodingersGat:
Weird, I didn't see those parts. Fixed.

jkriege2:
When removing the tilde: Did you do it in KiCAD or with a texteditor? In the latter case: Could you please open the lib in libEdit and save it there? I think when I ran into that the last time, the ~ was generated by KiCAD ...

cu
JAN

SchrodingersGat:
Ah, right. It occurs when the name field is set to invisible! 

I will add a check for this to KLC.

SchrodingersGat:
It's great when Travis works well :)

SchrodingersGat:
@jkriege2 can you merge this so we can tackle https://github.com/KiCad/kicad-library/pull/1078 ? :)

PR #1079, opened 2017-03-01
Create intersil library and add HIP210x and HIP408xA

evanshultz:
Images:
image

Is this the right way to handle parts with different packages that force different pinouts? Let me know if I should be doing it differently.

All KLC checks pass. I didn't add the non-"A" versions of HIP408x since they're NRND.

Datasheets:
- http://www.intersil.com/content/dam/Intersil/documents/hip2/hip2100.pdf
- http://www.intersil.com/content/dam/Intersil/documents/hip2/hip2101.pdf
- http://www.intersil.com/content/dam/Intersil/documents/hip4/hip4080a.pdf
- http://www.intersil.com/content/dam/Intersil/documents/hip4/hip4081a.pdf


SchrodingersGat:
You could just use the actual part numbers from the datasheet:

* HIP2100IBZ (SOIC)
* HIP2100EIBZ (EPSOIC)
* HIP2100IRZ (QFN)
* HIP2100IR4Z (DFN)

Then add package info to the description string e.g. "SOIC package"

evanshultz:
@SchrodingersGat I thought about that, but it seemed far more confusing to me. How would a user know that "IBZ" is SOIC and "IRZ" is QFN? Even if the info in the description, which I added, that way is less obvious.

And I also noticed that the "I" is for temp grade and "Z" is lead free, which aren't related to the package. So that wouldn't be a good precedent, I didn't think, since we wouldn't want different versions of symbols for parts in different temp grades.

And finally, I found other parts that appended "_QFN" or such to the symbol name in this case. The library still needs clean up so maybe that's discouraged now?

So... that's why I did what I did. What is the best way for KiCad? I'm happy to change to whatever makes sense once we reach consensus.

evanshultz:
@SchrodingersGat @jkriege2 Any comments? How should it be done?

SchrodingersGat:
I would lean towards using the footprint as often the part number can include things like packaging (e.g. tape / reel / tube ) in addition to extra info e.g. PbFree

SchrodingersGat:
Thanks for the contribution

PR #1080, opened 2017-03-01
silabs: Create CP2102N-A01-GQFN24 symbol

bradfa:
CP2102N family are USB to UART bridges in various QFN packages.  This is just the 24 pin variant.


SchrodingersGat:
Great work :)

PR #1081, opened 2017-03-01
Major fixes to linear library

evanshultz:
There were a number of things wrong with this library so I tried to fix them as best as possible. I understand this is a large number of changes, but many of them are quite minor. Please take a look.

Except for EC03, which is often bogus, and a few notes below, all symbols pass KLC.

Changes:
- Add MC33078, MC33079, MC33179
- Add NE5534
- Correct footprint info on OPA376IDCK
- Add NC pin 2 to MAX472
- Add keywords to MAX471/472
- MAX4081F power pins to 'W' type
- LTC6102* box outline width from 0 to 10mil
- Add missing NC pins to LTC6082
- Add datasheet links to LM311, LM319
- LM319: Add NC pins and correct pin 8 name to Gnd2
- LM311: GND pin to type 'W'
- LM2903: One set of power pins for all units
- Added LM3600 to DCM but without a datasheet because I don't find an "official" one. TI seems to have kept only LM13700.
- Add missing pins and datasheet to LH0002H
- LF398: Add missing pin 2 and update documentation
- Add missing NC pins to CA3080
- AD8213: Make one set of power pins for all units
- 1408: Place pin 14 on 100mil grid and update documentation

Still to do:
- Review and fix the many lame datasheet links
- L0002H: expose E3/E4 pins

Questions:
- 1408: Should it be renamed to "MC1408", does it really go in this library, and can I use a substitute datasheet (TI DAC0808)?
- Can pin names be lowercase?


evanshultz:
Please comment on the below symbol for LH0002H:
image

I added the E3 and E5 pins. Between all the pins and their names, I can't find anything I really like. This isn't terrible, but I hope somebody can find an even better arrangement.

Thanks!

evanshultz:
I don't find a datasheet for LM13600 from an official source, which at this point is TI. What is the requirement for datasheet links? Must they come from the component vendor?

yhlien:
I quickly checked LM13700 datasheet and found
""The output buffers of the LM13700 differe from those of the LM13600 in that their input biass currents....". The pins arrangement of these two are identical as well. So  it might be a good idea to set LM13600 as an alias of LM13700?

evanshultz:
@yhlien Thanks for the reply. Good point! Yes, they can be an ALIAS and that would make the library smaller. There are probably other cases where this can happen and I'll take a look thanks to your comment.

Still not sure about an "unofficial" datasheet...

yhlien:
As long as it is an alias of LM13700, its datasheet would be directly referred to LM13700... Actually I am quite surprised by the fact that TI does not mark LM13600 as obsolete and suggest LM13700 as a replacement. They simply make LM13600 disappear...

evanshultz:
@yhlien That's not quite how an ALIAS in KiCad works. An alias shares the same schematic symbol, but the DCM file for the library will still have a separate entry for each part. For example, LM393 and MC33078 could share a schematic symbol but their datasheets and functions aren't quite the same.

It _is_ weird how 13600 disappeared. 13700 is clearly better and a guaranteed replacement, but I was quite surprised too...

SchrodingersGat:
I've disabled EC03 checking, let's see how that goes :)

SchrodingersGat:
Here is my suggestion for LH0002H

capture

SchrodingersGat:
Or something like this:

capture

SchrodingersGat:
### LM2902N
* Description actually contains datasheet

### CA3080
* Please add default footprint as it looks like there is only one option 

### LTC6082
* Fix footprint filters

### AD8213
* Add default footprint
* Fix FP filter

### LF398
* Add FP filters

### 1408
* Please move this and similar to adc_dac.lib
* Rename to MC1408
* Replacement datasheet is OK if you can't find the right one

evanshultz:
I should point out that I'm working on a lot of changes here. I'll be sure to point out when I think it's ready for a final review before merging. After the KLC violations I went through the DCM file and now on to the LIB file. Incremental feedback and review would be great.

Thanks for the suggestions @SchrodingersGat! Here's what I've gotten done so far:
- LH0002H: I believe the power rail pins need to be separate. I see resistors between them for a current-limiting function at http://www.calogic.net/pdf/LH0002_Datasheet_Rev_A.pdf and http://www.ti.com/lit/an/snoa725a/snoa725a.pdf. Your second example is like mine but with the E pins spread out; I'm OK with that.
- LM2902: fix description
- CA3080: I see DIP and SOIC in the datasheet so I updated the FPFilter accordingly
- LTC6082: Modify to only the SSOP version; the DFN package wouldn't have contacted the thermal pad
- Update F3 and FPFilters on all AD* symbols

Other changes:
- Add AD86x0, TL062, and OPA213x
- AD813: Remove "1" text in the middle of the body

Additional questions above and beyond what I asked earlier:
- The ADCMP35x symbols don't have a pinout that matches either SOT-143 footprint. In this case I understand we don't create multiple footprints but I should change the pins on the symbol from the datasheet to fit our existing footprints, right? (KLC 7.6)
- Several parts specify a generic and package-specific version, like `LM324` and `LM324N`. In other cases there is a generic and then all the package-specific variants, like for `LM319`. Because we choose the package in CvPcb, I vote for only one of two options:
a. Only a single generic part if the pinout is common
b. Having only package-specific versions (no generic) if there are different pinouts
- There are often version of a symbol with different temp ratings, like LM118, LM218, and LM318. I'd find it acceptable to just specify the commercial/common version (the one starting with '3') in KiCad and letting the user manually change the number if they want a higher-temp version. Thoughts on that?

@SchrodingersGat @jkriege2 @poeschlr There are a few things about that above that potentially touch on KLC 2.0 rules. I appreciate if you could take a look and comment. Of course, anyone else is welcome, too. Thanks!

SchrodingersGat:
* ADCMP35x - Not sure what you mean here? These symbols seem to match the SC70-4 footprint OK?
* If there are different pinouts then there should NOT be a generic symbol (generics should only exist for when there are multiple footprint options)
* This issue (e.g. different temp ratings) is why I am hesitant to use a suffix such as _SOIC to name symbols. We leave ourselves no room to specify different unique parts e.g. Automotive / Industrial variants. This is why manufacturers have part codes, but again, they often contain extra information that we don't need... This requires some more thought.

evanshultz:
Continuing discussion with @SchrodingersGat:
-Update LH0002H symbol:
image
-1408: Renamed and saw that symbol pinout doesn't match any datasheet I can find online and has no pin 16. Since SOIC and DIP pinouts are different, will wait to figure out how we name them and then move.
- LF398: Similar to above in that there are different packages with different pinouts. Need to resolve that before updating this part.
- ADCMP35x: The physical part matches the SOT-143R_Reverse* footprints, but the pin numbering is different. The footprint's big pin is 1 but the datasheet's big pin is 2. KLC 7.6 would instruct me to change the symbol pinout to allow the existing footprint to be used, right? (I find it odd that the pin increment direction was changed to clockwise on this 'reverse' footprint instead of keeping them counter-clockwise and making pin 2 the big pin.)
-Temp ratings: I'm not clear on the problem. Does KiCad need to carry every variant? I would think only form-compatibility is required, which means we need to specify package. But for function compatibility, like temp rating or automotive qualification, the user could change the PN on the schematic after placing the 'base' part we keep in the library. 'AD7533*' in adc-dac.lib is an example of using the manufacturer suffixes.

New changes:
- Rename '4558' to 'RC4558' and '1408' to 'MC1408'

evanshultz:
ADCMP35x: I was wrong. It's actually a 4-pin SC-70 but [with one fat pin](http://www.analog.com/en/design-center/packaging-quality-symbols-footprints/package-index/sot-small-outline-transistor/sc-70.html). A smaller SOT-143. We already have a normal SC-70 with 4 small pins, so what we need is a naming convention for this case and I'll make the footprint.
Another example of symbol/package conflicts: LM311N is a SOIC package from National, but the FPFilter has SO, DIP, and TSSOP packages. And there are other LM311 packages which aren't listed there but would require a different symbol. I'll move this to KLC 2.0 thread...

More questions:
- Why can't comparators and opamps share a symbol? They look the same.
- Should we put a generic single/dual/quad opamp symbol in device.lib?

SchrodingersGat:
I'm for a generic single/dual/quad amp in device.lib - except that the pinout can change from device to device!

evanshultz:
Changes:
- Add LM4562/LT1016/LT1116/TL3116
- Fix FPFilters and default footprints throughout
- Renamed "OP077" to "OP77" and "ADA4075" to "ADA4075-2", consistent with the datasheets
- Move LH0002H pins to be on grid
image

To-do:
- Combine symbols (using ALIAS) where possible

To-do in separate PRs:
- Add TO-100-10 package for LM319H
- Add LFCSP-8-1EP_2x3mm_Pitch0.5mm for AD8218
- Micro8 (Case 846A) for MC33178
- Add UQFN-10_1.4x1.8mm_Pitch0.4mm for NCS210/1/2/3/4/5
- Add LFCSP-8_2x2mm_Pitch0.5mm for ADA4075-2
- Add VSON-8-1EP_3x3mm_Pitch065mm for OPA2333
- Make generic symbols. There are common dual and quad pinouts that cover 95%+ of each of those parts. For single, there might be a few generic versions worth making (SO8, SO8 with balance pins, SOT-23-5/SC-70-5 and the like, etc.)

Questions:
- LTC6362 could have pin 9 "behind" pin 6 that's only used on the DFN package, but this would make the pin count incorrect for the MSOP version in CvPcb. I can't see a way around having two separate symbols, one for each package, in this case. Is that right?
- Any problem combining opamps and comparators? For example, LM393 and MC33078 have the same symbol but one is a comparator and one is an opamp. I don't see why they can't be an ALIAS of each other.

evanshultz:
TL3116 image:
image

Datasheets are in DCM.

SchrodingersGat:
> LTC6362 could have pin 9 "behind" pin 6 that's only used on the DFN package, but this would make the pin count incorrect for the MSOP version in CvPcb. I can't see a way around having two separate symbols, one for each package, in this case. Is that right?

Yes please have separate symbols. Use _DFN and _MSOP suffix

> Any problem combining opamps and comparators? For example, LM393 and MC33078 have the same symbol but one is a comparator and one is an opamp. I don't see why they can't be an ALIAS of each other.

Even though the symbol is the same, as the functionality is different I would suggest a new symbol rather than alias.

> Temp ratings (LM118, LM218, and LM318 all exist)

If they are ALIAS parts, probably fine to leave them there for now. I'm unwilling to spend too much time worrying about aliases until the new file format comes in, otherwise it's way too big a job.

evanshultz:
@SchrodingersGat You asked to me notified when I removed MC1408 from this lib when you brought it into th ADC/DAC lib. I just did it. This PR isn't quite ready yet, but getting closer.

evanshultz:
@SchrodingersGat @jkriege2 @poeschlr Can I request a moratorium on changes to the linear library while I'm working on this? I'm made many changes above and now there's a bunch of conflicts. I think it's in all of our best interest to finally get our libraries KLC-compliant, and if we can't "shut down" a library during a big update it will be much, much more difficult. https://github.com/KiCad/kicad-library/wiki/Status-of-the-libraries is totally broken in formatting (on my computer, at least) but would seem to be the logical place to capture this as long as we were all aware.

evanshultz:
The master has received PRs that broke my branch. I have made changes all over the this library, and doing it all again is extremely unpleasant. Any help or suggestions to fix this?

Even after copying the master into my branch and verifying the offending part is identical, line-by-line, the conflicts persist. Normally I'd expect if a diff turns up 0 results, then the conflicts should be resolved. But that's not the case. See LTC6362 in the lib file or this snip from Notepad++ (trust me on this):
image

The dcm file has examples where I've made corrections but for whatever reason they're now causing problems. In this case, my PR should win. See here:
image

evanshultz:
Or I just bail on this and submit a new PR for the same patch?

SchrodingersGat:
@evanshultz I agree that this is annoying - having multiple competing changes in the same file is difficult to mange. I would suggesting using an external diff tool - I use [diffmerge](https://sourcegear.com/diffmerge/) but there are others out there.

I will try to hold any merges to this file until you are done.

jkriege2:
Please fix merge conflicts!

Best,
JAN

evanshultz:
@jkriege2 See discussion above. I got hosed with merges to master while I was making these updates. I know about it, but am thinking making a new PR is the best way to go.

More changes:
- Removed MC1408 as it's been merged in the ADC/DAC library
- Import LMC6482 from #1156, then align description and correct datasheet URL
- ADCMP35* and MIC845*: made all versions as ALIASes
- CA3080* pin 5 moved to the top since that's the common position for biasing the pin
- Added LM556/LM6172/L272_DIP16
- Removed "QSC" text on LM4250
- Removed redundant parts with deprecated suffixes (such as LM358N and leaving LM358)
- Revamped LTC6362 to use named pins instead of separate text
- Split OPA2333 into SON and non-SON packages; correct SON footprint
- Split TSV912 into DFN and non-DFN packages
- Added asterisk at the end of all FPFilter strings
- Moved GND pins to bottom of AD8xxx current sensors (some power pins still not at top)
- Moved MAX471/472 power pins to be on top and bottom; made 471 GND pin visible
- Corrected all KLC violations (which doesn't catch the AD8xxx power pins not at top)

And more questions:
- Should escaping the hyphen in a FPFilter prefer "?" or "*"? Does it matter?
- AD620/LF398/LM311/LM318/MAX423* have overlapping text; do we accept this? (I say no and would like to fix it.)
- All 555 are ALIAS but LM555N and NE555 don't show VCC or GND pins. Any ideas why?
image

I probably need another pass to catch other issues and be ready to merge (in a new PR due to the cruftiness of this one). I suspect there are a number of parts which have unique footprints that don't match the symbol which I haven't discovered yet, but let that not stop 95% of the updates from being made. And I haven't split out power pins on multi-unit parts and would like to do that later is everyone else is OK with waiting a bit. Feedback is greatly appreciated! @jkriege2 @SchrodingersGat @poeschlr

evanshultz:
@jkriege2 @SchrodingersGat @poeschlr Can I get review and comments? Currently this will make linear.lib KLC-compliant, which (I think) is a big milestone. I'd like to get this fixed up and merged before other PRs hose me again. Thanks!

jkriege2:
Hi!

some comments:
* [ ] LM311 Should we add the output transistor explicitly to the symbol (I think that's absolutely not obvious) as the DS does it: 
2017-04-27 07_19_56-lm311 pdf - adobe acrobat reader dc
* [ ] LM319: I added a PR for that TO-100 package: https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/22
* [ ] DS-link for LF398 does not work ... it should be http://www.ti.com/lit/ds/symlink/lf398-n.pdf
* [ ] LF398: could you add variants (ALIASES) for LF198 and LF298 (see DS above)
* [ ] LM2902: remove `SSOP*` and `MSOP*` from FPFilters, or add more description to these FPFilters (dimensions)

That's what I spotted for now ;-)
cu
JAN

evanshultz:
@jkriege2 Thanks for the review!

- LM311: I don't like this at all. Having the pins present is, for me, quite enough. And we aren't doing this on other symbols that have separate ground pins. I vote no.
- LF398 datasheet: Fixed, and thanks for catching that.
- LF398: I mentioned this here under section 4. I don't think temp grades should not be listed separately since it clutters up the library without adding a unique part (only a grade/rating/version of the same part). But this is an open issue.
- LM2902: Corrected. Thanks again!

Please keep the comments coming!

jkriege2:
+ Why not stick to what the datasheet does?

evanshultz:
LM311
Well, then we're deadlocked. @SchrodingersGat @poeschlr Transistor or no?

It seems you're talking about TI's datasheet (http://www.ti.com/lit/ds/symlink/lm111.pdf), where the output transistor is prominently shown. Note that this is the datasheet linked by KiCad.

Here are vendors of this generic part that *don't* show the transistor external anywhere:
- ST (http://www.st.com/content/ccc/resource/technical/document/datasheet/2d/22/50/0e/7e/39/43/64/CD00001072.pdf/files/CD00001072.pdf/jcr:content/translations/en.CD00001072.pdf)
- Fairchild (https://www.fairchildsemi.com/datasheets/LM/LM311.pdf)
- On Semi (https://www.onsemi.com/pub/Collateral/LM211-D.PDF)
- National (http://physics.gac.edu/~huber/classes/phy270/SpecSheets/LM311_National.pdf)
- Linear Tech (http://cds.linear.com/docs/en/datasheet/lt0111.pdf)

And National is the originator of this part (at least it uses their "LM" prefix).

But back to the transistor, I've never seen it drawn in any schematic I've ever seen. And I've seen a lot. LM311 is a very basic part that any moderately experienced user should know about it. I've never met another analog engineer who didn't know 311, 319, 324, 339, etc. because they're everywhere.

LM319 is very similar in this respect, but it doesn't have an external transistor shown. I guess maybe you're bothered because you saw the datasheet, Jan, and thought it would be helpful?? (I don't know.) I'm predominantly an analog engineer and I don't bat an eye because just a triangle is the standard way to represent a 311 and 319. And whoever originally created those symbols also understood just a triangle to be adequate and/or correct.

And another reason: Schematically, a LM393 or LM319 or even an opamp may be a substitute for a LM311 with minor changes (say you originally need one gate but then found you needed two). Now it's much more complicated to add because the symbols are dramatically different.

I'm digging in my heels on this because it's so atypical in my experience and because I just personally don't like it. Do we have other opinions and considerations to make this decision?

jkriege2:
I'm a physicist (so not a professional electronics guy ;-) ... used that thing a few years ago ... generally I find it better if such things are stated explicitly in the schematic. If I see a triangle with one output at the tip, I assume a standard OPAMP and would not expect a strange output stage ... Especially for the LM311 you can use that transistor to drive an LED (E to GND and LED with R from Vcc to C of the out-Transistor) ... seeing that in a schematic I would really wonder ;-) ... 

Maybe my view is closer to what a hobbyist would expect ...

But that's just my opinion ... let's see what the others say

evanshultz:
@SchrodingersGat @jkriege2 @poeschlr 
Any more feedback on this? Because of later PRs I'm now out of date, but if you don't find other issues with the changes above I'll create a PR which incorporates later changes to this lib and we can merge this.

jkriege2:
Found some more:
* [ ] `LTC6102xDD` and `LTC6102xMS8` should also use the `_MSOP`- or `_DFN`-nomenclature ... Also: Waht does the `x` in the name stand for + there are variants `-1` for no obvious reason ...
* [ ] `MCP601xyz`, `MCP603xyz` do not use the `_SOIC`, `_DIP`... nomenclature
* [ ] `MCP6566` do not use the `_SOIC`, `_DIP`... nomenclature
* [ ] `LTC6102...` should/could use power-pin stacks to make symbols smaller
* [ ] `L272_DIP16`power-pin stack for V- is not correct (1 visible PWR-IN + all others PASSIVE)
* [ ] maybe invisible NC-pins of OPAMPS should be move a bit closer to body, as done for  this one:
2017-06-05 19_34_04-choose component 206 items loaded This will prevent (a bit better) an unwanted connection to an invisible NC-pin in the schematic editor.

... Plus some violations from scripts:
```
Checking symbol 'LM311':
  Violating Rule 4.6 - Pin types
    Pin electrical type should match pin function
    Power pins should be of type POWER INPUT or POWER OUTPUT
     - Pin GND (1) @ (0,-300) is of type OPEN EMITTER
Checking symbol 'LM319_TO':
    Datasheet field must be EMPTY
```

Otherwise. I don't see anything more ;-)

Best and keep up the great work!

JAN

evanshultz:
Thanks Jan! The pin named GND on LM311 really is an open emitter. It's typically connected to ground and always called that on all datasheets I've ever seen, though. So I think that one is good to go.

I'll look into and address the other issues you caught shortly. Thanks again!

jkriege2:
LM311: OK ...

Since you mentioned LT311/LT111 above ... maybe add those as ALIAS, although they are obsolete?
Also maybe add the newer version LT1011 (http://www.linear.com/product/LT1011)

JAN

evanshultz:
Hi Jan,

Thanks again for the feedback!

Changes:
- Yeah, LTC6102 parts are confusing. The "x" is for the temp grade and whoever created them originally used the LT package names. The "-1" is a valid part of the symbol name as this version adds an enable pin. However, there are no "HV-1" versions. Take a look at the datasheet. I think it's all good now.
- I also think most MCP* parts (see below) are now OK. I renamed them where required and added MCP6002_DFN and MCP6L92/MCP6L94.
- Several of the single-opamp MCP* parts have a SOT-23 package in different pinouts. How should we capture that? Perhaps "_SOT23-1", "_SOT23-2", etc. is most scalable? Unfortunately, that doesn't help the  user select the pinout and its ugly, but the symbol preview will show it. *Any suggestions?*
- L272_DIP16 now has all invisible pins set to PASSIVE type.
- I didn't mention any other LT parts, but why not add them? I included LT1011.

I'll move NC pins on opamps when I create a new PR.

evanshultz:
@jkriege2 @SchrodingersGat 
Do you guys have comments on the item above? How should we name symbols when there are parts which come in the same package but have different pinouts?

SchrodingersGat:
Can't you differentiate using the manufacturer P/N?

evanshultz:
@SchrodingersGat @jkriege2 
Indeed we could, but I thought the whole point of the `_SOIC` type of suffix was to eliminate the manufacturer suffix. I recall this discussion during the KLC 2.0 update. Since the manufacturer suffix can vary for the same package, it doesn't scale. I was suggesting a scheme that would work regardless of manufacturer. But perhaps you were thinking in this weird, one-off case the manufacturer suffix is OK??

SchrodingersGat:
If there are two variants of the symbol in the same package but with different pinout, the only way of differentiating these from the datasheet is the manufacturer part numbers.

A user will pick which variant they want to use based on the data in the datasheet. Any other way of naming the symbols would just add extra noise. 

Personally, I think the whole thing is easier when the manufacturer PN is used across the board, but this would require the KLC to be updated.

> Perhaps "_SOT23-1", "_SOT23-2", etc. is most scalable?

But how would a user know which version they want? They would have cross-reference this back to the naming convention in the datasheet.

Thoughts? :)

evanshultz:
> If there are two variants of the symbol in the same package but with different pinout, the only way of differentiating these from the datasheet is the manufacturer part numbers.

Unless we have our own scheme for KiCad.

> But how would a user know which version they want?

That's why I mentioned above that the symbol preview will show the pinout.

This is a bit tough. We have two suggestions above: something sequential that I brought up and the manufacturer suffix. The first one is arbitrary and the second one is manufacturer-specific and breaks the philosophy we started with KLC 2.0. How should we proceed?

I looked at a few component libraries I could find online (EAGLE, Altium, EasyEDA, etc.) and they appear to all use manufacturer suffixes. Does anyone else know of any "open" libraries to provide gather more input?

In this case, if we amend KLC for cases where there are sole-source parts with different pinouts in the same package, I'm willing to use the manufacturer suffix. But it's annoying that this isn't a nice, universal solution. I pushed for using package suffixes instead of manufacturer suffixes, which I still favor, but I'm willing to be wrong and backtrack.

SchrodingersGat:
You're right it's a complex problem.

From my perspective, I would expect the naming of *unique* symbols to use the manufacturer naming convention. Introducing a secondary convention that is mapped to the manufacturer convention is unnecessarily complex, IMO.

Generic symbols should specify the package as human-readable text as they are not tied to a specific manufacturer part. e.g. `Opamp_Dual_SOIC-8`

I think that we should be moving towards the following convention:

For manufacturer specific parts:

1. Symbols use at minimum the base MPN 
1. For symbols with only one matching footprint, this is enough to identify the device
1. For symbols with multiple footprints, each footprint must have its own symbol, and the footprint is designated using the appropriate suffix from the extended MPN code
1. The footprint name e.g. `SOT-23-5` should be appended to the footprint description. This means the user can search on the footprint name and find the correct symbol.
1. If there are multiple variants of a symbol (e.g. different voltage ratings) then the following options are available:
   1. Single *generic *symbol that does not specify voltage e.g. `MCP809-xxTT`
   1. Each available variant is added as a specific ALIAS (in this case there should *not* be a generic symbol)

This may be a large enough change that it's rolled into the KLC 2.1 discussion.

We also need to work out whether we group devices based on function or manufacturer.

Thoughts? @evanshultz @jkriege2 @poeschlr

jkriege2:
@SchrodingersGat: sounds good to me!
JAN

SchrodingersGat:
I now realize that the KLC as it currently stands is worded in completely the opposite fashion. I don't think this is what I intended.

If there are no objections I will update the KLC to fix this.

evanshultz:
@SchrodingersGat and others...

First, let me say that I pushed for the `_SOIC` nomenclature. It seemed good at the time, and I still like it for a number of reasons, but I've realized 2 things:
1. EAGLE and other hobby CAD tools, from what I see, use manufacturer suffixes.
2. There is only a relatively small number of parts where this matters (semiconductors which don't have exact crosses like 1N4148, BC846B, connectors, etc.). For most parts in the KiCad library, the part either is unique or crosses from a different vendor are truly crosses and we needn't care about package.

So I may have proposed something that ultimately wasn't the best for KiCad users. Have I seen the light?

If so, perhaps going back to manufacturer suffix in those cases is best? (Sorry again.) ALIAS could be used in the cases were different vendors have the same part with a different suffix (such as [On Semi MC33078P](https://www.onsemi.com/pub/Collateral/MC33078-D.PDF) and [ST MC33078N](http://www.st.com/content/ccc/resource/technical/document/datasheet/group0/2d/7d/0e/f1/e8/03/42/4c/CD00000474/files/CD00000474.pdf/jcr:content/translations/en.CD00000474.pdf) both being DIP8 versions of the same dual opamp).

OK, now to your point above. I like the concept, and that is what I had in mind even if KLC doesn't capture it as we thought. Could you perhaps give a few examples to illustrate everything (this may be able to drop into KLC once agreed upon).

I'm going to ask a few questions which may be dumb but this is (hopefully!) for absolute clarity:
- Is the above, which may have been your preference, Oliver, a better way to go? If so, I'll make the changes to at least this library and try to fix anywhere else I see it in the library.
- Would item 3 still allow SOIC and DIP packages to share a symbol? Perhaps you mean only when the package forces a different number of pins (adding thermal pad, for example).
- Perhaps we should decide on a method for item 5 instead of leaving it open. This way the library is more consistent. Would it be easier for users if all parts were listed as ALIASes? I think I favor that approach.
- How should we handle manufacturer PNs where there are extra characters between the base PN and the package-specific suffix. One example is [Linear Tech LTC6102](http://cds.linear.com/docs/en/datasheet/6102fe.pdf) where the temp range separates them. Should that character be ignored? Be replaced with `x`? Something else?

Grouping is a big topic as well. I started the conversation (though it was derailed) at https://github.com/KiCad/kicad-library/issues/1082.

SchrodingersGat:
> Is the above, which may have been your preference, Oliver, a better way to go? If so, I'll make the changes to at least this library and try to fix anywhere else I see it in the library.

I think this approach is more professional and is going to be better in the long run. If we don't do it now, we'll have more work when we decide to change it later :)

> Would item 3 still allow SOIC and DIP packages to share a symbol? Perhaps you mean only when the package forces a different number of pins (adding thermal pad, for example).

It is still *allowed* but I would strongly prefer moving to a more atomic approach where each symbol is directly  associated with its correct footprint. We won't make the push to change this in the current symbols until the new library format, I would imagine, as that will (hopefully) solve a lot of these issues anyhow.

> Perhaps we should decide on a method for item 5 instead of leaving it open. This way the library is more consistent. Would it be easier for users if all parts were listed as ALIASes? I think I favor that approach.

You're right, consistency is best. I much prefer each symbol being added as an alias, rather than generic parts. However, there is a limit to this. I don't think for e.g. we should allow each packaging variant for a symbol, or PbFree variants, etc. That's a lot of clutter. we should also consider whether temperature variants (etc) are included, or simply functional differences.

> How should we handle manufacturer PNs where there are extra characters between the base PN and the package-specific suffix. One example is Linear Tech LTC6102 where the temp range separates them. Should that character be ignored? Be replaced with x? Something else?

I think that `x` as a separator is a reasonably widely accepted standard.

SchrodingersGat:
I have updated the [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) and the [FAQ](https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions) - please let me know if you see any obvious errors.

evanshultz:
@SchrodingersGat Finally got around to reviewing your changes. This sounds great to me. I have three thoughts:

1.  FAQ: I think more examples, especially covering the unique cases above, would be helpful. Those cases are: where the an otherwise identical symbol can have different pinouts in the same package, when there are multiple MPNs for ostensibly the same part.
1. FAQ: To leave out extraneous info in a MPN, can character omission be used? So instead of `LTC4357xMS8` could it be `LTC4357MS8`?
1. KLC: Sections 3.3 and 3.7 look awfully similar to me. Are they redundant or can they be combined?

I'd like to add the extra examples to the FAQ if you don't mind. I can use the MC33078 example above and then MCP6566.

SchrodingersGat:
1. Yes more examples would be a good thing

2. I think it is helpful to show that a part is a stand-in for multiple symbols? Perhaps not `x` or `-`, I'm not sure what works best in the general case.

3. You're right they are duplicated. I think they can be combined

suzizecat:
For the filling characters, IMHO,  the `x` is the best solution (in exemple, ST uses them) since it's neither often used in MPN, nor used to add informations as the `-` may be used.

image

evanshultz:
More examples added to the FAQ. @SchrodingersGat, I'll let you edit KLC to combine 3.3 and 3.7.

I've updated this PR to cover the MCP symbols, but there's a problem. First, they're horrendously ugly simply because of their naming scheme. For example, `MCP6002-x/MC` is the name of the DFN MCP6002 symbol. `x` is for the temp range and the forward slash is actually used in the orderable part number. But a forward slash isn't allowed (KLC 1.7) so that doesn't work. Perhaps we leave out illegal characters if they appear in MPNs? @SchrodingersGat @poeschlr @jkriege2

evanshultz:
Update the following to the MPN style of symbol naming:
- MCP* (mentioned above and added where needed)
- LF398
- LM319
- LTC6082
- LTC6102
- OPAx333
- OPAx340
- OPAx376
- TSV911

Add:
- MCP6L91 in 8-pin packages
- LF398H
- Missing LTC6102
- Missing OPAx736
- Missing TSV91x

I also combined a number of symbols (for example, LMV321 and MCP6001U and OPA333xxDCK).

Next: Add in the parts that have been pulled into the linear lib since this odyssey started. And split out the power symbols to conform with the current KLC. I'd be happy to have a review of the lib as it stands.

evanshultz:
Changes:
- Moved NC pins closer to part bodies
- Add SHDN pins to LTC6081xDD
- Fix overlapping LM318 pin names
- Split LM318 into variants by package
- Combine more symbols where possible
- Add NJR opamps
- Minor other corrections

evanshultz:
Merge/update all changes since I forked:
- [Update LTC6362](https://github.com/KiCad/kicad-library/commit/783d36ea4ee71a603376f3408ef07e8a5d7263e1#diff-5cc792ff4ef191f02499f03eb6260f74) / [Remove old one](https://github.com/KiCad/kicad-library/commit/709ca35dc88c741c5f7c586272c2f93ce2418504#diff-5cc792ff4ef191f02499f03eb6260f74)
- [LMC6482 already added](https://github.com/KiCad/kicad-library/commit/fc0d5e0e21de3d1fd0a2809ae52ae9170f646bd7#diff-5cc792ff4ef191f02499f03eb6260f74)
- [Added ADA4807-1](https://github.com/KiCad/kicad-library/commit/3d3bbeeb3ca13b4244d9df706cd701b1ad183be9#diff-5cc792ff4ef191f02499f03eb6260f74)/ [Updated](https://github.com/KiCad/kicad-library/commit/9b3b00f7df2ed8303866143306672fb055a59e79#diff-5cc792ff4ef191f02499f03eb6260f74) (standardized DCM entry; updated FPfilters)
- Went ahead and added all ADA4807-x parts
- [Add THS4631D](DCM: Add THS4631 family)
- And then all the other THS4631 variants
- [Add OPA2695 in SOIC](https://github.com/KiCad/kicad-library/commit/7a5c48169d01be5bc938d6f650b926bdb340fdc6#diff-5cc792ff4ef191f02499f03eb6260f74) (updated name to use MPN and standardize)
- And, for fun, OPA2695 in QFN

evanshultz:
I need to revert LTC6362 to my old symbol style and then update it. I had it fixed better.

But there are two issues I need help with:

1. There is an illegal character (`/`) in several Microchip parts. How to handle that?
1. It appears that the Analog Devices parts with a hyphen (`-`) in the part name aren't being found. I will play with it some more, but does anybody know how to fix this or have an idea how we should name these parts in KiCad?

I plan to fix up LTC6362 tomorrow and then I'd like to try and merge this. Split the power pins into a separate symbol right after merging this big body or work. Does anybody have a problem with that?
@SchrodingersGat @jkriege2 @poeschlr

SchrodingersGat:
1. Simply remove the `/` character e.g. `I/PT` becomes `IPT`. 
1. What do you mean by "aren't being found"?

evanshultz:
1. OK. Done.
2. I found my error.

LT6362 has been reverted to the style I had originally, and then split into two variants, one for each package type. The PR didn't match the style I'm using with respect to pin orientation/placement, showing pin names instead of "dummy" text for pin names, etc.

Next I'm going to review if any of the ADI parts can be combined instead of all being discrete symbols.

Does anyone object to pulling this without splitting out the power supply pins?

evanshultz:
LTC6362 images:
image

evanshultz:
OK. I think this might, just maybe, be done. One KLC error is that LM311's GND pin is actually an open emitter, and so I set the type as such, but KLC is flagging it because of the pin name. No problem (and see discussion above).

I'd like to submit this to be pulled and then do separate power pins later. Any objections? Can you all (or at least someone) review this and all set a moratorium on merges to the linear lib until this is done?

jkriege2:
sounds good to me ... unfortunately I won't have time for a closer review within the next 2 weeks, but If you kept up the good work (see my other reviews on this PR), I'm quite sure this should go through smoothely ;-)

Best,
JAN

evanshultz:
Changes:
-Add package names to all descriptions in DCM
-[Add LTC1100](https://github.com/KiCad/kicad-library/pull/1346#issuecomment-313881360) and modify
-Add: TLC277/TLC279/RC4560/RC4580/SA5532/LM2904/AD797/OP249/LT1012
-Split symbols due to different (incompatible) packages: TLC555/LM6161 (no 10-pin footprints are in the lib)/LMC555
-Added DFN package for LM358 (ST only)
-Removed extraneous NCS21x parts after confirming with with On Semi that only NCS210, NCV210, and NCS210 were ever released (and then not in UQFN package)

evanshultz:
Changes:
- LM6171
- LM6321
- Eliminated high-temp versions of LM6361, which included all 10-pin versions

@SchrodingersGat @jkriege2
I know you guys are busy with the reorganization and other things, but I'd really like to know if we've sorted out the naming scheme. I wrote a reply at https://github.com/KiCad/kicad-library/issues/1398#issuecomment-314017031 which I hope sums up what needs to be done. If that's right I will need to make changes before this gets merged.

evanshultz:
I realized there were a number of outstanding PRs that might fall into the linear library, so I searching around to incorporate them. Surprisingly, I found only one: [OPA2691](https://github.com/KiCad/kicad-library/pull/1282).

evanshultz:
Made updates according to the latest KLC, which uses no suffix for generic parts with common pinouts, full MPN for unique parts (multiple symbols if there are different packages or pinouts), and package name suffix (`_SOIC`, for example) when there are multiple vendors with different package names for the same part.

Changes:
- LF398: TI in TO can, TI and LT in DIP8, TI in SOIC14, LT in SOIC8
- LM319: many vendors for SOIC14/DIP14, only TI in oddball TO can

I also found a duplicate DCM entry for LM733H.

@SchrodingersGat @jkriege2 @poeschlr 
This is all the known changes I'm aware of for this library. I'll leave this PR idle for a couple days to gather comments, then probably start a clean PR to merge this major update. If there are any known or suspected gotchas when basically blowing out a whole library and starting over, please let me know.

evanshultz:
I realized that there might be other parts in other libraries that belong in `linear.lib`. After searching by hand and for "op amp", "opamp", "amp", "comp", "current", and maybe some other terms, I found:
- `analog_devices.lib`: Moved AD623 and several instrumentation amps
- `sensors.lib`: AD8418 was duplicated here, moved INAxxx parts, moved ZXCT* parts

evanshultz:
With #1437 merged, closing this. Phew!

SchrodingersGat:
What a saga! Thanks for your fine work :)

PR #1085, opened 2017-03-03
Fixed many instances of duplicated pins

SchrodingersGat:
- Found with new KLC scripts (Rule 3.3)


SchrodingersGat:
I have touched a lot of components here so a lot of Travis noise as they are non-compliant to a lot of rules. Look at the diff to see what has really changed :)

PR #1097, opened 2017-03-11
Update ftdi.lib

mcvanili:
adding FT2232D


jkriege2:
Hi and thanks for the contribution.

Please note that we have a set of rules that symbols for this repository have to follow. They are acaaled the KLC and can be found here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention
Currently we are in the process of an overhaul of these rules and the new version is currently available here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC Please stick to the latter one! There is also a set of Python scripts that check some of these rules: https://github.com/KiCad/kicad-library-utils

These scripts are also ru automatically on every new submission. You can find the results here:
2017-03-08 22_03_01-created footprint library housings_lcc by yo3iti pull request 1 kicad_housi

About your component, the major issues are:
1. Please add description, keywords and datasheet link (these are stored in the accompagnying DCM-file)
2. All pins must lie on a 100mil grid, but many of your pins are on a 50mil grid. Please move them
3. All text fields should have a size of 50mil (also the hidden ones!)
4. pin length of 200mil should be enough for your symbol
5. The box should ahve a line-width of 0.01" and should be filled with "background", i.e. "yellow"
6. Please specify a footprint in the footprint field, if there is only one possibilitiy and a set of FPFilters that cover all possibilities as described in the KLC
7. please move the REFDES and label to positions like with other symbols in the FTDI lib.

Best,
JAN

PR #1098, opened 2017-03-11
Update ftdi.dcm

mcvanili:


jkriege2:
Hi!

This should be part of you other PR https://github.com/KiCad/kicad-library/pull/1097. We need to review these two files together!

Best,
JAN

PR #1099, opened 2017-03-11
adding FT2232D

mcvanili:


jkriege2:
Hi!

very nice work now ... only a few minor thing to change:
1. Footprint should be the full name, i.e. `Housings_QFP:LQFP-48_7x7mm_Pitch0.5mm`
2. FPFilter should be accordingly `LQFP*7x7mm*Pitch0.5mm*`
3. DS-link should be directly to the PDF: `http://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT2232D.pdf`

If these are fixed, I will merge1

Best,
JAN

mcvanili:
Hi,

thank you for your patience. I'm new to github, so it took a while to understand how it works. 

I've already added modifications you mentioned, but now this component is a bit different from all the others in ftdi.lib because of this changes (e.g. other componens have link to the IC description page, but not datasheet). 

Should it be like this?

jkriege2:
yes, usually we link the datsheet directly (at least in all other libs), so I would stick to it here too.

Thanks for the work you put in!

JAN

PR #1103, opened 2017-03-13
maxim: Add symbol for MAX7325AEG+

joe-skb7:
MAX7325AEG+ is an I/O port expander, in QSOP-24 package.

Signed-off-by: Sam Protsenko <joe.skb7@gmail.com>


jkriege2:
Hi!

thanks for the contribution. I checked your symbol and found only one minor thing to correct: The FPFilter should be `SSOP*3.9x8.7mm*Pitch0.635mm*`. Please change that so I can merge this symbol.

Also note: We are currently working on an updated version of our KLCs: https://github.com/SchrodingersGat/kicad-library/wiki/KLC ... Those also details the rules for FPFilters a bit more ;-)

best,
JAN

joe-skb7:
Hi @jkriege2,

Actually MAX7325AEG+ has QSOP-24 package, not SSOP-24 one. And I believe it's not completely the same as SSOP-24. I'm aware that we don't have a footprint for QSOP-24 yet, but I intend to add it, as a next step.

Or you mean I just need to change the mask to `QSOP*3.9x8.7mm*Pitch0.635mm*`?

Thank you.

jkriege2:
Hi!

hmmm ... then I would suggest adding the footprint-PR and we can review them together (just mention this PR there). Then we can get everything right with one shot ;-) and the link will work!

Best,
JAN

joe-skb7:
Ok, I see your point. Footprint should be added before the symbol, so it can be referenced safely. I'm going to use footprint from here: http://ww1.microchip.com/downloads/en/PackagingSpec/00000049CH.pdf (because existing footprint for QSOP-16 was done using the same spec). Will try to do it till the end of the week. Then, hopefully, we can merge this symbol.

joe-skb7:
So, I created pull request for QSOP-24 footprint, [here](https://github.com/KiCad/Housings_SSOP.pretty/pull/31). Seems like the checker script gives false positive error (I mentioned that in my comment in that pull request). Anyway, once it's got merged, we can get back to this patch here.

joe-skb7:
@jkriege2 Ok, the QSOP-24 footprint was merged (I shared the link earlier). Is it ok now to pull the symbol?

Thanks.

jkriege2:
Hi!

one final request. please set the FPFilter to `QSOP*3.9x8.7*Pitch0.635mm*`. then  I can merge!

Best,
JAN!

joe-skb7:
@jkriege2 Done. Thanks for guidance ;)

jkriege2:
nice work
Thanks,
JAN

PR #1105, opened 2017-03-17
Add generic BCV61/62 to device library

evanshultz:
See PR #1047 for the details. I hosed that up so let's try again.


jkriege2:
Since this is a generic symbol, please remove FPFilter and footprint field.

Otherwise fine and I will merge!

evanshultz:
I removed the footprint. While SOT-143 is a valid footprint, I see that this limits the possible footprint selections in CvPcb. Which is bad. But the user can always remove the footprint filters if they like. And this is a pretty unique configuration of transistors which is unlikely to have loads of potential packages. But still, I get it. So remove all FPFilters?

The other dual transistor parts I added have a smattering or footprints in their FPFilters, so should I remove FPFilters on them as well?

jkriege2:
Hi!

I would say yes & yes (can be done in this PR)

@SchrodingersGat : What do you think?

JAN

SchrodingersGat:
Default footprint should be removed but I think the filters should remain. Remember the filters can be disabled in CvPcb if users don't want to filter.

evanshultz:
@jkriege2 Can this be merged now?

evanshultz:
@jkriege2 @SchrodingersGat Are we able to merge this now?

SchrodingersGat:
Looks good to me, thanks :)

PR #1106, opened 2017-03-17
Add BCV61/62

evanshultz:
See PR #1047 for details. I messed that up so let's try again.


jkriege2:
nice work!

Thanks,
JAN

PR #1110, opened 2017-03-19
New 7 segment symbols

Schlumpf:
Replace generic 7 segment symbols by new symbols and add KCSx02.
I moved the pins to top and bottom side, because with pins left and right, you can't place them next to each other to create multi digit displays.

image


SchrodingersGat:
@Schlumpf this is a great contribution, thanks!

Please fix the KLC issues (click on the "Details" button where it says "All checks have failed"). There are a few things that it has issues with.

Schlumpf:
I fixed a typo in footprint name, but that's not what travis complains.
I only touched the generic and the new KCSx02 parts. The generic symbols don't have a datasheet link, I think that is correct.
That is the only "error" of my symbols. The other "broken" symbols are not edited by me.

SchrodingersGat:
Ok, those other ones can be fixed up at some other time. Thanks for the contribution :)

PR #1111, opened 2017-03-19
Fix wrong pin number for EN.

bullestock:
The 4504 component has two pin 15. The one named EN should actually be pin 13.


poeschlr:
Would it also be possible for you to fix the klc violations of this symbol?
(Yes i know most of the symbols in this library violate the klc.)

At least correct the electrical types of the pins. (The power pins should have power in as electrical type)

Adding the documentation information would also be appreciated. (at least a description and keywords. I'm not sure about adding a datasheet because this is a general symbol for parts by multiple manufacturers. But maybe add the datasheet of the part you use as an example.)

Also put the pins on a 100 mil grid. Most of the pins would line up if you move the symbol 50 mil to the left. (then the center of the symbol would be on the y axis.)
If you additionally move the upper 4 pins 50 mils up then they would line up aswell and the center of the symbol would be at the origin.

Really good would be to edit all lines and give them a 10 mil line width. and to fill all rectangles and poly lines with fill background.

bullestock:
I have fixed some of the issues.

SchrodingersGat:
@bullestock based on the checks (see the "All checks have failed" text below - click on "Details"):

https://travis-ci.org/KiCad/kicad-library/builds/213595211

Please address the three errors described there

bullestock:
I'll look at it as time permits, but those violations were also there before I fixed the rather more serious error of the wrong pin number.

poeschlr:
I think most of the errors will go away if you move the top four pins (VCC, VSS, GND and EN) 50 mil towards the top.
Yes this also requires editing the poly line that describes the top outline.

Also just open the field definition of the symbol and edit the text size of the fields to be 50 mil and all errors reported by travis will vanish.
If you are done with that could you post a screenshot of the symbol aswell?

jkriege2:
Since this is a small fix for a small bug, I will merge without enforcing KLC! ... it's not really clear how we want to adress all the logic ICs anyways (powerpins!!!).

Best, JAN

PS: Nice catch, thanks for contributing the fix!

PR #1112, opened 2017-03-20
Added BQ29700 battery protection

kajusK:
Datasheet [1] says it's BQ2970, but BQ29700 is BQ2970 with factory programmed configuration. I guess it's better to create BQ29700 and make aliases when needed. Or would BQ297XX be better?

[1] - http://www.ti.com/lit/ds/symlink/bq2970.pdf


poeschlr:
> Or would BQ297XX be better

I would go with BQ297xy. This would match the notation used in the datasheet. (last line of the table on page 4 of the datasheet.)

Now to the symbol itself:
Please add pin 1 as type NC invisible
Symbol outline should be 0.01 inch

For details see new klc:
https://github.com/SchrodingersGat/kicad-library/wiki/KLC

kajusK:
Updated, the screenshot is here 
2017-03-22-120456_1366x768_scrot

kajusK:
Is it ok now? Or is there anything else I should change?

kajusK:
Hi @SchrodingersGat, I updated the part as requested:

2017-04-02-112115_1366x768_scrot

SchrodingersGat:
Great! Thanks for the contribution

PR #1113, opened 2017-03-20
adding ATAES132A

jeromebeguin:


poeschlr:
The symbol outline should be set to 10 mil (0.01 inch)
See new klc point 4.2.ii: https://github.com/SchrodingersGat/kicad-library/wiki/KLC
Also group the pins of spi/i2c together.

If i read the datasheet correctly the communication type is selected via the part number and is not user changeable.
Because of that it might make sense to have two different symbols. One for spi variants and one for i2c variants.

For everyone else searching for the pin definition: it is on page 67 of the datasheet.
http://www.atmel.com/Images/Atmel-8914-CryptoAuth-ATAES132A-Datasheet.pdf

Note: currently the symbol looks like this. @jeromebeguin please add a screenshot and a link to the datasheet in future pull request descriptions. This makes reviewing pull requests a lot easier. (for a datasheet like this it also helps if you give a hint where the relevant information is located.)
screenshot from 2017-03-20 23-54-34

jeromebeguin:
Hi Poeschlr, 

Thank you for your review. Sorry for the datasheet, I've put the link of the complete one but I used the [short one](http://www.atmel.com/Images/Atmel-8914S-CryptoAuth-ATAES132A-Datasheet-Summary.pdf) for creating the part.
Do you think it's really worth to have 2 schematics if one fits both ? 

Best

Jerome

poeschlr:
> Do you think it's really worth to have 2 schematics if one fits both ?

I think that with two different symbols the ERC can help a bit better.
Example: For spi types the si pin is a data input. For i2c types the sda pin is bidirectional. This means for spi devices erc could find problems if the user does not connect an output pin to this input pin. If you have it as a bidirectional pin drc can not detect that. (With one symbol you need to have it as bidirectional.)

Have a look at the long description on page 67 of the long datasheet. (It gives a bit more detail about the differences between spi and i2c)

@jkriege2 or @SchrodingersGat what do you guys think about this?

SchrodingersGat:
I would lean towards keeping it as a single symbol as it is only a single device (not two devices that have a compatible symbol).

jeromebeguin:
Could you merge if it's ok this way ?

jkriege2:
The symbol `ATAES132A` is fine ... but the PR also deletes some SAM µCs and stuff ... please fix those merge problems, before we can proceed.

Best,
JAN

jeromebeguin:
Follow on PR 1157

PR #1114, opened 2017-03-21
Add generic opamps to device library

evanshultz:
Images:
image

As discussed in PR #1081, at least @SchrodingersGat and I are in favor of generic opamps. The quad and dual seem straightforward, but the singles can be messy. I took a guess at what might make sense. Please consider what's best and think up pros and cons that I've probably missed. Feedback welcome!


jkriege2:
Hhmmmm ... some thoughts on this:
- especially with the SOT-23 I would be careful ... just for fun have a look at this DS: http://ww1.microchip.com/downloads/en/DeviceDoc/21314g.pdf ... there are 2 single-variants in SOT-23-5 and one in SOT-23-6
- 2x and 4x OPAMPS might be OK ... although there are exceptions (of course), e.g. http://www.st.com/content/ccc/resource/technical/document/datasheet/02/9c/bc/75/59/9c/47/bd/CD00000054.pdf/files/CD00000054.pdf/jcr:content/translations/en.CD00000054.pdf http://cds.linear.com/docs/en/datasheet/10134fe.pdf https://www.onsemi.com/pub/Collateral/LM339-D.PDF http://www.ti.com/lit/ds/symlink/rc4136.pdf http://www.ti.com/lit/ds/symlink/lm3900-n.pdf
- single OPAMPS: To me the naming is not clear ... what does "`NULL`" mean? I would go for a naming that has somehow all important pin-numbers in teh name (see e.g. the crystals, diodes or transistors in `device.lib`).

Generally thinking about whether I would use that or not: I would not use it in schematics (I think) ... usually I would select the OPAMP that I want (usually you want a very specific set of props, or it is so generic, that you can put in a 741/TL061/... to start with and then exchange later, especially when I just want to illustrate a schematic). What I would use those for is as a base to derive a new OPAMP ... then I only have to check once whether the pinout is correct and then just derive my OPAMP. Also all the OPAMPs should (!!!) have approx. the same symbol, so they can easily be changed for another type.

A bit off-topic: What about power-pins as separate subsymbols for OPAMPS? Can be handy when combining several OPAMPS ...

Just some thoughts on this,
JAN

evanshultz:
@jkriege2 Thanks for the comments!

- Yes, I was aware of SOT-23 differences, even of the same part. Not sure how to handle this, maybe just getting rid of this version altogether, but the pinout I chose is the most common in my experience.
- There are always caveats. However, in the LTC datasheet it even mentions the dual op amp differs from the standard pinout. I agree that the dual and quad are extremely common and make sense to add.
- `NULL` is to null the part, and is sometimes called "null" or "balance" or "compensation" (or some combination of those terms). The key point is having connected pins at 1-5, or 1-8, or 1-5-8. Do you have a better suggestion for symbol naming?

We can't cover all opamps. But to me, these are useful all the time. I will know the pinout for most generic opamps and I can use these, just like I use the NPN or inductor in the generic library. I can understand a mental divide because an opamp is more complex than a resistor or a single diode, but consider the objective of the library. This is a "one-size-fits-many" library full of non-specific parts. The user knows these are not a specific part and will need to consider using any part from device.lib. Here as well. For me it's reasonable to have many edge cases here than on a specific part.

Excellent point on the power pins! I was planning to do that as I near completion updating linear.lib, and it would apply here too. Perhaps we can settle other issues how to handle generic opamps and then save that for last, either by pulling it and I'll update power pins together or by holding off on pulling until the symbol is updated. Either way, keep this is concert with linear.lib. Thanks for the reminder!

evanshultz:
@jkriege2 @SchrodingersGat @poeschlr Thoughts about the above? How to handle generic opamps?

evanshultz:
@jkriege2 @SchrodingersGat @poeschlr What to do about the above? I'd like more input to make a decision.

SchrodingersGat:
Personally I think there are too many possible variants to make the generic op-amp symbols useful. What do others think?

poeschlr:
The generic symbols could be seen as a "template" for adding new symbols. This could make it easier for anyone to either build their own lib or to contribute to this library. 

The danger is that new users might expect that they are able to just use this symbols and connect some footprint to it without checking if there exists a component that has this features with the pinout given in the symbol. (I fear new inexperienced users are more likely to use generic symbols.)

If there is a pinout that is used by many (or most) opamps, then yes a generic symbol might be useful.
Maybe give some example parts that have this pinout in the description. (So that users can find out what do order if they use this part by accident.)

evanshultz:
The dual and quad opamps should have little discussion. Most duals and quads come in DIP and SOIC footprints and they share the same standard pinouts.

For singles, I know there are a number of varieties and would be willing to remove them. Is that satisfactory? (There must be a good way to name singles but all my ideas are super convoluted.)

As far as making mistakes, anybody learning a CAD tool has a lot to learn and some growth will probably happen the hard way. Generic means not for a specific part, and users should know that. I don't believe generic opamps will make KiCad more dangerous, although that's subjective, and it will provide flexibility.

evanshultz:
@jkriege2 @SchrodingersGat @poeschlr 

So what do we do? Seems dual and quad are rather well-accepted above, but single in in contention. Is that right? Drop singles and merge the rest? Support a more limited number of singles?

evanshultz:
@jkriege2 @SchrodingersGat @poeschlr
Any thoughts on this? Do you accept dual and quad generic opamps? If so, I can remove the singles from this PR and we can get it merged. The singles can be added later if we have a good idea how to manage it.

poeschlr:
If the dual and quad op ambs indeed have a common pinout for most manufacturers i vote for adding these symbols. (indicate in the description that it is a generic symbol.)

I think it really might be a good idea to discuss the singles separately. (maybe make a study of what pinouts are used. Maybe we find one that is quite common and can use that for a generic symbol)

SchrodingersGat:
I am still of the opinion that op-amps are not a "generic" device, but, if there is a generic pinout that matches many manufacturers (e.g. doubles, quads) I am ok with that.

evanshultz:
Dual and quad are extremely common. Like a diode in that you can find hundreds of opamps that fit into the standard dual and quad pinouts, with only a few duals and quads that don't conform. It's a de facto standard and that this discussed above.

I added "Generic" to the end of the quad and dual symbols and add pin numbers to the package types in the DCM file.

I then removed all the singles, which often have varying packages or some NC other variable-function pins so they're not nearly so homogeneous.

I believe this is a very workable solution. Any other objections?

evanshultz:
@SchrodingersGat @poeschlr @jkriege2 @hackscribble 
My understanding of the above is that there is majority agreement to add generic quad and dual opamps. Are there more changes here, or could this be merged?

evanshultz:
Oops. Accidental branch deletion.

evanshultz:
@jkriege2 @poeschlr 
It seems both of you guys support dual and quad generic opamp symbols, and Oliver expressed agreement above since those have a common pinout. Can this be merged?

Heiko-L:
I like the idea to have generic opamp symbols with "real" parts as aliases.

My suggestion for naming would be: _Generic.[type].[package].[pin assignment]_

Names get a bit ugly then:
Generic.Single.DIP-8.Nc-Inv-Ninv-Pneg-Nc-Out-Ppos-Nc
Generic.Single.MSOP-8.Nc-Inv-Ninv-Pneg-Nc-Out-Ppos-Nc
Generic.Single.SC-70-5.Ninv-Pneg-Inv-Out-Ppos
Generic.Single.SOIC-8.Nc-Inv-Ninv-Pneg-Nc-Out-Ppos-Nc
Generic.Single.SOT-23-5.Ninv-Pneg-Inv-Out-Ppos
Generic.Single.SOT-23-5.Out-Ppos-Ninv-Inv-Pneg

Anyone has some better abbreviations for pin assignment?

So you have **one** symbol for a lot of "real counterparts".
Would be even better if KiCad had the footprint and footprint filter fields inside the .dcm

evanshultz:
This PR has been open for a long, long time. I've removed the package we didn't all agree on since they weren't consistent, while dual and quad are 99% homogenous and without complaints. Can this be merged?

poeschlr:
Is it a good idea to have the v+ and v- pins duplicated on all units?

Would it not be better to have a separated unit only for the power pins? (If you design it with the same pin spacing as now the user can place this unit on top of whatever unit he wants.)

evanshultz:
I'm happy to do that. Will making this change allow this PR to be merged?

poeschlr:
I can't see a reason why not.

evanshultz:
Power pins are now separated (separate and stacked):
image

evanshultz:
@poeschlr So can this PR be merged now?

poeschlr:
Looks good. (Sorry i did forget about this.)

PR #1115, opened 2017-03-22
Hamamatsu SiPM S13360

matyro:
Added Hamamatsu SiPM


matyro:
I will add a 3D modell to the  footprint later this week

poeschlr:
The footprint filters do not match the footprint you committed.
You either need to add a "*" up front or you need to add "MPPC_Hamamatsu_".

matyro:
I removed the documentation use this symbol for generic SiPM's.
Otherwise I can add a bunch of different one from SensL and Hamamatsu and convert this one back

jkriege2:
Hi!

a few comments on this PR:
1. Please move the general SiPM symbol to device.lib ... there we collect general symbols. Also then change the pins, as we have pin1=K and pin2=A for ALL diodes. Also the generic symbol shold not have a filled adatsheet field and the FPFilter should be something like `MPPC*`. Finally rename this to `D_SiPM` for consistency with the other diode-like detectors, such as `D_Photo`, `D_Radiation`
2. Bring back the specific symbol with the correct footprint-filed AND FPFilter. This one may be A=1 and K=2 as proposed by the datasheet. The diode triangle should NOT be filled (compare to the diode symbols in device.lib (not the ALT-versions!!!)

In both cases: move the anchor to the center of the device (as done for all other diodes). the rule that pins should be on a 100mil grid is valid for ICs ... not for small csymbols like diodes, resistors, transistors ...

Best,
JAN

matyro:
Not sure if the errors in automatic check are ok or real errors...

jkriege2:
Some more comments:
1. Nice work! Thanks so far ;-)
2. In the denerich `D_SiPM` you should write out MPPC also, maybe in the description. Also: These are SPADs (single-photon avalanche diodes), right? So should we also have that name here? 
3. In `opto.lib` you have now a "generic" S13360-30xxCS and the aliases S13360-3025CS S13360-3050CS S13360-3075CS ... Please rename the actual symbol to S13360-3025CS as you don't need the S13360-30xxCS. Then add description/keywords/datasheet for each of the aliases. 
4. the datasheet-field should be empty. Datasheets are only linked in the DCM-file.
5. The footprintf field should be `Opto-Devices:MPPC_Hamamatsu_S13360-30CS`
6. FPFilter should then be `MPPC*Hamamatsu*S13360*30CS*`

Please note that we are currently working on a rework of the KLC. You can find the new version here: https://github.com/SchrodingersGat/kicad-library/wiki/KLC ... that one should become the official version very soon. The new version explains a lot more than the old versions. Also there are some FAQs on several topics: https://github.com/SchrodingersGat/kicad-library/wiki/FAQ

The Travis-scripts will also be updated in time. But some of the errors that you see now (pin-placement on 50mil grid not on 100mil grid) will also be there in future, but can be ignored for simple symbols like diodes ... 

Best,
JAN

poeschlr:
> Not sure if the errors in automatic check are ok or real errors...

Parts of it are real errors.
If you have aliases, these need their description and documentation fields filled out separately.
(Use the description filed to highlight the differences between these parts. The datasheet and keywords can be copied from the "parent")

To do that simply open each alias in the symbol editor and open their respective properties and fill out the fields as needed.

matyro:
The last commit should contain all necessary changes

matyro:
Did i miss something?
Why the close and reopen?

evanshultz:
This will trigger the rules checker to re-run, which was being used to de-bug the rules checker. You should revisit the failed steps, but this was hopefully just a one-time thing. Sorry for any confusion.

matyro:
The last time it worked the error was that the pins were on a 50mil grid and not 100.

Currently travis fails with 

Traceback (most recent call last):
  File "comparelibs.py", line 16, in <module>
    from print_color import *
ImportError: No module named print_color

SchrodingersGat:
Hi @matyro on windows you can either ignore colours with `--no-color` option or install colorama with `pip install colorama`

jkriege2:
A final change please:
Add a strike-through line to the diode symbol, so the symbol looks like the other diode symbols in the lib:
2017-04-05 08_05_10-choose component 336 items loaded

Best,
JAN

jkriege2:
.... and the triangle lines of the diode should be 0.01" thick, whereas the strike-through-line is 0"

Best,
JAN

jkriege2:
Hi!

please make the datasheet field of D_SiPM invisible (see Travis log ... KLC 4.1 you can ignore for this type of device and 4.10 for the generic device in device.lib).

When the changes are done, I can merge!

JAN

jkriege2:
nice work!

Thanks,
JAN

PR #1126, opened 2017-03-28
Move MC1408 from linear library

evanshultz:
Following on with the discussion at https://github.com/KiCad/kicad-library/pull/1081 about updating linear the library, I'm moving out this part which certainly goes here. I will remove it from linear once this PR is merged.

Images:
image

Since TI (formerly National Semi) makes a compatible version still on the market, I made DAC0808 an ALIAS. I created both SOIC and DIP versions with the correct pinouts I found in the datasheets. No KLC violations.

Datasheets:
http://www.pinballpcb.com/datasheets/OnMC1408-dac.pdf (had to use a generic source)
http://www.ti.com/lit/ds/symlink/dac0808.pdf

One question I posted in the aforementioned thread that wasn't answered was about pin names: Can they be lowercase, as was done here, or should they always be uppercase?


SchrodingersGat:
I think that they *can* be lowercase but they should match the datasheet. The datasheet you've linked seems to indicate that the pin names are uppercase?

evanshultz:
@SchrodingersGat Thanks. The original "1408" in the library used mixed case, but all the datasheets I saw were only uppercase. I wanted to check if lowercase was even allowed before I made any changes.

evanshultz:
@SchrodingersGat Done. All pin names are now capital letters and no KLC violations.

SchrodingersGat:
Looks good, thanks for the contribution :)

SchrodingersGat:
Tag me when you remove the original parts

PR #1127, opened 2017-03-28
fix REF50x0 hidden pins types to NC to satisfy DRC

diggit:
As title says + kicad did some whitespace cleanup.


diggit:
Wrong text size. Fixing requires symbol design changes. This PR is meant as hotfix for wrong pin types, so I did not fix it.

PR #1128, opened 2017-03-28
fix LTC6362

diggit:
- fix output pin types
- split MSOP and DFN version (DFN has pad - +1pin)
- fix FP filters - DFN had wrong pitch
- assign default FPs

+kicad did some whitespace cleanup


SchrodingersGat:
Thanks @diggit :)

PR #1134, opened 2017-03-31
Add Nucleo 144 board

El3ctr0nic:
Add F429 and F746 microcontrolers boards


SchrodingersGat:
I think that you have made a typo in the symbol name:

Should NUCLEO144-FR429ZI instead be NUCLEO144-F**4**29ZI?

El3ctr0nic:
Thanks :) . I don't know how it was possible ;/
I think it's ok now.
Footprints 95% done. 
Merge Pls :)

Regards,
Marcin

SchrodingersGat:
Please see errors in the Travis log :)

El3ctr0nic:
Done :) Rravis pass :)

 I have one more problem. When I try check footprints usin kicad-library-utils i get output from  check_kicad_mod.py ;
"Please, resave the files using KiCad to keep indentation standard."
What I do wrong??
I'm poor with python ;/

SchrodingersGat:
It's not your fault that this is happening. Essentially when you use the `--fix` option for the scripts, it makes changes to the `.kicad_mod` files and then resaves them. Sometimes it does not completely match the indentation of the original file, which means that when you open the file in KiCad and re-save it, the file will change.

The new scripts (which were uploaded yesterday, so you should grab them) do a much better job of copying the KiCad native file format.

SchrodingersGat:
And thankyou for the contribution

Issue #1123, opened 2017-03-25
Body diode backwards on BSS138 et al.


horvathp:
I believe the body diode is drawn with opposite polarity on the transistors.lib/BSS138 (2N7002 etc.) model. 
https://www.fairchildsemi.com/datasheets/BS/BSS138.pdf
On a side note, there are generic MOS/NPN/PNP etc. transistors in device.lib with different pin arrangements. I think it would make sense having generic NMOS/PMOS with body diode in device.lib as well.

horvathp:
Furthermore, BSS138 has aliases
TP0610T and VP0610T.
However, these are P-Channel, is BSS138 N-Channel. 
Isn't this a mistake?

cpavlina:
Yeah, you're correct, it's backward, and the aliases are a mistake.

SchrodingersGat:
@horvathp thanks for pointing this out - could you perhaps submit a PR fixing this issue?

jkriege2:
@horvathp a side-note: The NMOS/PMOS in device.lib do HAVE a body diode ... a fix of transistors.lib would be great!

horvathp:
Indeed they have, I was looking at an older version.

SchrodingersGat:
Fixed in https://github.com/KiCad/kicad-library/pull/1135

PR #1135, opened 2017-03-31
fix some MOSFETs in transistors.lib

horvathp:
Proposed fix for the issues mentioned here:
https://github.com/KiCad/kicad-library/issues/1123


SchrodingersGat:
Fantastic, thanks for the fix

PR #1136, opened 2017-04-01
Make Q_xMOS_xxx KLC-compliant by shortening the gate

horvathp:
Shorten the gate terminals to 200 mil and add a small segment to make up the difference. All this in order to silence KLC violation warnings. If this makes any sense.
BTW I'm not sure why the SPARK_GAP component has been moved around along the way, maybe the alphabetic sorting rules have changed recently in the library editor?


SchrodingersGat:
The ordering does seem to change sometimes. Just have to put up with it until the symbols are in different files. Thanks for the fixes

PR #1139, opened 2017-04-03
Fix BSS138 (and aliased 2N7002 MMBF170) diagram from (incorrect) P-ch…

therealprof:
…annel MOSFET to N-channel MOSFET

Signed-off-by: Daniel Egger <daniel@eggers-club.de>


SchrodingersGat:
@therealprof it appears that you have not included some files here, if you have made ALIAS changes?

therealprof:
The ALIASES were there before so they don't show up as changed. The change addresses the incorrect schematics for BSS138 *and* 2N7002 *and* MMBF170 at the same time.

SchrodingersGat:
Thanks :)

PR #1141, opened 2017-04-04
Adding LM4041-ADJ

cedricboudinet:
Hello
Here is a PR for adding LM4041-ADJ in LP package.
If the drawing is fine, I can add it for the other packages.


poeschlr:
Please make the datasheet field invisible.
invisibledatasheetfield

jkriege2:
Looks mostly good. please modify the saymbol as shown below:
2017-04-05 19_32_14-part library editor d__kicad_kicad-library-review_library_references lib

This should correct the too long pin ADJ!

cu
JAN

evanshultz:
This symbol is using a Schottky style tips on the cathode (straight) while the part should use the zener style (at an angle). See the datasheet.

Also, and this is preference and subjective, putting the cathode down instead of to the left will likely be the best orientation in most schematics. I'd prefer that.

jkriege2:
About the tips: yes .... we should then also correct all the other symbols in that lib ... ABout the orientation ... we can also leave it as it is ... then it's in line with e.g. `D`or `D_ZENER` or the other symbols in this lib (just rotate it in KiCAD, when using it ...)
Best,
JAN

cedricboudinet:
I corrected the pin as JAN suggested.
There are still some errors about placement on the grid but they also exist for LM4040: the diode is quite the same but without the feedback pin.
Maybe both of them should be corrected or is it fine?

jkriege2:
Pin placement is fine ... we accept 50mil grid for small components such as resistors, caps, diodes ... thanks,
JAN

PR #1146, opened 2017-04-04
Fix KLC errors and improve footprint details (ON Semiconductor)

hackscribble:
Fixed KLC errors for these two devices.


hackscribble:
My first PR to the library.  If this goes OK, I will work on the other two devices in `onsemi.lib` that have KLC errors.

hackscribble:
I have updated the PR with footprint detail updates to the devices in the library.  

There is one KLC error showing which should be ignored - it relates to the device which has multiple power outputs - they are set as: one is POWER OUTPUT, the rest are PASSIVE.  See https://github.com/KiCad/kicad-library/issues/1159#event-1044069917

SchrodingersGat:
Thanks @hackscribble :)

PR #1154, opened 2017-04-06
correction on TPD4EUSB30

jeromebeguin:


jkriege2:
There is no `TPD4EUSB30` in here ...

jeromebeguin:
Very strange I have it on my repo. Please double check.

jkriege2:
please have a look at:
https://github.com/KiCad/kicad-library/pull/1154/files

jeromebeguin:
I don't know what you mean... 

image

jkriege2:
Aahhh ... OK; I somehow overread the word "correction" ... sorry about that ... now for the PR:
Why did you stack up the NC and the input pins? If you do that the NC pin will be connected to the actually used pin ... which is usually NOT what you want ... is there a rationale behind that?

If you meant to implement for the proposed layout from the datasheet, I would rather have the NC-pin that is grouped with an actual pin (e.g. 7+4 or 10+1) to be placed next to each other. Then the user can (if he wishes to) unhide the NC pins and connect them to the actual pin. This way it is clear from the datasheet what happens and is not hidden in the lib. Also the way of doing it in this PR violates several rules from teh new KLC 2.0 https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention. ... also note that the datasheet explicitly states "Not normally connected" for these pins on page 3!

Also the pin-type of the NC pins should be not_connected and the pin type of all the other pins (IMHO) PASSIVE. COuld you correct that please?

Best,
JAN

jeromebeguin:
Haha, I was expecting that question. Actually it's a bit strange how they made this chip. But it seems the pins are not connected, but they use the pins for routing the diff. pair. Check in the datasheet on page 11. This how they propose to route the chip. It's just a way to achieve that. If you have a better idea...

jkriege2:
I would say go for the proposed way of having NC pins next to the pin it is gruped with ... maybe even unhide the NC-pin (although that would also violate KLC ... but I think here that would be acceptable) ... I would say it is always better to have such a tweak EXPLICITLY in the schematic, than implicitly, especially since the DS points to both uses and is not really clear on that ...

Best,
JAN

poeschlr:
> Haha, I was expecting that question. ...

For future pull requests: It is a lot easier for everyone involved if you mention oddities in the description of the pull request. This way you get better feedback and the maintainers have an easier life.
A screenshot of your symbols and the link to the used datasheet would also help a lot.

jeromebeguin:
@jkriege2 how should I do your way while keeping the symbol nice. Because if I'm changing the NC pins to visible as they are over the other pins it's gonna be really unreadable. Is there another way to connect them ? 

image

jeromebeguin:
Or I just put them visible aside to the other pins and we have to connect them manually inside the schematics. Is that your idea @jkriege2 ?

poeschlr:
Do not stack NC pins with any other pins. Stacked pins are always connected.
(In your example pin7 would always be connected to pin 4. I don't think this is intended. You also get an erc error in this case because a NC pin is connected to another pin.)

jeromebeguin:
@poeschlr actually this is exactly what I want to do.  
image

jkriege2:
@jeromebeguin Yes that's what I meant ... put the actual pin and the corresponding pin directly next to eac other (unhide the NCs) and they can easily be connected in the schematic without problems, if that is desired!

Also this should not crowd the symbol too much ... 

Best,
JAN

jkriege2:
Hi!

yes that's how I thought this should look ... Please remove the commenting text and maybe put it in parantheses into the description, or (very) small inside the symbol?

Best,
JAN

evanshultz:
@jkriege2 Was there anything left to do on this symbol? If the last commit resolved all issues, can this PR be re-opened and merged? Or if the forks were messed up, is the symbol at least ready to merge?

PR #1157, opened 2017-04-07
adding ATAES132a

jeromebeguin:
follow of PR 1113


jeromebeguin:
Something is going wrong, I've never touched the AVR-JTAG, don't know why it's now modified...

jkriege2:
The JTAG should be fine ... your PR removes whitespaces that should not be there anyways (these were introduced by earlier version of KiCAD and are corrected by newer versions). So that's fine!

jeromebeguin:
So it's fine. The VSS pin is at the right place, don't know why Travis found a mistake.

jkriege2:
Symbol looks fine. One thing I noted:
Please split the symbol into ATAES132A_SOIC and ATAES132A_DFN as the latter (DFN) needs an additional pin for the bottom pad. Then you can also set the footprint field for each one and remove the 1EP from teh FPFilter.

Also there seem to be variants with EITHER I2C OR SPI ... did I read teh datasheet correctly there? Would it then make sense to also have variants for those (i.e. end up with 4 variants ATAES132AxR_SOIC and ATAES132AxR_DFN for I2C and ATAES132AxQ_SOIC ATAES132AxQ_DFN for SPI)? ... Or can you reconfigure the device to use either SPI or I2C interface?

Best,
JAN

hackscribble:
A PR was just merged by @SchrodingersGat to fix the test for Vss position: https://github.com/KiCad/kicad-library-utils/pull/106

Not sure how quickly it gets into Travis.

jeromebeguin:
We already had the discussion. Check PR 1113. When you talked about splitting symbol you mean create 2 separate one. Right ?

jkriege2:
Yes I mean splitting into two separate symbols.

Personally I would be in favour to split ... these devices may have the same number, but their function is IMHO significantly different ... + it's more explicit that way.
@SchrodingersGat What do you think on this?

But one split you definitely need to make is between the two package types (due to the different pin counts)!

Best,
JAN

SchrodingersGat:
The two package variants definitely need to be split. However if the same symbol can be used in two separate ways (SPI / I2C) I would leave that as a single symbol.

jeromebeguin:
The footprint is the same this is just the ordering code who change. What is the best way to deal with this ? Still not clear to me.

jkriege2:
If that's agreed on, you can leave it at one symbol (SPI+IIC) per package and add the variants as ALIASes?

JAN

jeromebeguin:
For the variant name what do you prefer ? 

- ATAES132A-I2C  / ATAES132-SPI

- ATAES132A-XR / ATAES132A-XQ

the R and Q comes from datasheet: 
image

SchrodingersGat:
@jeromebeguin you should use the manufacturer variant name e.g. `EQ` for SPI, `ER` for I2C.

Then, add "SPI" or "I2C" to the symbol description so that it is searchable in the component selector.

jeromebeguin:
I've added the alias. I think we have found a good solution for that part.

SchrodingersGat:
@jeromebeguin please remove the "generic" version of this part (not specified as SPI or I2C). Otherwise looks good to merge

evanshultz:
@SchrodingersGat As stated in KLC 3.6, the vendor package should not be used but instead the generic package type. And extending that, I wonder if it's better to use a generic functional name instead of `ER` and `EQ` (actually, the `E` is common so only `R` and `Q` are differentiators)?

This would mean the symbols names are something like `ATAES132A_SOIC_SPI`, `ATAES132A_SOIC_I2C`, `ATAES132A_DFN_SPI`, and `ATAES132A_DFN_I2C`..

jeromebeguin:
Let me know what you prefer.

PR #1160, opened 2017-04-07
Adding muRata NMA series to dc-dc lib

cedricboudinet:
Hello
I created schematics for some muRata dc/dc converters.
I put some nonexistent pins as 'not connected' for filtering, maybe should I remove them and create a new footprint?
Bests


poeschlr:
> I put some nonexistent pins as 'not connected' for filtering, maybe should I remove them and create a new footprint?

Making a correct footprint would be the preferred way.
If you do this could you provide a link from this pull request to the footprint pull request.

It would help a lot if you could provide a screenshot of the symbol. (And for the footprint pull request a screenshot of the footprint would also be a good idea.) Providing a link to the datasheet in the description might be a good idea too.

cedricboudinet:
OK I'll do a specific footprint.
I'll put it in https://github.com/KiCad/Converters_DCDC_ACDC.pretty

cedricboudinet:
@poeschlr : I've done a PR for the footprint: https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/13.
Tell me if the footprint is fine, then I'll correct the symbol.

poeschlr:
You could add your footprint to the footprint field. (Keep the footprint filter as is.)

poeschlr:
I don't think passive is the correct electrical type for the 0V pin.
If this are isolated dc/dc converters than i think the 0v pin should be of type power output. Otherwise maybe power input. (Only if it has an internal connection to vin-)

cedricboudinet:
It is an isolated DC-DC Converter so it should be power output.
However, sometimes when a card is supplied with 5V, I use a NMA to generate -12 and +12V but I want a common GND. Then I connect Vin- to 0V. (I don't think there is any chance to damage the component doing this.)
In this case, if I declare 0V has a power output, it gives an error in Electrical Rules Checker (conflict between pins).
Power input is not a good option neither because it is not a power input at all.

So what is the best solution passive or power output ?
Bests

poeschlr:
I would vote for power output. Yes you get a warning in the case you connect one of the power outputs to another supply, but i think this is a correct erc warning. (After all you connect two power supplies.)

@SchrodingersGat @jkriege2 what do you guys think?

evanshultz:
Looking at other parts in dc-dc.lib, TSR-1-24xx is the only similar part I see. It has the GND pin as Power Input (W). This seems correct to me. Ground is not generated by the Murata NMA, it's *using* ground. It is generating the output by drawing power from the input, but ground just is. the NMA doesn't output 0V, it uses ground as a reference.

@cedricboudinet As Rene said, you should include a screenshot of the symbol/footprint along with the datasheet link with the PR submission. See [KLC ](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) and the [FAQ](https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions). And since these are 1kV isolated, it is fine to use a different reference than ground as the module will only care about the voltage differential on the input to derive the output voltage.

cedricboudinet:
Here is a screenshot of the symbol:
nma0512sc
The datasheet is a available at: http://power.murata.com/data/power/ncl/kdc_nma.pdf

cedricboudinet:
Ok for me I'll put 0V as power output

poeschlr:
@SchrodingersGat, @jkriege2 i think this symbol looks good enough to be merged.
As i stated above i would vote for the 0V pin to be power out. @evanshultz voted for power input. I think one of you two needs to decide what the correct interpretation is.

jkriege2:
POWER_OUTPUT is fine I think ... those device have a galvanic isolation, so this actually a "source" for an isolated GND.

Otherwise: Nice work! Thanks to everyone!

Best,
JAN

jkriege2:
One last remark:
The FPFilters should be changed: They should only contain a single entry (the footprint field is set!), which should be `DCDC*Conv*muRata*NMAxxxxSC*` and `DCDC*Conv*muRata*NMAxxxxDC*` This way we filter only the applicable footprints ... and if variants should be available: also these.

Best
JAN

PS: I can merge when this has been changed!

jkriege2:
Thanks,
JAN

PR #1164, opened 2017-04-11
Added several sensors to sensor.lib

nuntipat:
Here is the list of sensors included in this PR APDS-9960, BMP280,  LIS2DE12, L3GD20H, Si1145, Si7021, TCS34725, TCS40DPR, TMP102, TSL2561 and VL53L0X.

I've checked every parts against KCL. However, there're 3 parts (APDS-9960, TCS34725 and VL53L0X) that use non-standard footprint. Should I create new footprints for each part in some .pretty repos? If so, can you please suggest which pretty repos these footprints should belong to.

Thank you


jkriege2:
Thanks for the contribution!

Yes, having footprint for these would be a great addition. Also please move all of these sensors that are optical devices (e.g. TSL2561) to opto.lib!

Also: For easier and quicker review, somne screenshots and datasheet links directly here in the discussion would really help!

Best,
JAN

PS: I'm currently on holiday and if no one else has a look at this in the meantime, I will do a proper review of your PR end of next week.

poeschlr:
Sorry for the late review. (It seems to have been forgotten.)


**APDS-9960** 
please move to opto lib.
I would change the LEDA and LEDK pins to type passive and move them together. (on the left side.)
Maybe move them in the order (top to bottom) LEDA, LEDK and then LDR.
The specialized footprints could go into the [opto-devices footprint](https://github.com/KiCad/Opto-Devices.pretty) lib.

**BMP280**
I get a 404 error for the datasheet link you added to the symbol.
correct link: https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BMP280-DS001-18.pdf
It seems the spi interface can also be used for I2C
- The SDI pin should be bidirectional
- The SDO pin should be bidirectional

Footprint could go into the [Housings_LGA](https://github.com/KiCad/Housings_LGA.pretty) lib. Name: LGA-8_2x2.5mm_Pitch0.65mm

**LIS2DE12**
The SDO/SA0 pin should be bidirectional (in for i2c, out for spi)
SDA/SDI/SDO should be bidirectional
You could stack the GND pins. (The visible pin should be power input the invisble pins should be passive. Be careful, the symbol editor does not like this. Finish the pins before stacking them.)
I would put the interupt pins next to the spi/i2c interface pins. They are part of this interface. (My interpretation is that the main dataflow is from sensor to controller via these 6 pins.)

**L3GD20H**
The SDO/SA0 pin should be bidirectional (in for i2c, out for spi)
DEN should be an input
Footprint could go into the [Housings_LGA](https://github.com/KiCad/Housings_LGA.pretty) lib. Name: LGA-16_3x3mm_Pitch0.5mm

**Si1145**
should go into opto lib
SCL is the clock? if so then it should be of type input
Pins 5,10 are missing. Include them as type NC and invisble.
Footprint could go into [Housings_DFN_QFN](https://github.com/KiCad/Housings_DFN_QFN.pretty) lib. Name: QFN-10_2x2mm_Pitch0.5mm

**Si7021**
Pins 3,4 are missing. Include them as type NC and invisble.
SCL is the clock? if so then it should be of type input

**TCS34725**
should go into the opto lib
SCL is the clock? if so then it should be of type input
Footprint could go into [Housings_DFN_QFN](https://github.com/KiCad/Housings_DFN_QFN.pretty) lib. Name: DFN-6_2x2.4mm_Pitch0.65mm

**TCS40DPR**
Everything ok

I will check the other symbols tomorrow.

poeschlr:
**TMP102**
SCL should be an input
Footprint should go into [TO_SOT_Packages_SMD](https://github.com/KiCad/TO_SOT_Packages_SMD.pretty) lib: [SOT563 is the same as SC-75-6](http://www.onsemi.com/site/cdroms/micropackages/pdf-docs/onsemi/SOT563%20footprint-prof%23B3380.pdf)
I would name it SC-75-6 with SOT563 mentioned in the description field.

**TSL2561**
SCL should be an input
This component exists in multiple packages. You set the footprint field. This should be indicated in the symbol name. (The name should be TSL2561FN)

**VL53L0X**
please move to opto lib.
SCL should be an input
GPIO1 should be an output. (or bidirectional?)
The gnd pins could be stacked. (The invisible stacked pins would need to be passive.)
The specialized footprint could go into the [opto-devices footprint](https://github.com/KiCad/Opto-Devices.pretty) lib.

nuntipat:
@poeschlr Thank you for reviewing this pr and for your detail suggestions. I should be able to edit and create every footprints tomorrow.

poeschlr:
@nuntipat could you take a look at my suggestions above?

SchrodingersGat:
Hi @nuntipat sorry that this has taken so long. We have now started the process of moving the symbols to the new (and improved!) library at https://github.com/kicad/kicad-symbols

Can you please resubmit this PR at the new library? We will no longer be accepting PRs to this one (it is now considered "legacy")

PR #1169, opened 2017-04-18
Added TO-92 version of LP2950 voltage regulator

hornang:


hornang:
Check script failed because I cannot have '/' in the symbol name. Texas Instruments uses the "/NOPB" suffix extensively in their part numbers. So I guess the question is if the symbol name should refer to a specific manufacturer and part nuber or just refer only to the vendor independent part number. 

I see the discussion is ongoing: https://forum.kicad.info/t/standard-symbol-field-names-initiative/4870

So for now I would go ahead and remove the "/NOPB" suffix?

SchrodingersGat:
HI @hornang please do not add `/NOPB` - this is too much information! For the library we should not need to differentiate between Pb and PbFree part variants :)

In other cases where the MPN has a `/` character, please replace that with a `-` character.

The reason for this is that in a future version of KiCad, symbols will be stored in their own files and cannot have `/` in the filename.

hornang:
@SchrodingersGat Sounds good to me. I will submit an updated version without `/NOPB` tomorrow.

evanshultz:
Hi @hornang! When submitting a symbol, please submit the datasheet and an image of the symbol. You will find this and many more details about the library standards at the [KLC ](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention)and [FAQ](https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions) wiki pages.

Also, for this part, the ACZ suffix means a TO92 package. KLC requires to use the generic package name, not the manufacturer suffix, so please change "ACZ" to "_TO92". This is in addition to removing "/NOPB".

hornang:
@evanshultz Right. I will correct the ACZ to _TO92.

Regarding submitting the datasheet, I thought now that I have referenced the datasheet URL correctly in the library as per KLC (4.10 v)? And where to you want the image of the symbol to be? I suppose you mean to place it in the pull request so people can preview it easily rather in actual git commit(s)?

Sorry, this is my first github/kicad-library pull request ever so I'm still learning.

poeschlr:
> And where to you want the image of the symbol to be? I suppose you mean to place it in the pull request so people can preview it easily rather in actual git commit(s)?

Yes that is the idea behind that.
The same is true for the datasheet link.

It would be best if everyone could provide this information in the description of the pull request. It makes reviewing pull requests a lot easier.

hornang:
image

http://www.ti.com/lit/ds/symlink/lp2950-n.pdf

poeschlr:
Could you add the 3V and 5V versions as alias?

poeschlr:
In the description you wrote:
> 100mA Low Drop Positive Voltage Regulator, Fixed Output 3.3V

I think it would be better to write "Low Dropout"
I would also add the input voltage range to the description.

hornang:
@poeschlr I can add the 5V and 3V version as independent symbols, but I find little information about how the `alias` field is used. Can you enligthen me? 

Regarding the description I will update it. I based my description on a generic impression of the other descriptions in the regul-library.

poeschlr:
From the docu:
http://docs.kicad-pcb.org/4.0.5/en/eeschema.html#component-aliases

Just add aliases as described in the documentation to the component you already made. After you made the aliases they appear in the list of components in the library.

Aliases have the same symbol but the information in the dcm file is separate for each alias. (The information in this dialog: http://docs.kicad-pcb.org/4.0.5/en/eeschema.html#component-documentation)

So after you created the aliases simply fill out the documentation for each alias and commit your changes.
(Make use of the copy from parent button. This saves some work)

hornang:
@poeschlr I don't know if you get notification when I update my branch, but the aliases are now added.

poeschlr:
Looks good.
Everyone who posts a comment in a discussion gets a notification. (Mentioning someone directly also makes sure they get an e-mail.)
I'm also watching this repo which means i get an email everytime something happens. (A new commit, a comment on a pull request, ...)

jkriege2:
Hi!

to me the symbol seems a bit large, compared to newer other symbols in the lib:
2017-04-21 14_47_46-eeschema kicad-library-review _ d__kicad_kicad-library-review-project

So please change these things:
* [ ] Please make it the same size as LM7805CT!
* [ ] In addition, the pin-numbers should be visible
* [ ] Also since this is explicitly TO-92 I would set the footprint field to `TO_SOT_Packages_THT:TO-92_Inline_Narrow_Oval` (this seems to be standard form) and set the FPFilter (remove the generic `TO*92*`) to `TO*92*Inline` This will also match the FORMED LEAD OPTION

Otherwise. Nice work!

JAN

hornang:
@jkriege2 Sure I can make it smaller. Difficult to know what existing components I should compare against since many doesn't follow KLC, but if LM7805CT is about it right I will use it as the baseline.

jkriege2:
I think LM7805CT should provide a good baseline ...

Best,
JAN

jkriege2:
A sidenote: I'm working on an overhaul of this lib here: https://github.com/KiCad/kicad-library/pull/948

hornang:
LM7805CT does not conform to KiCad Library Convention. I ran the script:

```
Checking symbol 'LM7805CT':
  Violating Rule 4.1 - Pin placement
    Pins must be placed on regular grid, and be of correct length
    Pins not located on 100mil grid:
     - Pin IN (1), @ (-400, 50)
     - Pin GND (2), @ (0, -250)
     - Pin OUT (3), @ (400, 50)
```
LM7805CT seems to be designed with 50mil grid which is not according to the convention.

jkriege2:
Uuupps ... I thought I already fixed those problems :-(
I will have a look into that this evening and will come back to you on that. OK?

cu
JAN

hornang:
Sure, no problem. I just double checked the latest commit in master 9bfe6edfbaded584790c660a22cce75d41820aa1 and its not correct there as well.

hornang:
Any news on this subject?

jkriege2:
Hi!

since I'm working on an overhaul of regul.lib (see https://github.com/KiCad/kicad-library/pull/948), I chose to add the LP2950 (in TO-92 and TO-252) to that PR. Also I noted that the pin-numbers in your PR are wrong. The DS gives a bottom view ot TO-92
2017-05-07 11_45_03-lp2950-n pdf - adobe acrobat reader dc
which then translates to OUT=1, GND=2, IN=3 with our TO-92-packages:
2017-05-07 11_46_24-library browser to_sot_packages_tht c__development_kicad_share_kicad_modules

So can this PR be closed, when we merge #948?

best,
JAN

PS: Sorry for the delay!

jkriege2:
PS: Symbols look like this now:
2017-05-07 11_47_57-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib
2017-05-07 11_47_43-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib

hornang:
@jkriege2 Okay. Sounds good! Then I assume I don't have to update this pull request anymore and somebody will close it when appropriate.

jkriege2:
yes, I can close it when my PR is beeing merged!


JAN

poeschlr:
@jkriege2 Am i correct in assuming this can be closed?

jkriege2:
yes, the components are there in the official lib.

JAN

poeschlr:
Closed because the contribution has been included in another pull request.

PR #1172, opened 2017-04-19
creation of +3V8 Power symbol

jeromebeguin:


jeromebeguin:
This is a new power symbol based on the other ones. Only the value have changed. 
image

image

PR #1175, opened 2017-04-20
Update OMRON Relays for KLC compliance

n0dyjeff:
Added description, datasheet, and keywords. Changed reference designator
from "RL" to IEEE standard "K".

### DIFF
--- a/library/relays.dcm
+++ b/library/relays.dcm
@@ -91,13 +91,13 @@ F https://www.elpro.org/de/index.php?controller=attachment&id_attachment=8663
 $ENDCMP
 #
 $CMP G5Q-1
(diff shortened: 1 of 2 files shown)


n0dyjeff:
OK, I clearly don't know what I am doing. I reviewed what I thought were the latest KLC notes, but didn't find anything helpful. I would be happy to make the changes if I just knew precisely what needs to change and how. Any direction you could provide toward getting these checks to pass would be appreciated.

n0dyjeff:
I finally figured out the problem. It had nothing to do with KLC compliance. It had to do with DOS line endings finding their way into the file and messing up the checker. Resubmitted and the checks passed both locally and with Travis

jkriege2:
Thanks!

JAN

PR #1176, opened 2017-04-20
fixed several KLC issues in relays.lib + changed all remaining REFDES from RL to K (Standard!)

jkriege2:
see above


SchrodingersGat:
Thanks Jan

PR #1179, opened 2017-04-20
Added 3D model for SOP-4 housing

n0dyjeff:


poeschlr:
Your step files seem to be exported with freecad 0.16 or with an old 0.17
The problem is that these versions have a bug in the step exporter:
https://forum.freecadweb.org/viewtopic.php?f=3&t=20455

Please either export them with a 0.17 version that is fixed or with 0.15
You can check if the step model is "corrupted" by importing it in freecad. If you get an assembly object something is wrong.
screenshot from 2017-04-21 01-25-24

Could you share your source files as well? (Assuming you designed these models yourself. If not what is the license of the models?)
The source files should go on the https://github.com/kicad/packages3d-source repo.
As soon as the models are merged here it would also be nice if you could also share them in the https://github.com/KiCad/packages3D repo. (The packages3d repo will replace the 3d models directory of the kicad library repo in the future.)

Could you also post screenshots where you show how these models look like with the intended footprint?

poeschlr:
This pull request not only includes the sop 4 file but the relay as well

n0dyjeff:
Thanks very much for your patience and your comments. All of the models 
were either purchased or downloaded from somewhere. I'm in the process 
of researching the licensing of the files; while I know they were all 
licensed for me to use, I wasn't so careful about ensuring that they 
were licensed for sharing. I do know that the G5Q-1A relay is licensed 
for sharing as I purchased it on SnapEda. Their license:

Individual CAD symbols and footprints on SnapEDA are licensed with a 
Creative Commons Attribution-ShareAlike 4.0 International License (CC 
BY-SA) with the Design Exception 1.0. This license allows designers to 
use CAD files downloaded from SnapEDA for commercial and personal use 
free of charge, and make derivative works. If CAD files, or derivatives 
thereof, are shared publicly, attribution must be given to the source 
(SnapEDA and the original author, as applicable), as outlined by the 
requirements of the license.

The others I'm not so sure of and may have to withdraw them if I can't 
determine their IP rights.

With respect to the problems with multiple branches showing up in the 
later pull requests, I apologize. I have been trying out GitHub desktop, 
and it's not real clear to me what it's doing. I will go back to the 
command line in the future if I expect to have multiple commits 
outstanding.

You are correct in noting that I'm using FreeCAD v16. As a complete noob 
to 3D CAD, I figured I should stay with a non-development release. 
However, having developed a small amount of confidence, I'll download 
v17 and save any files that I'm able to share with that version, with 
screenshots.

poeschlr:
> With respect to the problems with multiple branches showing up in the later pull requests, I apologize. I have been trying out GitHub desktop, and it's not real clear to me what it's doing. I will go back to the command line in the future if I expect to have multiple commits outstanding.

The reason for that might be that you branched away from the branch of the old pull request. You always need to branch away from the master branch.
(or more precise from the master branch of the official repo. I add this repo as upstream and my fork is origin.)
This means ideally you would do something like
`git checkout master (or git checkout upstream/master)`
followed by creating your branch.
`git checkout -b [meaningful branch name]`
than you need to setup your local branch to track a remote branch in your fork repository.
`git push --set-upstream origin [meaningful branch name]`

I never tried github desktop but all git interfaces should support something similar. You at least need the possibility to switch branches and to create new branches.

Regarding the license: I'm not sure if even the snap cad files should be used within our library. @easyw what do you think about that? Our repo is under gpl with a similar design exeption to the snabeda exception. The main problem i see is that the 3d files are currently distributed with kicad as one package. I'm not sure if it would be enough to acknowledge snabeda at the repo/file level. (Who ever looks at that?)

easyw:
@n0dyjeff @poeschlr @SchrodingersGat 

> Regarding the license: ...
Our repo is under gpl with a similar design exception to the snapeda exception.

I wouldn't start to share a mixed license models here... but this is only my thought ... 
Anyway for the SSOP4 it can be easily modeled using the FC scripts
https://github.com/easyw/kicad-3d-models-in-freecad
 ... it is only needed to add the right parameters That will add a model aligned to the others in terms of license and geometry
That would be also an occasion to learn how to use these tools and improve the libraries :smile:

easyw:
oops 
also SSOP4 is already don in 
https://github.com/KiCad/packages3D/blob/master/Housings_SSOP.3dshapes/SOP-4_4.4x2.8mm_Pitch1.27mm.step
I have requested to merge it just now
https://github.com/KiCad/kicad-library/pull/1184

n0dyjeff:
Since the license for this one is questionable, and you already have another, perhaps I should close the pull request?

SchrodingersGat:
@n0dyjeff yep :)

PR #1180, opened 2017-04-20
Corrected 3D model for SOT-223 package

n0dyjeff:
The existing model was rotated 90 degrees and needed to be scaled by 40%. Added STEP file as well.


poeschlr:
I assume that the same points i made in your other pull request apply here as well.
Additionally: At least the ssop-4 model is not fused. It would be better if you could fuse the model before exporting. (In part workbench of kicad use the boolean operator union [might be called fusion in older versions].)

poeschlr:
This pull request includes the files of your other pull requests in addition to the sot-223 files.

easyw:
@n0dyjeff @poeschlr sot-223 is already in packages3D kicad lib...
https://github.com/KiCad/packages3D/blob/master/TO_SOT_Packages_SMD.3dshapes/SOT-223.step
I still have to merge all 
TO_SOT_Packages_SMD.3dshapes to kicad-library ...
this dual lib issue is going to be quite time consuming

n0dyjeff:
OK, then should I close this pull request since there's already another acceptable model?

PR #1183, opened 2017-04-21
rework relays.lib to match https://github.com/KiCad/Relays_THT.pretty/pull/1

jkriege2:
* rework relays.lib to match https://github.com/KiCad/Relays_THT.pretty/pull/16 (also rename of 3D symbols)
* added some relays for which footprints existed, but no symbol
* fixed footprints and footprint-filters


SchrodingersGat:
Thanks Jan

PR #1185, opened 2017-04-22
Added STEP and WRL for MPL115A1 sensor

hackscribble:
Datasheet: 
http://www.nxp.com/assets/documents/data/en/data-sheets/MPL115A1.pdf

Related PRs:
Footprint: https://github.com/KiCad/Housings_LGA.pretty/pull/3
Source model: TBA
packages3D: https://github.com/KiCad/packages3D/pull/39

image


hackscribble:
Better model already exists.  Checking if it can be added to the library.

easyw:
if possible I would add this to the repo because this model has the source geometry in FreeCAD and not in proprietary CAD; that is meaning everybody can modify or derive a different but similar model from this

hackscribble:
OK, reopened.

SchrodingersGat:
Looks great, thanks :)

PR #1187, opened 2017-04-22
aligning library to packages3d pull #30

easyw:
https://github.com/KiCad/packages3D/pull/30


easyw:
@SchrodingersGat 
would it possible to align these models to packages3d?
Thx 
Maurice

SchrodingersGat:
Quite possible.

PR #1191, opened 2017-04-23
added series of DIN-connectors (+deleted old version)

jkriege2:
based on this: http://www.mouser.com/ds/2/18/40_c091_abd_e-75918.pdf

2017-04-23 12_51_37-eeschema test_addlin _ d__kicad_test_addlin


SchrodingersGat:
Were the pinouts incorrect on previous version?

jkriege2:
Hi!

no, previously there was only one:
2017-04-23 17_05_07-part library editor d__kicad_test_addlin_ _kicad-library_library_conn lib

Now I added all variants that I could find ... One that corresponds to the (old, deleted) `DIN_5` is this (which I forgot above):
2017-04-23 17_08_22-part library editor d__kicad_test_addlin_ _kicad-library_library_conn lib
(note: My last commit fixes the line-numbering therein)

OK this way?

SchrodingersGat:
Ah these are DIN and the others were Mini DIN. Thanks for both contributions :)

PR #1194, opened 2017-04-24
3D model for Molex MiniFit Jr 2x2

timtashpulatov:
Model converted from Molex original STEP file using FreeCAD and Wings3D


poeschlr:
Do i understand that correctly that this are files you downloaded from the molex webside?
What license do this files use?

timtashpulatov:
To be honest, I have no idea. When downloading 3D models from Molex site, all I get is this:

"I understand that the information I am about to download may not be complete or accurate, and the products depicted in drawings, models, specifications, and other types of content may not be to scale, or the most recent version. I further agree that these files will be used in accordance with the terms and conditions appearing in the web site LEGAL NOTICE including use of these files AT MY OWN RISK."

SchrodingersGat:
Hi @timtashpulatov unless you are the original author of the STEP file then I am afraid we cannot add this to our repository due to potential licensing issues.

However, there is currently a large effort to create high-quality 3D models that are optimized for KiCad.

You can see the current progress [here](http://github.com/kicad/packages3d)

These have all been scripted using [FreeCAD and Python](https://github.com/easyw/kicad-3d-models-in-freecad)

If you have any programming knowledge and are willing to help, it would be greatly appreciated! We are currently in need of Connector 3D models in particular :)

timtashpulatov:
Thanks Oliver, I will check these links :) I am not advanced user with FreeCAD, nor do I know Python unfortunately.

It'd still be great to see if original Molex STEP files could be used.

jkriege2:
Hi!I think you could use them without problems locally on your computer, we just cannot HOST them here ... You might want to check whether the footrpints already have the 3D links set and if not post a PR with at least that ... then KiCAD will use the models if they are found on your computer ... and not use any, if none are present!cuJANtimtashpulatov <notifications@github.com> hat am 24. April 2017 um 13:55 geschrieben:Thanks Oliver, I will check these links :) I am not advanced user with FreeCAD, nor do I know Python unfortunately.It'd still be great to see if original Molex STEP files could be used.—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.

timtashpulatov:
Hi Jan,

The 3D Shape Name for connectors is already filled (see screenshot attached).  What would be the best way to add local storage for 3D shapes so that it does not interfere with distribution library?

image

jkriege2:
Just ensure that the files are in the appropriate subdirectory with the correct naming (I think for current KiCAD you'll have to convert STEP to VRML) as shown above ...

JAN

PR #1196, opened 2017-04-25
Added MCP73871 (battery management + load sharing)

sdalu:
My first try at creating a component, so I certainly misunderstood a lot of the KLC

DS: http://www.mouser.com/ds/2/268/22090a-52174.pdf

mcp73871


jkriege2:
Hi!

nice work overall ... but why are two of the Vss pins POWER_OUTPT instead of POWER_INPUT?
2017-04-26 21_37_03-part library editor d__kicad_kicad-library-review_library_battery_management l

Also please change the FPFilter to `QFN*4x4mm*Pitch0.5mm*`

Thanks,
JAN

sdalu:
Fixed.
* I think Vbat can be used as Power Input or Power Output, I choosed output, don't know if it is the reasonable choice.
* For the FPFilter, shouldn't the pin count be present as well?

jkriege2:
I'll have to look into Vbat this evening.
About the pin-count: No it should not be part of the FPFilter, as the pin-count is filtered by the pin-count filter (THat's also why we need NC pins present as such ;-)

Please have a look at the current KLC 4.12:
and the FAQs on that:
https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions

Best,
JAN

sdalu:
Thanks, re-reading KLC, perhaps should I've done "pin stacking" for Vss, Vin, Vout, Vbat ?

jkriege2:
Hi!

What I would propose:
- stack Vin and Vss
- make one Vout PASSIVE (otherwise you'll get ERC when connecting the two) and stack them ... this will generate errors that you can savely ignore
- make Vbat Bidirectional (I think that best describes it) and stack them

Best,
JAN

poeschlr:
Are Vin and Vss power input pins?
Hidden power input pins are a bad idea which means if you stack these pins the hidden pins need to be passive.
But the symbol editor does not like that. You need to set the pins to their correct electrical type before moving them on top of each other. (If you open the pin properties menu of stacked pins, the symbol editor edits the pin names and electrical types of all these pins to be equal to the pin for which you opened the properties window.)

sdalu:
I stacked Vin/Vout/VSS using passive hidden, and stacked Vbat as bidirectional

poeschlr:
Thanks for your contribution and sorry that the final review took so long.

Issue #1026, opened 2017-02-20
Where to put shield footprints?


svheymann:
Is there a footprint lib for stuff like this i am just working on:
![grafik](https://cloud.githubusercontent.com/assets/25146520/23134832/8740a544-f796-11e6-863d-3cf1c30ba1dc.png)

or eg Arduino/Raspi shields?

Thanks!

jkriege2:
I would say there are two options:
1. You add a project template:
2. You add the actual (micro)controller board as a part that can be mounted on the lowe-side or upper-side of the PCB, see my current PR for Arduino: https://github.com/KiCad/Modules.pretty/pull/3

Best,
JAN

svheymann:
Hmm. What i try to have in the end is a symbol with multiple parts (for each pin header) like this:
![grafik](https://cloud.githubusercontent.com/assets/25146520/23154480/50e8e16e-f80e-11e6-8477-e2e2f3e9d02e.png)

Where all the signal names are already in place like this in eagle:

![grafik](https://cloud.githubusercontent.com/assets/25146520/23154514/833d5e92-f80e-11e6-826c-66bceb730cf0.png)

And when you add one of the connectors the footprint (or part of it, or you can delete the parts you dont need?) for this shield appears in the netlist/pcbnew.

I would find it more intuitive if a shield would be a "part" in the kicad-lib that you can simply add like a normal connector but with multiple sub-parts?

That would have the advantage if you like to add one of the connectorparts later in the design process you could do that using the lib symbol or am i wrong?

jkriege2:
Then, what I proposed in 2) is what you want to do ... whether the Symbol is split or not is a different Story (and I think Splitting should be OK here).

Best,
JAN

svheymann:
Do i understand that correctly, i should add the positive footprint (upper side of the board)?
So that someone can move the needed connectors (provided by the symbol) to the correct position on the lower side of the (negative) shield and then delete the board footprint?

jkriege2:
The footprint should contain all connectors ... If you want variants where only 2 of 5 connectors are present you should make several varying footprints with names describing that ... but I don't really see the sense in that ...

svheymann:
Ok. The PR i referenced adds the board with all connectors on which the shield can be mounted.

svheymann:
I splitted the 5 connectors to 5 different symbols in one lib. Now you can choose which connectors you use on the shield board and move it to the correct position using the board template in the PR. The symbols look like this:

[...]

PR #1200, opened 2017-04-29
Add Olinuxino A20 shield symbol

svheymann:
As discussed here: [Issue 1026](https://github.com/KiCad/kicad-library/issues/1026)

I decided to make a symbol similar to the nucleo144 boards with one single symbol instead of 5 splitted symbols for every single connector. 

I dont know if it is better to have this all-in-one solution or the splitted version. In the splitted version you can add the correct pin headers in the BOM but you have to place the connector footprints by hand to the correct position. And you have to mirror the connector pad numbers so it matches the olinuxino-numberings, which might lead to design errors. In the all-in-one-version the placement is already done and it is easy to see where the cutouts have to be. But it is harder to remove single connectors from the design.

I left the GPIOs in the order of the connectors and did not sort them as they can be configured for all kinds of stuff. And that way you have at least a chance to find them on the board and in the olinuxino-docs again.

I will also add a shield footprint matching this symbol.


svheymann:
After some time to think about this... The best solution would be if there was a multipart-symbol and a multipart-footprint where you can define which of the connectors are used and which are not. And still you could select a specific footprint for each connector like the 1.27mm headers and the placement relation could be defined in this multipart footprint.

Just right now i am thinking about splitting the big symbol into 5 again. Maybe a correct BOM is worth more than that little less work of placing the connectors correctly?

PR #1201, opened 2017-04-29
Added TPIC6595

scibi:
I will appreciate any feedback as this is my first symbol :)


poeschlr:
Please do not use hidden power input pins. (Hidden power input pins are global labels. This is a dirty hack that sadly still survives to this day.)
You can stack them but don't use the electrical type power input in the hidden pins. (Set the invisible pins to passive)
To stack them and have power input only on the visible pin is not that easy though. (The symbol editor does not like this.)
You need to set the electrical types while the pins are not stacked. Then move them on top of each other.

scibi:
@poeschlr I've changed electrical type of hidden pins but now `checklib.py` complains about violating KLC:

```
Checking symbol 'TPIC6595':
  Violating Rule 4.3 - Pin stacking
    Mismatching pins cannot be placed in the same location
    Pinstack @ (-100,800) have different types
```

Should I ignore this error? Maybe it's better just not to stack these pins?

jkriege2:
Hi!

yes, this error can be ignored when stacking power pins (scripts do not really know what someone wants to do, it merely checks against more or less general rules ... if there are violations, but there is a reasom for them, we will ignore them ... remember: everything is checked by hand in addition ;-)

About your symbol: Nice work! Only one change remaining: You set two FPFilters (correct as the DS states 2 possible packages), but you also filles the footprint field. The latter should only be filled, if there is a single option for the package. Please empty it. Then the user can select the package, when changing over from the schamtic to the PCB and will get the correct options (due to the FPFIlter)

Best,
JAN

jkriege2:
Hi!

one more hint: When submitting, please add a screenshot of your symbol + link to the DS ... this makes review a lot easier ,-)

cu
JAN

scibi:
Hi Jan,

Thanks for the tips. I will commit fixed version today. I've got another question - when TPIC6595 is in PDIP package it has suffix N (TPIC6595N), wheny it's in SOIC package it has suffix DW (TPIC6595DW). Should I somehow add them?

image

jkriege2:
Hi!

yur symbol is fine, as both packages have the same pinout, They can simply be listed in the FPFilters and the suer selects whichever package he wants to use, when assigning footprints. The alternative would be to make two symbol with set footprint fields (and only the matching FPFilter) with these names (new KLC!):
- `TPIC6595_SOIC`
- `TPIC6595_DIP`

Do it whichever way you like (I would prefer the first one, as it does not clutter the lib so much ;-)

cu
JAN

evanshultz:
@scibi 
I'm reviewing unmerged PRs that add parts to texas.lib. Please advise if you're going to update this PR so it can be merged, or otherwise it will be closed. We'd like to have your contribution added to KiCad's official library, but we need to reduce the number of open PRs. Thank you.

scibi:
@evanshultz Could you explain what I have to do to have this part merged please? The obvious thing is resolving conflicts - I will do it. But what else? There is a warning:

```
Checking symbol 'TPIC6595':
  Violating Rule 4.3 - Pin stacking
    Mismatching pins cannot be placed in the same location
    Pinstack @ (-100,800) have different types
     - Pin PGND (1) : POWER INPUT
     - Pin PGND (10) : PASSIVE
     - Pin PGND (20) : PASSIVE
     - Pin PGND (11) : PASSIVE
```
but @jkriege2 wrote that I can ignore it.

poeschlr:
rerunning travis.

evanshultz:
@scibi 
Thanks for answering. Yes, first is to resolve the conflicts with your fork.

For the next error, the stacked PGND pins are shown here:
```
X PGND 1 -100 -800 200 U 50 50 1 1 W
X PGND 10 -100 -800 200 U 50 50 1 1 P N
X PGND 20 -100 -800 200 U 50 50 1 1 P N
X PGND 11 -100 -800 200 U 50 50 1 1 P N
```

In this case, I'm not quite sure what's wrong. It looks OK to me and I don't get errors running the check scripts locally. (The scripts may have been updated since the post above because they now understand stacked pins and don't throw an error. Be sure you're using the latest version.)

@poeschlr @jkriege2 
Maybe that the fork has a conflict Travis doesn't run properly? Do you guys know?

evanshultz:
@scibi Can you please fix the branch conflicts? Once that's done we'll see if the error goes away.

evanshultz:
@scibi 
Thanks for responding above, but I haven't seen a response from you since. I do want to get this PR resolved, either by merging it or closing it.

First, you need to fix the branch conflicts. Either do that in this PR or just create a new PR for this part.

Also, just one symbol named `TPIC6595` is correct since the pinout for all packages is the same.

Then, I have some feedback:
- Swap the positions of "U?" and "TPIC6595"
- We do not need to use the logic pin info, so use normal 100mil pins everywhere
- Add available footprints at the end of the description; they are `DIP-8/SOIC-8`
- add `8bit` to the keywords

Thanks, and I hope to add your first symbol contribution to the library soon!

scibi:
@evanshultz 

Thanks for your feedback. I've resolved conflict and then I've done everything you asked for except "We do not need to use the logic pin info, so use normal 100mil pins everywhere". Could you explain what you meant?

evanshultz:
@scibi 
Sorry. I missed your message. I mean that you can use normal pins, the Line type. You do not need to use Inverted or Clock type pins. Just normal Line type pins of 100mil length will be fine for all pins.

I'll try to review the symbol tomorrow. Thanks for the contribution!

evanshultz:
@scibi
I took a look again. It mostly looks fine, but the pins are still too long. I mentioned above to use 100mil pin lengths. Can you please do that?

I also noticed that all the DRAIN* pins are PASSIVE type. They should be OPEN COLLECTOR (that will work for open drain).

Those are the big thing before I merge. But if you can, there are a couple of small things too:
- Use a comma to separate the packages in the description and capitalize every word, so it should be `Power Logic 8-bit Shift Register, DIP-8/SOIC-8`
- Pin name position offset is at the default 40mil, but 10-20mil will look better

Thanks!

evanshultz:
One more thing. When posting a symbol PR, please always include a link to the datasheet and a screenshot of the symbol. This will make it easier for the team to review and merge your contributions.

scibi:
Hi @evanshultz!

Thank you for your remarks. I've fixed all the issues. Don't you think that when there is a 2 digit pin number 100 mil is a bit too little? I've added the link to datasheet in the symbol itself but I will also add them in the PR in the future.

evanshultz:
@scibi Thanks! 100mil is fine for 2 characters or less. This is the standard in the library, with few intentional exceptions (there are some older symbols which violate this, but for new parts this is enforced).

We really appreciate your contribution to KiCad, and I look forward to more in the future!

PR #1207, opened 2017-05-01
added symbol for TDA7292 to audio library

maqifrnswa:
with aliases for TDA7269A, TDA7265, and TDA7265B

tda7292

Requesting exception to KLC 4.1 since the symbol is a triangle, it is hard to get pins in 50 mil increments.

Checking symbol 'TDA7292':
  Violating Rule 4.1 - Pin placement
    Pins must be placed on regular grid, and be of correct length
    Pin -Vs (1) @ (-100,-400) length (170mils) is not a multiple of 50mils
    Pin +Vs (3) @ (-100,400) length (170mils) is not a multiple of 50mils
    Pin M (5) @ (0,300) length (120mils) is not a multiple of 50mils
    Pin -Vs (6) @ (-100,-400) length (170mils) is not a multiple of 50mils
    Pin GND (9) @ (0,-300) length (120mils) is not a multiple of 50mils


jkriege2:
You can always make all pins 100mil long and connect the pin with the triangular otline with a line ...

Best,
JAN

jkriege2:
Some more comments on the symbol:
1. please change the FPFilter to `Multiwatt*Vertical*StaggeredType2*`
2. Would it maybe be better, if the symbol looked like the one in the DS
2017-05-01 11_13_02-en cd00236549 pdf - adobe acrobat reader dc
i.e. with a rectangular outline and function-sketch on the inside ... especially as I believe that most people will use this device with the proposed circuit and not use the two Amps separately? Otherwise maybe change the symbol to look like this:
2017-05-01 11_16_03-part library editor d__kicad_kicad-library-review_library_audio lib
I think that looks a bit better ... or even make it a bit smaller to be in line with the OPAMP symbols in linear.lib. See .e.g. the following for a possibility to add several additional pins:
2017-05-01 11_17_27-part library editor d__kicad_kicad-library-review_library_linear lib

Best,
JAN

maqifrnswa:
Thank you for the comments!

I made it the same size as others in the linear library, used lines so it passes the KLC checks, and fixed the footprint filter.

I'm open to making it rectangular, but here's why I kept it triangular for now: although most people use it in the proposed circuit, it really is just a non-inverting amplifier. You would have to put an input pin on the right side of the device, otherwise you have a long and confusing wire wrapping around the square (see pins 8 and 10 of the figure above). Which do you think would be best: two separate symbols so you can have inputs on the left and outputs on the right, or a rectangle with inputs on the right? 



the DS has input pins on the output side 
tda7292-2

maqifrnswa:
I'm interested in KLC fixing some audio library components and am looking for an opinion for this pull request and future work. I'll also post this on the KCL 2.0+ issue.

Audio library parts LM1876, STK435 and aliases follow what is proposed here. That is, those parts have to units for each stereo channel.

image

Audio library parts LM1877 and aliases use something similar to the datasheet above.

image

Nearly all other audio amplifiers follow the "op amp" style symbols. How do we get a consensus on what we should do? Perhaps you're right, maybe wesShould we just follow what the datasheet does? For example, the popular LM3886 just uses an "op amp" symbol (http://www.ti.com/lit/ds/symlink/lm3886.pdf) while the TDA7292 (here) uses a rectangular "IC" symbol.

jkriege2:
@SchrodingersGat @poeschlr @pointhi Comments on this?

About the symbol in the PR: Could you make the upper and lower pins end on one horizontal line?

THanks,
JAN

evanshultz:
I believe KiCad should have conformance within the library, and that means using "KiCad symbols" even when the datasheet shows something else. For example, if a user wants to change from one similar part to another, why shouldn't it drop right in? (And in other cases where there are multiple vendors for the same part, they might show different symbols making there no clear answer which is correct. Or the datasheet might show multiple symbols styles as mentioned above.)

It will also make maintenance of the libraries easier if there are fewer unique symbols.

As @maqifrnswa says, this part is fundamentally just an opamp. Sure, it has higher current and voltage capabilities than most opamps but that doesn't make it some other type of part.

I've seen total chaos and disarray in libraries, with each person making their own library for each schematic/PCB created. This fundamentally does not work. A better method, which I believe we're working on, is to create rules and then have the entire library conform to those rules for consistency.

Again, only my opinion.

maqifrnswa:
@jkriege2 updated footprint

image

@evanshultz Library-level rules, that sounds like a good idea. For example, the audio library has a dozen or so different symbols for
- an op amp with
  - two inputs (required)
  - one output (required)
  - +V supply (required)
  - -V supply (or GND) (required)
  - GND (if V- is present)
  - Mute (optional)
  - Standby (optional, sometimes combined with mute)

Perhaps this pull request's footprint should be the prototype for the audio library? Just add an extra standby pin between "M" and "V+" in case an "SB" pin is present?

jkriege2:
Hi!
I think the +/--pins should have a `+` or `-` inside the symbol (as common for OPAMPS). To achieve this maybe draw them by hand and set the pin-label to `~`.

Best,
JAN

jkriege2:
@SchrodingersGat @pointhi Opinions on this? Which symbol would you prefer (box or OPAMP-shaped)? Should we try and make the symbols in audio.lib conform to one standard?

Best,
JAN

maqifrnswa:
I made the symbol fit the opamp standard. power pins are closer to the input pins and labels are inside the symbol body.
image
(btw, I'm also the one doing https://github.com/KiCad/kicad-library/pull/1217)

SchrodingersGat:
This looks good, thanks :)

PR #1208, opened 2017-05-01
NXP ARM LPC176x: update and rationalise symbols

hackscribble:
Updated symbol for LPC1763FBD100 to meet latest KLC.  Added aliases for six devices and removed the old separate symbols for these.

```
LPC1764FBD100
LPC1765FBD100
LPC1766FBD100
LPC1767FBD100
LPC1768FBD100
LPC1769FBD100
```
Datasheet:
http://www.nxp.com/documents/data_sheet/LPC1769_68_67_66_65_64_63.pdf

image
image
image
image

PR #1209, opened 2017-05-01
NXP ARM LPC110x and LPC176x: update symbols / add symbols

hackscribble:
### 1. Fixed KLC problems in LPC1102UK.  Added similar symbol LPC1104UK.

Footprint PR:
https://github.com/KiCad/Housings_CSP.pretty/pull/5

Datasheet:
http://www.nxp.com/documents/data_sheet/LPC1102_1104.pdf

image

LPC1102UK pin table:
image

LPC1104UK pin table:
image

### 2. Updated symbol for LPC1763FBD100 to meet latest KLC.  Added aliases for six devices and removed the old separate symbols for these.

```
LPC1764FBD100
LPC1765FBD100
LPC1766FBD100
LPC1767FBD100
LPC1768FBD100
LPC1769FBD100
```
Datasheet:
http://www.nxp.com/documents/data_sheet/LPC1769_68_67_66_65_64_63.pdf

image
image
image
image


poeschlr:
The symbols looks good. But there are problems with the assigned footprint/ footprint filter.

The footprint wlcsp16 is currently not in the housings_bga library.
There is a WLCSP-16_4x4_2.17x2.32mm_Pitch0.5mm footprint in the housings_csp library. Is this the one your symbol needs?

The footprint filter should not match the pin count but possibly something else? 
Maybe something like: *WLCSP*2.17x2.32*

hackscribble:
Thanks @poeschlr The footprint changed libraries during its PR review, and I forgot to update here.  I'll change that, and also the filter.

poeschlr:
I think you now mixed your two pull requests.
Suddenly this pull request also changes for example LPC1763FBD100

And the footprint filter is still wrong. It is missing a couple of "*" (You probably just copied my suggestion from up there. Sadly some of the Stars i put in get used up as formatting symbols. I did not notice that yesterday.)

Correct would be \*WLCSP\*2.17x2.32*

hackscribble:
The LPC176* changes are from this PR: https://github.com/KiCad/kicad-library/pull/1208

Meanwhile, I will fix the footprint filter here.

poeschlr:
So do you want to combine this two pull requests or do you want to revert the changes to the LPC176* symbols in this pull request.

hackscribble:
If you could review the other PR and if it is merged, then there won't need to be any changes to this PR.

poeschlr:
If i get some time today than i will do it. Otherwise it will need to wait until beginning of next week. (I don't have access to a computer for the rest of the week.)

hackscribble:
Just spotted an error in the other PR :disappointed: so will merge the two into this one - that seems the easiest way.  I'll post a comment when it's ready for further review.

hackscribble:
Ready for further review.

poeschlr:
Thanks for this great contributions.

PR #1214, opened 2017-05-02
Add USB Type-C Plug and Receptacle Symbols

Ratfink:
This patch adds USB Type-C Plug and Receptacle symbols to the conn library.  Accompanying footprints can be found at https://github.com/KiCad/Connectors_USB.pretty/pull/3.  The plug and receptacle need separate symbols because their pinout is slightly different (plug is missing two pins and has two other pins named differently from receptacle).  I aimed for consistency with the other USB connector symbols, with ground and shield on the bottom, and all other pins on the right, with VBUS at the top of the right side.  The two symbols are drop-in replacements for one another, which leaves some gaps in the plug symbol's pins; if this is disagreeable I could shift the appropriate pins closer together.

Pin numbering is per the [USB Type-C Cable and Connector Specification, Revision 1.2](http://www.usb.org/developers/docs/usb_31_033017.zip) (contained within the linked ZIP), except for the shield pin which doesn't seem to be assigned a number in the standard.  Because of the A*n*/B*n* pin numbering, it's not easy to extend the pin numbering to add the shield.  I decided on S1, but if another name is preferred I'd be happy to change it.


jkriege2:
nice work !
JAN

PR #1221, opened 2017-05-06
Add ADS1298_TQFP

raiden00pl:
checklib.py script shows "Violating Rule 4.6 - Pin types" for CLK pin (37), but in this chip CLK can be digial input or digital output. I'm not sure which electrical type should I set in this case :)

PDF: http://www.ti.com/lit/gpn/ads1298
ads1298_tqfp


jkriege2:
nice work!

Thanks,
JAN

PR #1223, opened 2017-05-07
3d models for molex usb plugs. (Downloaded from molex)

rpfand:
Added 3D model of Molex USB A Plug SMD PN 48037-1000.

Mates with soon to be checked in Molex_USB_A_PLUG_SMD_48037-1000


jkriege2:
Hi!

thanks for the contribution. 
Could you please let us know where the models are from? If you made them yourself, please also provide the source-files (e.g. FreeCAD) in this REPO: https://github.com/KiCad/packages3D-source
If you downloaded the model from somewhere? Which is the license of the model and is it compatible with the LGPL-license of this repo? Are you allowed to redistribute it here? If not, we unfortunately won't be able to acept the model here.

Also: The WINGS-files are no longer necessary, but a STEP-file would be great. Finally please also submit the WRL- and STEP-files here: https://github.com/KiCad/packages3D (in addition!!!), as this is going to be the 3D mechanical CAD repo in the future.

Thanks,
JAN

rpfand:
I downloaded the model from the molex product page at.

http://www.molex.com/molex/part/partModels.jsp?&prodLevel=part&series=&partNo=480371000&channel=Products

I brought it into FreeCad, exported to Wings, positioned and scaled it to the foot print so that it aligned properly without adjustment in the 3D viewer.

I am not a lawyer. I don't understand why they would create 3D objects of their parts and not want anyone to use them.

From their download page

I understand that the information I am about to download may not be complete or accurate, and the products depicted in drawings, models, specifications, and other types of content may not be to scale, or the most recent version. I further agree that these files will be used in accordance with the terms and conditions appearing in the web site LEGAL NOTICE including use of these files AT MY OWN RISK.



But looking at their LEGAL NOTICE page which is very generic in order to cover everything, it may appear that they don't want images of their product to be distributed.  They give you a download load link and then don't allow distribution. Doesn't make sense.

jkriege2:
usually companies provide these models for personal use, but not for redistribution. If they allow redistribution, they will usually say so explicitly. Otherwise I would assume there is no right to redistribute. It's the same as with say Fotos on the internet. Typically you may view them and store them locally on your computer, but you're not allowed to use them on your own webpage without the explicit consent of the copyright-holder.

So the only thing you can do is drop a message to MOLEX and see whether they answer. If they say it's OK, please post that e-mail here and we see from there.
If you'r unwilling to do so, or they say NO, we'll close this PR.

Best,
JAN

PS: I checked the MOLEX page (http://www.molex.com/legal.html) and the important sentences (to me) seem to be:

> The entire content (images and text) of this Website is copyrighted and may not be distributed, downloaded, modified, reused, reposted or otherwise used, except as provided herein, without the express written permission of Molex.

>  You may download content only for your personal, non-commercial use.

IMHO this expludes reposting the models here, as it is NOT PERSONAL USE. i.e. the only thing that can be done is asking them.

PPS: Note that you can still use the models on your personal computer. When submitting the footprint, KLC require you to add a link to the 3D model even if the model does not exist (yet). So when that link is set and the modl is available locally, it will be used!

rpfand:
I came to the same conclusion. I sent Molex an email asking for permission
to use and distribute the image both on my product and into the KiCAD
library. I would hope they see the benefit of this. I'm not an artist and I
have no interest in duplicating / drawing a 3D object, especially for
something that has already been done.

I know that you are doing the right thing and I am not complaining. Please
don't take this as a complaint. The difference with this and internet
photos is that Molex sells these parts and part of that sales generation
process is they have to document those parts. It makes no sense to me why
they themselves would not want these images to be used by their customers.
They went to the trouble to create the images to showcase their products.
They aren't an imaging company.

Thank you and sorry for your inconvenience.

Richard

SchrodingersGat:
@rpfand thanks for your desire to contribute, I am closing this PR as we cannot accept those models as discussed above. Any further contributions will be greatly appreciated!

PR #1228, opened 2017-05-09
Adding TI LMX9838 Bluetooth Module Lib Support 

gsgill112:
I am new to this, But am not able to make out Error regarding Size of text "Unknown" Which i was not able to find in my library.


gsgill112:
Agghhh again unknown Error ??

SchrodingersGat:
Hi @gsgill112 thanks for your desire to contribute :)

Firstly this symbol should be placed in the `rfcom.lib` library that already exists.

Regarding the "Field UNKNOWN" error:

The text size for the `Footprint` and `Datasheet` field must be set to 50 mils (even if the values are empty)

diggit:
Oh, I remember when I wrote that "UNKNOWN" in utils. Probably not the best description :D
It is known which field causes problem, but it has no name, so it is not easy to describe it.
Anyway, it should be easy to let checklib do such fix for you.

PR #1203, opened 2017-04-29
Added Multiwatt-11 model

maqifrnswa:
EDIT (other pull requests):
KiCad/packages3D-source#7
KiCad/packages3D#45

This is my first model, let me know if anything needs to be updated.

Footprint already exists and already references this model, this model fits the existing footprint.

This is based on:
http://www.st.com/content/ccc/resource/technical/document/datasheet/1a/b1/d0/90/8b/8d/40/fe/CD00236549.pdf/files/CD00236549.pdf/jcr:content/translations/en.CD00236549.pdf

multiwatt-11_vertical_staggeredtype2
multiwatt-11_vertical_staggeredtype2-2

PR #1231, opened 2017-05-09
Adding TI LMX9838 Bluetooth Module Lib Support with 3D Shapes 

gsgill112:
Aggh, Build fails every time Hopefully rectified all errors, Had to get rid of the unknown


gsgill112:
+ LMX9838
Checking symbol 'LMX9838':
    Component LMX9838 has metadata errors:
     - Symbol name should not be included in description
  Violating Rule 4.11 - Default footprint
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint field 'lmx9838' should be of the format 'Library:Footprint'        #### " I have Checked it with various other footprints , this is ridiculous " #######
    Footprint filter 'lmx9838' not correctly formatted
     - Does not contain wildcard ('*') character
    Ground and negative power pins should be placed at bottom of symbol
     - Pin GND (3) @ (800,1100)
     - Pin GND (4) @ (800,1000)
     - Pin GND (30) @ (800,600)
     - Pin GND (31) @ (800,500)
     - Pin GND (32) @ (800,400)
     - Pin GND (24) @ (800,300)
     - Pin GND (17) @ (800,900)
     - Pin GND (18) @ (800,800)
     - Pin GND (29) @ (800,700)
     - Pin VCC (10) @ (-800,-1000)
    Datasheet field must be EMPTY        ##### If I leave it empty it gives an error , else it gives an error ... ??? ########

Shackmeister:
Did you make the 3d model yourself? I'm that case what software did you use?

jkriege2:
Hi!

first of all: THanks for your contribution. We have a set of rules for new symbols here, called the KLC and the check-scripts (Travis) check against some of these rules. You can find the rules here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention and some FAQs on them here: https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions


to explain a bit some of the errors that you get;
1. description should be e.g. `Wireless MCU and Smart RF Transceiver Module`, no need to have `TI LMX9838` in there
2. keywords should be e.g. `RF transceiver MCU`
3. datasheet-link is fine (in the DCM-file, in the GUI, where you set keywords and description)
4. The datasheet-field in the fields (where there is also the footprint-field) should be empty
5. The footprints in KiCAD are sorted into XYZ.pretty-libraries, Your's would then fit best into https://github.com/KiCad/RF_Modules.pretty so the footprint-field should be `RF_Modules:NAMEOFYOURFOOTPRINT`
6. The FPFilter should be set as described in the KLC
7. For the GND-pins I will have to check your actual symbol (later). BTW: Could you please post a screenshot of your symbol?

Finally, We group symbols into larger libs. Yours would fit best into `rfcom.lib`. So please move the symbol there.

Best,
JAN

PS: Please let us know the source of your 3D model, as we have to make sure that the model may be hosted here (if you downloaded it from a manufacturer: They usually grant only the right to use the model privaely, but not to redistribute it, especially not unter the LGPL, which is the license of this Repo!)

gsgill112:
@Shackmeister & @jkriege2 , Yes i have made the 3D model using FreeCAD for mac and few tweaking of the parameters in kicad for making it size fit. 

@jkriege2 : Thanks for the tips, this will surely make things smoother. :) Will try to correct those above mentioned errors and recommit. 

As this is my first contribution to kicad, thanks for the help.

gsgill112:
Ohh and heres the symbol screenshot 
screen shot 2017-05-09 at 11 08 28 pm

P.S> I am a eagle guy and after about an year away from hobby electronics just came back in, It been about 2-3 hours for me installing and get myself up and running with kicad. Again using kicad for the first time :)

Shackmeister:
@gsgil12
When using freecad you should add the kicad stepup macro https://sourceforge.net/projects/kicadstepup/
You can find alot of Q&A's regarding using it on the forum 
You should use this to align the model to the footprint and add the correct material properties.
Please note the footprint settings should be 1 1 1 for scale and 0 0 0 for rotation and offset. 
KiCad stepup will output both a .step file and a scaled (1/2.54, this way it works natively with kicad) .wrl file. Both should be added to this repo (kicad-library) in the 3d packages folder and also be added to this repo https://github.com/KiCad/packages3D
Which is gonna be the new 3d repo 
Lastly you should add your .FCSTD source file to this repo 
https://github.com/KiCad/packages3D-source

I would also suggest you check this
https://github.com/KiCad/packages3D/blob/master/README.md
And also this
https://github.com/KiCad/kicad-library/pull/1203#issuecomment-298437708

jkriege2:
OK, some more comments:
1. You do not need to close this PR and open a new one ((done that several times already, as I've seen), you can simply push changes to your branch of this repository with git. Then the changes will appear here!
2. Please move all the power-pins to the bottom (GND) or top (Vcc ...) borders of the rectangle.
3. usually we group pins by function and often have inputs on the left and outputs on the right (altough it usually does not make sense to e.g. split an RS232- or SPI-interface between left and right ... those would be grouped (often of the left). 

Please reveiw the KLC that I mentioned, as they also give lots of examples/images. In addition please orient yourself on the other symbols in this lib.

Otherwise: If you have questions: Ask away ;-)

best,
JAN

PR #1236, opened 2017-05-10
Add NC7SB3157

cszawisza:
Add NC7SB3157
nc7sb3157


cszawisza:
Apply KLC, now library looks like this
nc7sb3157

cszawisza:
last iteration :)
nc7sb3157

jkriege2:
Hi nice work!

please change the following minor things:
1. The datasheet links should be to the manufacturer directly, i.e. https://www.fairchildsemi.com/datasheets/NC/NC7SB3157.pdf
2. FPFilter should then be `SOT?363*` or `*SC?70*` and `SOT?886*` (this covers both available packages!)
3. Since there are two package options, you should empty the footprint-field.

Best,
JAN

cszawisza:
Hi!
I left all filters, hope that it will not broke the mapping.

jkriege2:
Mostly fine now. Please remove the pin-count from teh SC-70 FPFilter (that's matched by the pin-count filter when assigning the footprints), i.e. it should be `*SC?70*` instead of `*SC?70?6*`

Best,
JAN

jkriege2:
Great work!

Thanks,
JAN

PR #1237, opened 2017-05-12
Fixed type: Jonhson counter -> Johnson counter

farindk:
corrected simple typo


jkriege2:
thanks for spotting!
JAN

PR #1239, opened 2017-05-13
removed old wings files and added new scripted step and WRML models

Shackmeister:
to_sot_smd_wrl_test


SchrodingersGat:
Thanks for the fine work.

Shackmeister:
Please also merge the same PR in packages3d

PR #1241, opened 2017-05-15
Fixed alignment of KA pin on dual colored LED

ashley-b:
Makes it hard to connect wire on normal grid size (50mils)


jkriege2:
Thanks for spotting!

JAN

PR #1244, opened 2017-05-16
add SN74LVC4245A into texas lib

cszawisza:
sn74lvc4245a

I also add aliases, 
SN74LVC4245ADBR
SN74LVC4245ADWR
SN74LVC4245APWR
SN74LVC4245APWT

But sadly I don't know is there a way to set footprint to each alias separately, please point me some doc if there is one


jkriege2:
Hi!

no, this cannot be done with ALIASes ... We do it as follows:
1. If there is a single footprint, set the footprint field and set the FPFilter as required
2. If there are multiple footprints with the same pinout, only set the FPFilter (list all) and leave the footprint field empty. Remove the footprint-specific extensions of the symbol name!
3. If there are multiple footprints (with different pinouts): create several symbols. Add the footprint to the symbol name, e.g. `SN74LVC4245A_SOIC`, `SN74LVC4245A_DIP`, ... and in each set the FPFilter and footprint fields as stated in rules 1 and 2 above.

BTW: I think that symbol should go to `74xx.lib`, or is there a reason why not?

cu
JAN

PS: For details see the KLC (https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) and FAQs (https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions)

cszawisza:
Hi!

SN74LVC4245ADBR
SN74LVC4245ADWR
SN74LVC4245APWR
SN74LVC4245APWT

1. Got that :)
2. This is my case, according to datasheet, pinouts in all cases are the same, so should I remove aliases? (AFAIK package is described by DB / DW / PW letters in aliases names) 
3. Ok

I decided to put this symbol in texas insted of 74xx because. 
1. It seams that only TI produces them
2. This is not a standard 74xx chip, it's similar to 74425 chip (http://assets.nexperia.com/documents/data-sheet/74HC_HCT245.pdf) but it has different pinout
3. AFAIK there are plans to "refactor" 74xx library, but without any clear idea how to do it, so I decided to not add my symbol into this lib.

But of course I will not insist to leave this symbol in texas.lib

bobc:
At time of writing, NXP have an equivalent chip, so I suggest this symbol should go in 74xx.

In the case of package variants, we specify the base symbol `74LVC4245A` and add package spec to description e.g. `Octal Bus Transceiver and 3.3-V to 5-V Shifter With 3-State Outputs, SSOP/SOIC/TSSOP`

SchrodingersGat:

PR #1245, opened 2017-05-17
Add MCP9804 temperature sensor

hackscribble:
Added two versions of MCP9804 temperature sensor for MSOP and DFN packages.

image

Datasheet:
http://ww1.microchip.com/downloads/en/DeviceDoc/22203b.pdf


jkriege2:
Hi!

nice work! One small thing to correct though:
- For the DFN-symbol you need to add a pin 9 `GND/PAD` (DS says: "Exposed Thermal Pad (EP); must be connected to GND.")
- Also please remove the pin-count from the FPFilters (they are not necessary, as this is matched by the pin-count filter). then  the FPFilters should be `DFN*3x3mm*Pitch0.65mm*` and `MSOP*3x3mm*Pitch0.65mm*`

Best,
JAN

hackscribble:
Doh! Silly mistakes :smile: 

Added the extra pin - called it "EP" to match the pin table in section 3 of the datasheet.  Also updated the filters in both parts.

image

jkriege2:
Hi!

one question:
Why do you have `*` in front of the FPFilters?

also:
Maybe add EP/GND? to indicate that this needs to be tied to GND (see DS)

cu
JAN

hackscribble:
> Why do you have * in front of the FPFilters?

Too much time spent with Infineon devices that have "P" or "PG" as a prefix before the conventional package name.  I'll remove the `*`.  

And OK about the pin name.

jkriege2:
nice work!

Thanks,
JAN

PR #1246, opened 2017-05-18
Add TUSB7340 from Texas Instruments

jjuyon:
Hi,
Here is a PCIe 1X to 4 downstream port USB3.0
I didn't find th corresponding footprint so it is coming soon


jkriege2:
Hi!

looks mostly good ... should I wait until the footprint is available or should we add it later by a modifying PR?

cu
JAN

jkriege2:
PS: With the FP-name you proposed, the FPFilter should be `WQFN?MR*9x9mm*Pitch0.6mm*`

PR #1247, opened 2017-05-18
Rename Diodes_ThroughHole.3dshapes

nickoe:
It seems that Diodes_ThroughHole.3dshapes was not renamed in
fa12864fcbd29d90967553951be6359957fceab5 which renamed ThroughHole to
THT. Also the pretties refer to THT, so this seems like an obvious
mistake. (There are still some Choke libs with fullname)


SchrodingersGat:
Thanks, good catch.

PR #1254, opened 2017-05-24
New component: TPS74401 in texas lib

jjuyon:
Hi,
Here is the TPS74401RGW and TPS74401RGR (alias) from Texas Instrument


poeschlr:
The in and out pins could be stacked. (Would make the symbol more compact)
  - all pins on top of each other, only one pin is visible. 
  - The visible pin is of type power input//output the invisible pins are of type passive. 
Be careful the symbol editor does not like this. You need to finish the pins while they are not stacked and move them on top of each other afterwards. (If you open the pin properties of any stacked pin, the editor changes the electrical type of all pins in the stack!)

Don't set the electrical type of an invisible pin to power input. Invisible power input pins are global labels. (This is the dirty hack that currently enables power symbols to work.)

If you don't want the pins to be stacked, change the electrical type of all IN pins to power input. For the OUT pins choose one pin to be your power output pin. The other pins should be passive. (Why only one power output? If you connect multiple power output pins, ERC will throw an error.)

The FB pin is an input and should therefore be on the left side of the symbol. (If possible)

This symbol is only for the VQFN package. please add the suffix _VQFN to the name.
Maybe make clear that the pin 21 is the exposed pad by giving the pin name EP? (or pad/tap as it is called in the datasheet.)

If the vqfn package exists in the library you could fill out the footprint field. (Less work for the user.)

poeschlr:
If you stack the power pins, the test script will complain.
This is because https://github.com/KiCad/kicad-library-utils/pull/116 has not yet been merged.

jjuyon:
Ok,
I did the modifications.
I also stacked the unconnected pins but the check failed only for these pins. Should I leave these pins stacked or not?

For the footprint field, I can't fill it because the alias component don't have the same footprint

Best

jkriege2:
you may not stack NC pins, as this would connect them all together implicitly, which is really not what you want ;-)

For footprint:
1. If both ALIASes have the same footprint: then there is no need to have the ALIAses, simply have the correct FPFilters and leave the footprint field empty
2. If you want one symbol entry per footprint: you need two separate symbols, which have FPFilters for only the one footprint + the footprint field filled.

JAN

jjuyon:
@jkriege2 Ok for the separate NC pins.

I don't get your point 1 about footprint... For me the aliases are used to gather components with same functionality, same pin-out in a same symbol even if they have the same footprint or not. For example the MAX232 in Interface.lib : multiple footprints and multiple aliases.

So Is it OK to keep the it like this? Or maybe I remove the footprint dependency in the component name (RGW and RGR)?

Julien

jkriege2:
Hi!

today it is not possible to have different footprints in ALIASes. Therefore you cannot use them to distinguish version with the same pinout, but different footprints. Therefore you will need to make two separate symbols for each VQFN version, maybe add `_VQFN5x5mm` and `_VQFN3.5x3.5mm` then, OR remove the ALIAS and simply have a symbol `TPS74401_VQFN`, which then has FPFilters for both variants. Also ensure that the FPFilters are specific for the footprints: `Texas*VQFN*RGR*` and `Texas*VQFN*RGW*`. Finally this symbol should go to `regul.lib`, once my overhaul #948 is merged! Maybe also try and make the style conforming a bit more to the style set in that PR (a bit smaller body, position of labels at top ...)

Best,
JAN

jjuyon:
Ok,
I merged my lib after the #948 has been merged and then I moved the TPS74401_VQFN into regul.lib.
But now "The Travis CI build failed" with so many errors...
any Idea?

Best
Julien

suzizecat:
Well, git seems to think that you have edited the whole regul.lib
Since there was an update recently (cf #948) it's kind of strange thet there still some error. Are you sure that you have properly updated your version of regul.lib after this change ?

Best,
Julien

poeschlr:
I would personally do the following.
Export your symbols from your branch.
Synch your local master branch with the master of this repo.
Create a new branch
Import the new symbols into this new branch (via the symbol editor import function)
push your branch and create a new pull request. (close this one afterwards.)
This is the easiest and safest way to do this. 
Rebasing/merging can be dangerous and we might loose some of the changes made by @jkriege2. **This PR should only be merged if there are only changes that concern your contribution.**

poeschlr:
By the way just now i had the problem that git always thought the regul lib and dcm are changed even after i used all normal options to force git to get the current online version. (Not even reset --hard did the trick)
It seems there where some problems with line endings.
I had to do the steps described in this answer on stack exchange https://stackoverflow.com/a/41041863 to get back to a clean repo.
(Don't ask me why. It did work, i'm happy again.)

jjuyon:
Ok. Thanks for your help.
I think I found my mistake. I work on a local copy (clone) of my fork of kicad-library.
But in parallel, to be up to date, I have added the kicad-library official as a remote called "upstream"on my local git. and I often pull with:
`git pull upstream master`
But this time I forgot to to precise I just typed :
`git pull`
Thats explain the two last commit on my fork are wrong and have to be removed...

jjuyon:
So now it is clean and synchronized with the official repo, I removed the two dirty commits I had done.
Also I applied the @jkriege2 suggestions and moved the component to regul library.

poeschlr:
There are now 207 changed files.

jjuyon:
@poeschlr
Ok that's strange !! It just concern line ending. I'm on Linux, so line end are LF and in the offical repo line end seem to be Windows CRLF. Bu I don't know why I have this issue now.
So I will do what you suggest before ans restart from the beginning : Create a new fork, add the new component TPS74401_VQFN and after start a new Pull Request.

jjuyon:
Here is the new Pull Request : #1324

poeschlr:
As stated above there seems to be some sort of problem with the git setup. (Incompatibilities between different versions perhaps?)
When i newly cloned the repo a few days ago i had a lot of files that where marked as changed. (until i did the reset suggested in the stack exchange post.)
Now i have a problem each time i fetch a pull request. (Some files suddenly change when i change back to master.)
I don't really know what causes it. I just make the two commands listed above and hope for a clean repo. (Sometimes i need to repeat this procedure multiple times.)
By the way i'm also on linux. Using git version 2.5.5
@jkriege2 or @SchrodingersGat any ideas?

SchrodingersGat:
@poeschlr what are the files that are causing the issues? I have seen some problems with 3D model files.

poeschlr:
TlDr: Yes some 3d models seem to be problematic (if i force crlf correction somehow). My problems with the regul lib look more like a bug. (or i have configured something in a wrong way)

---

Currently i have always problems with the regul lib.
(And sometimes also with 3d models.)

More information:
I have cloned this repo directly. (removed the push remote off course. origin is this repo not my fork! This clone is only used to check pull requests.)
I can reliably reproduce it by:
`git fetch origin pull/1320/head` (or any other pull request. 1320 did not touch the regul lib so i used it as a testcase)
`git checkout FETCH_HEAD` (After this i still have a clean repo according to git status)
`git checkout master`

If i remember correctly i also got into trouble by using git pull. (can't test this though because i don't have an out of date clone of this repo.)

after this, git status shows that regul lib and dcm have changed. (and `git diff -w` shows that there is a problem with line endings. `git diff` shows all lines differ)
`git checkout regul.lib` or `git reset HEAD --hard` do nothing
I need to `git rm --cached -r .` followed by `git reset --hard` to fix the problem. (until i check the next pull request of course.)

By the way this is a fresh clone of the repo.
At least once i had the problem that after my "fix" the repo showed a lot of 3d models as changed.
(All of them in modules/packages3d/TO_SOT_Packages_THT.3dshapes)
Repeating the procedure "fixed" the problem. (if autocrlf is on input or false. If on true, nothing i can do does fix the problem.)

Currently autocrlf is on input.
If i have it on true: the 3d models get always changed with my fix. Not by checking out master. (The regul lib still gets changed after i checkout master. But is shown as unchanged after my "fix")
Strangely the regul lib is even changed if i set it to false.
If i understand the [documentation](https://git-scm.com/book/tr/v2/Customizing-Git-Git-Configuration) [section: Formatting and Whitespace] correctly, i should use input. Setting it to false should mean that no files are changed. Ever.

If you don't get the same problem than i probably have a buggy version of git. (Tomorrow i will play around with my other pc that has a newer version of fedora.)

jjuyon:
I have a similar issue with CRLF. An example with another repo : Housing_DFN_QFN.pretty
I simply clone it and `git status` give me one file with CRLF line end which appears as modified : 
test

Because of my colleague doesn't have this problem, It seems to be a git configuration or git version issue. I'm using 2.7.4 and he is using 2.13.1 version.
I tried different core.autocrlf configuration, i also tried `git rm --cached -r .` followed by `git reset --hard`
and don't fix the problem.

Any idea?

jjuyon:
I just upgrade git to the version 2.13 from ppa repo for ubuntu. And now it seems OK

poeschlr:
Hm my version is even older. I really need to update my operating system.

suzizecat:
Got the same problem with LEDs.pretty repo, with a fresh install of Git/GitKraken. I don't got these on others repos.

PR #1255, opened 2017-05-24
Added LM4810 dual headphone amplifier to audio.lib.

mifi2909:
The current libraries do not yet contain a component for the LM4810. I would like to propose the version I created, based on the LM4811 that is already in the libraries (but in digital-audio.lib where it probably does not belong; I will report a separate issue to move the LM4811 into audio.lib as well).

From the datasheet:
"The LM4810 is a dual audio power amplifier capable of delivering 105mW per channel of continuous average power into a 16Ω load with 0.1% (THD+N) from a 5V power supply. Boomer audio power amplifiers were designed specifically to provide high quality output power with a minimal amount of external components. Since the LM4810 does not require bootstrap capacitors or snubber networks, it is optimally suited for low-power portable systems. The unity-gain stable LM4810 can be configured by external gain-setting resistors. The LM4810 features an externally controlled, active- high, micropower consumption shutdown mode, as well as an internal thermal shutdown protection mechanism."


mifi2909:
I have observed there are script errors:
Checking symbol 'LM4810MM/NOPB':
  Violating Rule 4.2 - Symbol visual style

I will try to fix those and recommit.

jkriege2:
Hi!

a few comment on your symbol (just form the text ... I will have an actual look this evening:
1. please empty the datasheet Field F4 (next to the footprint field F3)
2. the FPFilter for TSSOP should be `TSSOP*3x3mm*Pitch0.65mm*`
3. please rename the symbol to `LM4810_TSSOP` or simply `LM4810`, we usually don't add the postfixes (especially not the `/NOPB` etc.) to the symbols to keep the number of symbols in check.

More review this evening or in the next few days ;-)

Best,
JAN

mifi2909:
Done that @jkriege2 . 
I hope it is ok now.

mifi2909:
The LM4811 (that already was in the digital-audio.lib), like its sister the LM4810, is a dual headphone amplifier. The difference between these two members of the same family is that the LM4811 has push button volume control and the LM4810 does not.

I have removed the LM4811 from digital-audio.lib as it is not a digital device. 
I fixed the LM4811 to comply with the KLC and moved the new version into audio.lib. I get no errors from the scripts for the LM4811 (but lots of errors on other devices in digital-audio.lib that I did not touch).

Is this ok like this, or should I not have moved the LM4811 to be with it's sister component?

jkriege2:
please revert the changes to ` library-check.sh` in this PR ...

mifi2909:
Geezz... what have I been doing? Sorry.
That was unintended.

mifi2909:
Something went terribly wrong with cut/past in committing the changes. Can we please forget about this pull request, because it results in a lot of bad git history?
It would be better if I start again from a fresh clone.

jkriege2:
OK, just open a new one ... when doing it, please rework the symbol a bit more:
2017-05-24 20_20_05-part library editor d__kicad_kicad-library-review_library_audio lib

If you like, you could even think along these lines with a drawing of the internal structure (not necessary ... I just like ICs with those ;-) ... but then it makes them usually larger:
2017-05-24 20_25_36-part library editor d__kicad_kicad-library-review_library_audio lib

A smaller option would be:
http://m.eet.com/media/1064287/nsm_fig2t.jpg

Best,
JAN

PS: Looking forward to you next PR ;-)

mifi2909:
Thanks. I will do my best.
Working on it  :-)

mifi2909:
See PR #1257 .

PR #1256, opened 2017-05-24
Added LTC660

matyro:
Added [LTC660](http://cds.linear.com/docs/en/datasheet/660fa.pdf) current inverter


SchrodingersGat:
Hi @matyro 

Have a look through the [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention), specifically Rule 4.4

The symbol should be arranged logically, not necessarily following the physical pin locations on the footprint. Many of the other symbols in dc-dc.lib can be seen to follow a more logical layout.
Generally speaking, inputs on the left, outputs on the right, ground at the bottom.

Also, please ensure that all field sizes are 50mils (currently some are 60mils). Have a look at the warning that say "All checks have failed" link - there are some automated scripts that check a bunch of standard rules.

matyro:
Why is 60mills still default when 50mils is correct?

SchrodingersGat:
> Why is 60mills still default when 50mils is correct?

Unfortunately 60mils is hard-coded deep in the KiCAD source code.

You can change this option but it's in a very strange location.

In the schematic editor (NOT the library editor):

`Preferences` -> `Schematic Editor Options` : Select tab `Editing` and then edit option `Default text size`

SchrodingersGat:
Thanks for the changes. One final thing:

Please move the REF and NAME outside the symbol as here:

screenshot from 2017-05-24 22-01-15

Thanks!

SchrodingersGat:
Thanks :)

jkriege2:
Hi!

shouldn't that be an ALIAS for LTC1044 (and its ALias ICL7660), which are already part of dc-dc.lib ...
2017-05-24 20_29_32-part library editor d__kicad_kicad-library-review_library_dc-dc lib
At least it's pin-compatible ... you could then maybe clean up the symbol of the LTC1044 a bit (mainly text position I think).

A plus of the existing symbol is that a Cap fits nicely between pins 2 and 4 and the orientation is somewhat like a voltage regulator (Vin left, Vout right, GND bottom)

best,
JAN

matyro:
You are right, never checked the other symbols. The 660 is more or less the high current version of the existing part. I change the part to match the old one and clean the old one up.

I am not sure if i should alias the new part or create a new part.

jkriege2:
Please make an Alias, this makes the lib a bit more sleek and easier to manage.

Best,
JAN

matyro:
Moved booth 660 to alias and moved REF and NAME to the new position. Pins stayed the same

jkriege2:
OK.

one more thing: The ALIASes cannot cover different packages. Therefore we use either of two ways:
1. There is one separate symbol for each package (then the names should be e.g.`LTC660_DIP` and `LTC660_SOIC`, i.e. we don't use the manufacturer extensions), or
2. there is one symbol (if pinout of packages is the same) that has not footprint set, but has FPFilters that match all package variants. For the current symbol I would therefore suggest:
Add only one LTC660 and set that as ALIAS. Remove the CS8/CM8-variants. the FPFilters already cover the two packages. Finally empty the footprint field (which is a bug in the current lib version).

best,
JAN

matyro:
> Add only one LTC660 and set that as ALIAS. Remove the CS8/CM8-variants. the FPFilters already cover the two packages. Finally empty the footprint field (which is a bug in the current lib version).

Done

jkriege2:
@SchrodingersGat Can we merge this? I would say: YES!

JAN

SchrodingersGat:
Great work :)

PR #1258, opened 2017-05-24
many KLC fixes to opto.lib

jkriege2:
- many KLC fixes in opto.lib
- removed generic symbols (pinouts are too different for them to really make sense)
- checked packages and corrected many (used newly created footprints for those)
- notes on remaining errors:  there are small symbols in this lib (reistor-like LDRs, photodiodes, LEDs, ...) These will cauzse grid-errors and pin-length, but those should be ignored, as e.g. in device.lib, also these cause errors for background-filled
- notes on remaining warnings: TSOP,... and other opto-ICs (e.g. IR-receivers, ...) I think having power-pins on the right for these makes sense ... having the m on top will make the symbols very large and uggly and I think the current way is a common shape for those symbols.


SchrodingersGat:
This all seems reasonable - thanks for the effort @jkriege2

PR #1262, opened 2017-05-28
Added LM27761

matyro:
I added the [LM27761](http://www.ti.com/lit/ds/symlink/lm27761.pdf).

I found no matching footprint to use as alias so i created a new one

unbenannt


SchrodingersGat:
Please change pin type of Vout to `Power output`

PR #1266, opened 2017-05-29
Add MAX9814 - Microphone Amplifier with AGC and Low-Noise Microphone …

mifi2909:
MAX9814 - Microphone Amplifier with AGC and Low-Noise Microphone Bias

Description:
The MAX9814 is a low-cost, high-quality microphone
amplifier with automatic gain control (AGC) and low-
noise microphone bias. The device features a low-noise
preamplifier, variable gain amplifier (VGA), output
amplifier, microphone-bias-voltage generator and AGC
control circuitry.

Datasheet: [Datasheet with Maxim](https://datasheets.maximintegrated.com/en/ds/MAX9814.pdf)

Screenshot:
max9814

Footprint is new and has been merged recently: DFN-1EP-14_3x3mm_Pitch0.4mm

Note:
The NC pins and the Eposed Pad (EP) are to be connected to GND according to the datasheet on page 6. They are therefore available as passive pins in the component, to be connected in eeschema.
When I renamed the NC pins to 'NoC' the scripts were ok, but that violates the naming in the datasheet, so I decided I would ask for a waiver instead.

Scripts fail, because there are pins called 'NC' in agreement with the datasheet, but they are not of type 'Not connected' and they are not hidden. The datasheet recommends connecting them to ground, so they must be available in eeschema. Appearently the script checks for the name 'NC' of pins.

There has been a bit of discussion about this component in the forums: [forum discussion](https://forum.kicad.info/t/new-component-max9814-footprint-pin-names-etc/6557).

m


poeschlr:
Good symbol overall. (I like the "artwork")

Currently the gain pin is set to passive. I would interpret it as an input pin. (The A/R pin is very similar and you chose input there.)

Similarly the BIAS and CG pins should have the same electrical type. (I would vote for passive in this case. But to be honest i'm not really sure about it.)

I might even give the exposed pad the type power input. (If it should always be connected to GND)
Passive for the NC pins is ok.

mifi2909:
Thanks. I appreciate compliments (who does not?).
I will get back to you tomorrow.
I admit having wrestled with the pin types, but I tried to interpret the datasheet. I will have another look at it.
I remember deciding 'passive' for the GAIN pin because it is either connected to GND, or connected to Vdd or not connected at all.  Setting the type to 'input' suggests to me the Gain can be set to arbitrary values with arbitrary voltages on the GAIN pin, which is not the case. It does not interpret a true input signal. The 'leave unconnected' of the GAIN pin is a bit of a struggle.
Knowing the three possibilities for the GAIN pin, would you still opt for 'input'?
I will look at the other pins tomorrow.

poeschlr:
Regarding the gain pin:
I would call it an input pin. (Not an analog input pin put still an input pin.)
Especially because the A/R pin has the completely same description. It has one effect if it is connected to high a second when connected to low and a third effect if it is not connected. (And you used input there.)

Maybe @jkriege2  or @SchrodingersGat have a better insight.

SchrodingersGat:
I agree with @mifi2909 that the GAIN pin is "passive" - especially for prevent unwanted ERC errors if unconnected

poeschlr:
You get an unconnected error for all pin types other than not connect if you do not connect it.
If it should be passive than the same applies to A/R
screenshot from 2017-06-05 10-24-30

mifi2909:
I agree with @poeschlr that the A/R pin and the GAIN pin should be treated equally, since the datasheet has the same type of description. I also checked the rules for ERC and decided, based on the table that @poeschlr included that the best option is to set the pins to 'Unspec'. That will give a warning, but not an error. It is the only choice, I fear.
I will set both to A/R and GAIN to 'Unspec' then.
Agreed?

mifi2909:
I noticed there is a pin type 'Tri state' that I can set, but it is not included in the ERC checks table. The name suggests it would fit the GAIN and A/R pins. How about that?

poeschlr:
If the user wants the third setting the user can use the no connect flag. (This way the intention is also made clear.)
By the way. Three state is an output pin. This is definitely not any sort of output pin.

I don't care if you use passive or input for both pins but i think it should be one of these two. (As stated above i lean towards input put i would accept passive.)

mifi2909:
It is kind of interesting to find in the documentation:
Section 11.8.4 tells me a 'Unspecified' pin is to be used when the ERC checks do not matter, which suggests that it is what we are looking for here. In reality, that is not entirely true, because when I tested it, the ERC threw an error (not even a warning but an error).

mifi2909:
Considering all of the above, I have
set GAIN and A/R to input and
set BIAS and CG to passive.
For the time being we will have to accept that when a user uses the default, he might get ERC errors on some pins.

poeschlr:
I wonder if it would be better if the gain, bias and cg pins would be on the left side.
(I'm really not sure in this case. @SchrodingersGat what is your opinion on this?)

mifi2909:
Bias is on the bottom, because it has to be connected to GND via a C. It easier in eeschema (a few short connections are easier to the eye) to be close to GND. The same is true for CG.
GAIN could be argued about, I guess, but the left side is getting a bit crowded.

mifi2909:
I did what you requested @poeschlr and moved GAIN, CG and BIAS to the left.

New screenshot:
max9814-2

The table from page 6 of the datasheet describing the pins and their functions:
max9814-pintable

m

poeschlr:
I will look into this (and probably merge) as soon as i finished the review of jans regul lib pull request.

poeschlr:
Thanks for this great contribution. (And thanks for putting up with our special wishes.)

PR #1272, opened 2017-05-30
Bat54s

dejfson:
new symbol for BAT54S, updated symbol of BAT54A to comply with KDC (and just because I have used BAT54A as a reference to create BAT54S, so it would not be too wise to have one with KDC rules, and the other one without)


poeschlr:
Great work. Thanks for your contribution.

PR #1275, opened 2017-06-01
Symbol for EA DOGXL 160-7x104 LCD dispaly

Misca1234:
This use
https://github.com/KiCad/packages3D/pull/54


poeschlr:
Overall the symbol looks good.

I have a few questions (and suggestions.)
Should the VBx+/- pins be power input pins? (What does the note voltage converter mean in this context. Am i right in assuming that these pins are for supplying the lcd itself?)

You have VSS2 as pin 25 the datasheet gives pin 24 for it (In other words pins 24 and 25 are exchanged)

If i understand it correctly the serial interface is I2C/SPI. If this is correct, the pin SDA should be bidirectional.
SPI normally has the data output on a separate pin. I'm not sure (from cursory reading the datasheet) how spi is implemented in this display.

You submitted a footprint. Would it make sense to fill out the footprint field such that the symbol is normally connected with this footprint?

Please move the value field to the outside of the symbol.
screenshot from 2017-06-03 21-31-35

Misca1234:
I will fix this as well when the other ones are done
I will close this branch because my master was messed up and I will restart with a new push

PR #1281, opened 2017-06-04
Toshiba tpl2770

Misca1234:
Adding the Toshiba 20Mbps optocoupler, 3D model, schematic and footprint will be pushed


poeschlr:
Hm this pull request does not only include the opto coupler you mention but also the display of another pull request of yours.

Make sure each pull request has its own separate branch and you always start from the current master of this repo.

Misca1234:
I notice that as well, I have different branches for the two different components but they seems to be mixed up

Misca1234:
I have probably updated my forked master with the dispaly and then rebased my optocoupler against my forked master

Misca1234:
Adding the component Toshiba TLP2770 
Housings_SOIC     https://github.com/KiCad/Housings_SOIC.pretty/pull/20
kicad-library         https://github.com/KiCad/kicad-library/pull/1281
3D model             https://github.com/KiCad/packages3D/pull/60

bild

bild

bild

poeschlr:
The footprint field is not the same as a footprint filter.
  - The footprint field points to exactly one footprint. It does not include any wild-card characters
    - Accessed via the field properties dialog. (The big T in the top toolbar)
  - The footprint filter(s) are for matching multiple footprints at once.
    - Accessed via the symbol properties dialog. (The button with the opamp symbol plus a gear.)

Misca1234:
The footprint field are empty and can not be removed, so i need to point to the exact footprint instead

poeschlr:
Last time i seem to have missed that you set the output pin to be of type passive.

I think it should really be of type output.

Misca1234:
Yes, my misstake, but it have now been corrected

Misca1234:
By the way, if I want to add more of Toshibas SO-6L high speed opto couplers in the 27XX serie (which use same footrpint and look and feel), shall I created a new branch or should I just add them to this branch when TLP2770 have been accepted ?

poeschlr:
You can (in theory) add it to the same branch, but as soon as this pull request is merged you need to create a new pull request.

Also remember to resync your branch with the master branch.

I could wait for you do add further symbols here and merge everything at once or i merge this now and you create a new pull request later. What would you prefer?

Misca1234:
I will create a new branch, easier to seperate, so all are good then, great, thx alot for the help

jkriege2:
Hi!

two notes on the symbol:
Please add the strike-through line for the LED (see LED in device.lib) and replace the flash-arrows by straight arrows (we recently changed them in all of opto.lib).

Best,
JAN

Misca1234:
Ok, like this 

bild

Misca1234:
This model need So-6L model in 
https://github.com/easyw/kicad-3d-models-in-freecad/pull/68

poeschlr:
The filling of the diode should be accomplished by setting the fill foreground option of the polyline. Not by filling it with lines.
fill_foreground
I would also make the arrows symmetric.

poeschlr:
I think i now understand what @jkriege2 wanted. Don't fill the led symbol but rather put one line down the center of it. (For reference, this is the LED symbol in device.lib)
screenshot from 2017-06-11 10-55-41

Misca1234:
This branch is obsolete, 
the master was messed up, a new branch will be created

PR #1283, opened 2017-06-04
created new lib "graphic_symbols.lib" 

jkriege2:
- new lib contains a set of graphic symbols (e.g. warning signs) for a schematic (should serve as unique source for such symbols, also for logos!)
- includes symbols (Open Hardware) from logo.lib (MOVED!)
- QUESTION 1 ( @SchrodingersGat @poeschlr @pointhi @CarlPoirier ): There is also a lib graphics.lib, which contains several circles, rectangles, arrows, ... Do we need those (or some of those) and if so, should we integrate them into this new lib?
- QUESTION 2: There are errors, because I hide symbol name and REFDES (these are logos after all) ... But the KiCAD adds a tilde `~` before the symbol name in the LIB-file ... Does anybody have an idea how we should handle the names/REFDES for logos?

Newly created symbols in this lib:
2017-06-05 12_09_41-eeschema test_addlin _ d__kicad_test_addlin


poeschlr:
> Does anybody have an idea how we should handle the names/REFDES for logos?

We could use the same mechanism as is used for power symbols?
Start refdes with # and set the "power symbol" flag (This should tell kicad not to include it in the netlist/cvpcb)

Regarding the graphics lib: I don't know what the usecase for these symbols would be. I think we can delete this lib.

jkriege2:
I would also delete the graphics.lib ... do you know whether there is any documentation on the # in REFDES?

cu
JAN

poeschlr:
I just made the observation that power symbols have \#PWR as reference an the power flag has \#FLG
(So i would assume that we could use \#SYM or so here.)
I'm not sure what kicad does with these references. Maybe ask over at the mailing list.

jkriege2:
OK, I use REFDESs #SYM and #LOGO now + added several arrow symbols + deleted graphics.lib.
There are still arrows, as KiCAD adds a tilde `~` to the symbol name in the file ... when the symbol name text is hidden (but which really makes sense for these symbols!!!). So please ignore this, unless anybody knows why? @SchrodingersGat any ideas ?

Best,
JAN

SchrodingersGat:
I think we can add an update to KLC that checks for hidden labels.

1. Must be for parts with # prefix
2. Ignore extra ~ in this case

I like these additions and also vote that the existing graphic lib should go.

jkriege2:
Hi!

I proposed changes to KLC (https://github.com/KiCad/kicad-library/issues/1287) and the check-scripts (https://github.com/KiCad/kicad-library-utils/pull/126) that support these symbols. With them all symbols check out OK!

Best,
JAN

SchrodingersGat:
Great work :)

PR #1284, opened 2017-06-04
NXP LPC812*20: tidy up symbol / add new symbol for different footprint

hackscribble:
Tidied up existing symbol for LPC812M101JD20 in line with other recent PRs on this library (e.g. https://github.com/KiCad/kicad-library/pull/1209):

- simplified pin names
- reduced symbol size
- fixed metadata and footprint filter

image

Copied this symbol and changed footprint and footprint filter to create LPC812M101JDH20 which is TSSOP variant.

image

Datasheet:
http://www.nxp.com/documents/data_sheet/LPC81XM.pdf


SchrodingersGat:
Nice work!

PR #1286, opened 2017-06-04
NXP LPC81n*16: added symbols

hackscribble:
Created symbol for LPC811M001JDH16 with one alias.

image

Copied symbol to create LPC812M101JTB16, the XSON footprint variant.

image


poeschlr:
I'm not sure about the pin types of the PIDO_10 and 11 pins (pin numbers 7 and 8 for both symbols)
I think they are capable of beeing either output (open drain as you correctly stated) or input which would mean they should be bidirectional.
Have i misunderstood the datasheet?

hackscribble:
Thanks @poeschlr you're right, I think these should be bidirectional.  

I will change them, and also in two symbols in the same library that were merged yesterday.

poeschlr:
Looks good. Thanks for your contribution.

Issue #1287, opened 2017-06-05
KLC-modifications for graphical symbols/power-symbols


jkriege2:
Dear all (and @SchrodingersGat ):

as discussed earlier, I'm proposing here some changes to KLC to allow for graphical and power-flag symbols. This PR is accompagniied by changes to the KLC-check-scripts: https://github.com/KiCad/kicad-library-utils/pull/126

## graphical symbols (no pins, hidden REF, not in BOM)
These require:
- hidden REFDES
- REFDES with leading '#'
- may have no FPFilter and footprint
- may have hidden VALUE field, which causes the symbol name to have a leading tilde '~' (this is not stated explicitly in KLC, as it is an internal "Hack" of KiCAD, but the scripts should know about this!)

## power-flag symbols (1 pin, hidden REF, not in BOM, visible value)
These require:
- hidden REFDES is always `#PWR`
- may have no FPFilter and footprint
- VALUE field is visible
- may have only one pin
- pin signal name is the same as the symbol name

## required changes:

KLC 4.9.i should be changed to:
> The Reference field contains the appropriate Reference Designator (REFDES) and is _visible_. For power-flag symbols (e.g. `+5V`, `GND`, ...), the reference designator is always `#PWR`. For graphical symbols (no pins, not part of the BOM) the REFDES is _invisible_ and arbitrary, but starts with a hash-character `#` (e.g. `#SYM` or `#LOGO`).

KLC 4.9.ii should be changed to:
> The Value field contains the name of the symbol and is _visible_. For graphical (no pins, not part of the BOM) and power-flag symbols the value field may also be _invisible_.

please add to KLC 4.11:
> 4.11.vi For power-flag and graphical symbols, no footprint filter is required and the footprint field should be empty

add KLC 4.13:
> 4.13 Power-flag symbols
>  4.13.i power-flag symbols have a reference designator `#PWR` (cf. KLC 4.9.i)
>  4.13.ii power-flag symbols have exactly one pin of a POWER-type, which is _invisible_
>  4.13.iii power-flag symbols may have invisible value-fields (e.g. for Earth-symbols that are uniquely recognizeable by their symbol shape) (cf. KLC 4.9.ii)
>  4.13.iv the pin's name of power-flag symbols is the same as the symbol name

## PS: Some Images that may be used to illustrate the KLC:
[...]

PR #1289, opened 2017-06-05
reworked power.lib

jkriege2:
- fixed KLC-issues (mainly docu)
- added some missing symbols
- requires new KLC-additions from this: https://github.com/KiCad/kicad-library/issues/1287 ... and the check-scripts that go with it: https://github.com/KiCad/kicad-library-utils/pull/126
- one error remain:
```
Checking symbol 'PWR_FLAG':
  Violating Rule 4.13 - Power flag symbols
    Power-flag symbols follow some special rules/KLC-exceptions
    The pin in power-flag symbols has to be INVISIBLE
    The pin name (pwr) in power-flag symbols has to be the same as the component
 name (PWR_FLAG)
```
As I understand this is some special symbol: This is the only symbol with VISIBLE pin (PWR-OUT) ... does anybody know whether this has to stay as that? If so, we leave it. Otherwise I would be happy to change the pin visiblility ...

Best,
JAN


poeschlr:
The pin of the power flag needs to be power output but i think it could be invisible as well.

jkriege2:
Do you know what this symbol is used for?

Best,
JAN

poeschlr:
Yes to get the nasty erc warnings to go away if you have no power output pin on a net where at least one power input pin is located.

An example would be the connector that supplies your pcb.
Or if you have a passive component in series to your dc/dc converter. (A fuse, the inductance of a lc filter, ....)
With this power flag you tell erc: i promise this net has a connection to a supply.

evanshultz:
Doesn't the pin need to be visible?

I agree with Rene to leave it as is. Any change to this would need to be thoroughly tested since it's critical for making the ERC system work. I expect we all have loads of example circuits to test; I'm pretty sure every KiCad design I've done uses this symbol.

SchrodingersGat:
Agreed, we definitely should not change the POWER_FLAG symbol! Even if it breaks KLC it is a very special case.

jkriege2:
Then this PR should be ready to merge (Errors will go away, once https://github.com/KiCad/kicad-library-utils/pull/126 has been merged.

Best,
JAN

poeschlr:
Do you want to use this as a testcase for the library utils pull request or can this be merged?

jkriege2:
I tested it locally with the new scripts and they worked for me (all green, except PWR_FLAG) ... so IMHO it can be merged also before the PR for the utils, But @SchrodingersGat : Do you need this PR to counter-check the utils-PR?

Best,
JAN

SchrodingersGat:
Thanks Jan

PR #1291, opened 2017-06-05
Add AM26LV32, AM26C32 and AM26LS32 components.

krtkr:
Hi all!

Here is my attempt to contribute to KICAD libraries. I add AM26xx32 receivers, used in my project.

My patch pass checklib.py. Here is a screen shot:

am26lv32


poeschlr:
Quick first impression (I have not really looked at it because i am currently checking another pull request.)

If you want to use the inverting circle on the input pins, make the pins longer such that the pin number is not above the circle.

krtkr:
I would prefer to keep inverting circle to keep pin names compatible with TI's documentation and make it obvious that *B pins are inverting. Here is updated version's screen shot:
am26lv32_02

SchrodingersGat:

PR #1296, opened 2017-06-06
Microchip digital pots and DACs

brandoncurtis:
Added Microchip digital pots and DACs to microchip.lib


jkriege2:
Hi!

although some digital pots reside in vendor-specific libs (Xicor/Maxim), most of them are in `interface.lib`, so please move the new pots there. the same goes for ADCs/DACs and àdc-dac.lib`.

Some more changes:
- Also please check the FPFilters of `MCP49X2`. They should be in the same format as MCP41X1
- Please add the packages to the description, i.e. append e.g. `, DIP-8/SO-8/TSSOP-8
- Finally ensure that NC pins are present in the models (if there are any in the devices), as the pin-copunt filter relies on their presence to match the correct packages!

Thanks,
JAN

PS: Please add screenshots + DS-references to this PR, as that makes review a lot easier!

brandoncurtis:
In the case of [MCP49X2](http://ww1.microchip.com/downloads/en/DeviceDoc/22250A.pdf), where it's available in PDIP, SOIC, and TSSOP and they're all pin compatible, is it preferable to create:

1. a single part with three FPFilters (as in `MCP41X1` in interface.lib), or
2. a single part with aliases that specify the different package versions (as in `MCP4922` and `MCP4922-E/P`, `MCP4922-E/SL`, `MCP4922-E/ST` in adc-dac.lib), or
3. separate parts that specify the different package versions?

Aliases cannot specify different FPFilters, so (1) and (3) make the most sense to me and (1) seems like the cleanest option.  If the different packages are _not_ pin-compatible, then (3) is the clear winner.

evanshultz:
Hi Brandon,

Yes, use the first option. In cases where the multiple packages are not all the packages (for example if DFN was also offered), append `_DIP_SOIC_SSOP` to the symbol name. If all packages have a common pinout, then no suffix is needed.

jkriege2:
Hi @brandoncurtis are you still interested in this?

If so, please fix these issues:
1. fix the conflicts (i.e. merge the master into this)
2. since you have all non-generic variants, remove the generic X-ed out variant
3. FPFilters of MCP49X2 should be more specific (add details, like sizes ... as done for the other symbols)
4. add the packages to the description, e.g. as `, DIP-14/SOIC-14/TSSOP-14` or comparable.

Best & thanks,
JAN

SchrodingersGat:
@brandoncurtis we are moving the symbols to kicad-symbols for the new software release. Please resubmit a PR to https://github.com/kicad/kicad-symbols

Thanks

PR #1298, opened 2017-06-07
Adding AD7390

cedricboudinet:
I created a symbol for AD7390/AD7391 DACs.
There is a violating with rule 4.6: Vout should be Power Output. Is this possible to ignore this error? In this case Vout can not provide more than 1mA.


jkriege2:
Hi!

I will have a closer look this evening ... but as this is a DAC Vout actually SHOULD be an OUTPUT ;-) ... So this warning can be ignored. Note: THat is not an error, but "just" a warning, so you already got a green check for thze symbol from Travis ;-)

Best,
JAN

jkriege2:
Hi!

very nice work! I only found one tiny mistake: Please change the FPFilter from the `MSOP*3x3mm*Pitch0.65mm*` to `TSSOP*4.4x3mm*Pitch0.65mm*` as the MSOP does not match the package in the DS!

Then I can merge!

JAN

jkriege2:
A final request, could you please also add the package to the description? I.e. add `, DIP-8/SOIC-8/TSSOP-8`
That would be great!

Thanks,
JAN

cedricboudinet:
@jkriege2 : is it ok for merging?

evanshultz:
@jkriege2 Can this PR be merged now?

PR #1299, opened 2017-06-07
Added SN74LVC2G14 Dual Schmitt-Trigger Inverter

RaceMouse-of-Denmark:
Hi,

I got these errors when checking the symbol, but I can't find out how to fix them:
Checking symbol 'SN74LVC2G14':
  Violating Rule 4.3 - Pin stacking

If you can give me any pointers in the right direction I'll be really appreciated.
BR
/Fink


suzizecat:
Well, you must have some VCC/GND pins invisibles with the type "POWER INPUT/OUTPUT"
If you want to stack these pins, you should give the invisible ones the type "PASSIVE".

For the footprint, travis says that you should escape '-' characters with '*' or '?' which are wildcards... I'm not sure either, but i guess you can replace '-' by '?' (which stand for one character wildcard as far as i know) and travis would be happy ;-)

You can check out the [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) with the numbers travis gives you to have more informations (4.3 and 4.12)

jkriege2:
Hi!

the FPFIlter should be `SOT?23*` and `*SC?70*`.

ABout the symbol: please make the power-pins visible (invisible power-pins are currently used in the TTL/CMOS-libs,  but this will be removed in future (invisible Power-Pins are connected to a net with the name of the pin, i.e. all such ICs will be powered by VCC and GND-nets and you don't have any control. With visible power-pins this does not (implicitly) happen and you have full control.

Also you can think about splitting the symbol into two (or 3) subsymbols: one for each inverter + 1 for power-supply.

Best,
JAN

RaceMouse-of-Denmark:
Thanks a lot for the pointers. I solved the issues by emerging the 2 symbols into 1. Less elegant but quite useful :-)

bobc:
I think this symbol should go in 74xgxx, but there is already a 74LVC2G14 in 74xgxx, so I don't think this new one is needed.

SchrodingersGat:
@we are moving the symbols to kicad-symbols for the new software release. Please resubmit a PR to https://github.com/kicad/kicad-symbols if it is still needed

Thanks

PR #1300, opened 2017-06-08
Added LP3947

suzizecat:
Battery Charger IC up to 1A in LDO mode USB compliant.

Here comes the component

image

And the documentation (linked in the component) is right here [LP3947 Datasheet (from TI)](http://www.ti.com/lit/ds/symlink/lp3947.pdf)


poeschlr:
I think the footprint filter might need a bit of work.
I would suggest WSON\*4.0x4.0\*Pitch0.4mm\*
Currently we do not have a footprint for this device in our Housings_SON library.
Would you be prepared to supply the footprint as well?
(In that case i would suggest to link the footprint directly via the footprint field.)

As far as i understand it this chip has an exposed pad. This pad should have a separate pin in the symbol. (named EP or TAP with pin number 15)

I think i would move the i2c pins and the diff-amp pin to the left side. Then everything for controlling the device is on the left and the battery interface is on the right.

suzizecat:
Thanks for your suggestions and for the footprint filter, i'll add the exposed pad on the symbol. I didn't noticed it since it's not mentionned on the pin list...

Let's try to supply the footprint, i'll need it anyway but since i don't use Wings3D i'd rather not provide the 3D model myself...

Diff amp is an output that represent the current charge of the battery. So i guess it would be better to put it on the left.

To finish, i put I2C pins at the right to keep a square shape and reduce the size. If you think that it's better on the left side, i don't mind since it's more clear that way.

poeschlr:
First of all: it is not necessary to create 3d models. (It would be nice though.)
But if you want to play around with it:
We do no longer use wings to generate 3d models. There are multiple reasons for this. The main reason is that wings can not create models that can be used in a mechanical cad program. (We need step files for that)

Most new models are [generated via cadquery (python) scripts](https://github.com/easyw/kicad-3d-models-in-freecad/tree/master/cadquery/FCAD_script_generator).
Some specialized models are created with some sort of MCAD tool. (freecad 0.15 or 0.17 are the preferred options.)

If you don't want to look into this, maybe @hackscribble might be willing to provide a 3d model for your footprint. (He recently supplied the models for all other housings_son footprints.)

hackscribble:
> If you don't want to look into this, maybe @hackscribble might be willing to provide a 3d model for your footprint. (He recently supplied the models for all other housings_son footprints.)

Yes, happy to do that.  Looks like I'll be making some improvements to some of the models anyway based on feedback from @easyw, so will be raising a new PR on the models.

suzizecat:
I just created the corresponding footprint, i'll update footprint field when the PR will be accepted

hackscribble:
@suzizecat could you attach the footprint to a comment here (ahead of the PR) so that I can check the model against it?  Thanks.

poeschlr:
Please update the footprint field. (It is a lot easier for us maintainers if we can merge all pull requests in one go. Otherwise there is the danger that we forget one pull request.)

suzizecat:
Footprint updated

poeschlr:
Thanks for your contribution. 
It seems you did not only update the footprint filter and footprint field. (You shortened all pins it seems. and made the symbol a bit thinner.)
Looks very nice.

suzizecat:
Well, i may have lost a commit on the way, i had some issues...

PR #1301, opened 2017-06-08
Added Linx RXM-GPS-RM/FM

suzizecat:
Fixed symbol from #1297 

I created the linx lib and GPS symbols which should go inside.
I dont know if it wouldn't be best to include theses symbols in rfcomp...

Here comes the component : 

image

And datasheets can be found here : 
[GPS RM Datasheet](https://linxtechnologies.com/wp/wp-content/uploads/rxm-gps-rm.pdf)
[GPS FM Datasheet](https://linxtechnologies.com/wp/wp-content/uploads/rxm-gps-fm.pdf)

Julien


jkriege2:
A few requests, juding from the image:
1. make pins shorter (100mil should be enough) 
2. move REFDES/VALUE-labels to the top (compare to other symbls)
3. DO NOT STACK NC-pins, as these would be connected! They need to be separated.

Best,
JAN

suzizecat:
I think that 100mil pins are too short for 2-digit numbers as you can see in the picture below,
image

So i'll reduce them to 150mil instead.
I'm also movings refs as you said.

For invisible NC pins, should i spread them all around the circuit ?
Is this way better ?
image

jkriege2:
yes and yes ;-)

JAN

jkriege2:
Note: The hidden GND pins should be PASSIVE, but the single visible one should be POWER-INPUT. (You'll have to move the visible pin away from the stack, when changing the pin-type, as it will always set the type for ALL pins in the stack ;-)

suzizecat:
Damn, i was sure i changed it >.<
I'm commiting fixes.
Sorry that i don't have time to proprely setup the whole checking stuff on my computer. I'll try to do it asap

suzizecat:
Well... Looks like i messed up and merged two branch the wrong way...
My deepest apologies >.<

suzizecat:
Here it is ! 
Sorry for multiple commits and fails...

jkriege2:
Great work!

Only on thing remains to decide: the lib where this should go:

Please move these parts to `rfcom.lib` (there's already a GPS module there), as this limits a bit  the number of different libs that we have to manage (+ the uncertainty of where to put what).

Then I can merge!

Best,
JAN

jkriege2:
One more question: Do you also plan to provide the footprint? If so, it should go to https://github.com/KiCad/RF_Modules.pretty . The name should be `Linx_RXM-GPS`.

based on this, please also set the footprint field to `RF_Modules:Linx_RXM-GPS` and the FPFilter of this symbol to `Linx*RXM?GPS*`

Best,
JAN

suzizecat:
I'll create the footprint later, but the fields are filled ;-) 
Thanks a lot for your help !

jkriege2:
THanks for the contribution! Great work!

JAN

PR #1302, opened 2017-06-08
ST Bluetooth SPBTLE-RF(0) modules

suzizecat:
Modules SPBTLE-RF added to rfcomp library
Need custom footprint.

Here comes the module : 
image

And the datasheets : 
[SPBTLE-RF](http://www.st.com/resource/en/datasheet/spbtle-rf.pdf)
[SPBTLE-RF0](http://www.st.com/resource/en/datasheet/spbtle-rf0.pdf)


jkriege2:
Hi!

thanks for the contribution!

Two questions: 
1. Why are the pins GPIO2, EXT_LPCLK and ANA_TEST hidden? If there is no good reason for that, they should be visible!
2. Will you  also post the footprint?

Best,
JAN

suzizecat:
Well, it's my bad for not reading correctly the KLC i guess. Those pins are unconnected, but have a name in the datasheet (not just NC). So, should i make them unconnected and visible or rather named NC ?

I'll probably post the footprint, but, i'd rather like if someone could provide 3D model. BTW, how should i do to name the footprint since it's not a standard one ? Should i create a ST library named Modules_ST and add something like SPBTLE_Module_AAxBBmm ?

jkriege2:
Hi!

if you could create the footprint, that would be great. You do not need to supply a 3D model for this device (these are strictly optional and separate from the footprints ;-) ... we usually only provide a "pro-forma" link so the model is displayed, when it becomes available).

For the naming: Please put the footprint in this lib: https://github.com/KiCad/RF_Modules.pretty
The name should be `ST_SPBTLE` (i.e. `MANUFACTURER_DEVICE`).

Hope that helps,
JAN

PS: About the NC-pins: please make them visible, you could name them as follows (and set the pin-mode!):
- EXT_LPCLK/NC:  INPUT
- GPIO2/NC: BIDI
- ANA_TEST/NC: INPUT
Not ideal either, but should work (people can connect these if required for some reason + NC is in name + reproduced I/O props from datasheet + should conform to KLC

suzizecat:
After looking into the datasheet, i got a doubt about the fact that NC pin may be used because i found that in the DS : 
image

jkriege2:
I would leave the NC pins as they are now ... maybe you can enlarge the symbol a bit and set "pin name position offset" to 10 to decrowd the symbol a bit. Also moving the NC-text to the start of the pin names may help to make clearer, what you want to indicate:
2017-06-14 07_05_14-part library editor d__kicad_kicad-library-review_library_rfcom lib

suzizecat:
I've done some modifications, here how it's look

image

jkriege2:
looks way better (IMHO ;-) ... I'll have a closer look this evening ...

Best,
JAN

suzizecat:
I still think that the user shouldn't be  able to route these pin since they shouldn't be routed but well... He'll just have to have a closer look at the DS ;-)

BTW if you are one of the right persons to do that, could you have a look at the corresponding footprint ? It's the PR  KiCad/RF_Modules.pretty#14

suzizecat:
Hi ! Since it's a long time since the last comment, i'm just upping this PR...
Is there a way to notify a lib maintainer ? 
Anyway, i know that there isn't a lot of you guys (and you may have a life, sometimes, i'm sure ! :sweat_smile:) so no problem with delays and thanks a lot for your work !

Misca1234:
> Is there a (right) way to notify a lib maintainer ?
No there is not, beside posting comments in the thread like this.
Note, all posted comments will also show up as emails in peoples in boxes so any posted comment may get the thread attention.

I asked on the forum if there was a way or if we was supposed to post "Ready for review" or such, but there is no such process in place. 

The only way a push may get "higher attention" is if they mark a push as their "own" but even if a librarian mark it as her/his thread it does not mean it get 100% 24h attention. 
Other librarians can step in and review it as well.

It all comes down to when a librarian have the time

The only official signal in these threads between an author and the librarians is if the author write "Not ready for review", then this is interpreted by the librarians that they should ignore the push until further notice.

jkriege2:
Sorry, it took me a while to find some time to have a second look (work and all ;-)

1. please fix the conflicts by merging the upstream master into your branch.
2. Nice work, i.e. I will merge as soon as the conflicts are solved ...

best,
JAN

suzizecat:
Done :wink:

SchrodingersGat:
Thanks @suzizecat

PR #1307, opened 2017-06-10
Corrected FT232RL and ATMEGA168A-AU footprints

### LINKED ISSUE #1304: Incorrect footprints for ATMEGA168-20AU and FT232RL
I'm working on laying out an Arduino board using a TQFP AVR, but it looks like neither component has the correct footprint assigned.

Steps to reproduce issue:
1. Create new project
2. Add FTDI library (Preferences -> Component Libraries -> Add -> /usr/share/kicad/library/ftdi.lib)
3. Place ATMEGA168-20AU (or any alias of it) & FT232RL
4. Annotate schematic
5. Tools -> Assign component footprints

KiCAD faults with the following messages instead of correctly assigning the expected footprints to these parts:
screenshot from 2017-06-09 21-26-32
screenshot from 2017-06-09 21-26-51

I believe the correct footprints for these components are:
TQFP-32_7x7mm_Pitch0.8mm
SSOP-28_5.3x10.2mm_Pitch0.65mm

PhirePhly:
Closes #1304


PhirePhly:
I'm confused why KLC 4.9.iv dictates that we don't use the datasheet feature supported in KiCAD.

I am not interested in fixing any of the other convention issues present on these components; only the user-impacting issue of it not having the correct footprint.

jkriege2:
We do not use the datasheet field, the datsheet link should be in teh DCM file.

i.e. in the editor set the datasheet ink here:
klc_props


... but not here:

 
What you can do: Simply change the footprint and do not touch the datasheet field. Then I'll merge this as a small fix ...

Best,
JAN

jkriege2:
... maybe as an explanation: Using the field in the DCM-file has to do with the ALIASes: If you set the datasheet field, all ALIASes of a component inherit this field contents, the DS-link in the DCM-file is specific to each ALIAS. This is important, if you e.g. make very generic symbols like OPAMPS: Many OPAMPs have the same footprint, so you can make them all as ALIAS to one symbol, but then you want one datasheet per ALIAS, not the same datasheet for all symbols ...

Hopefully this mess will change in future with the new library dataformat ... but until then we have the special KLC-rule on that ;-)

PhirePhly:
Understood. Patch updated to only change the footprint.

Including these sorts of rationales in the KLC would make it much less frustrating for new librarians.

jkriege2:
Hi!

although this is not in there too (we might add it) ... there are also FAQs on the KLC, that explain a bit more:
https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions

Best,
JAN

jkriege2:
One more thing: As you changed the footprints, you should also change the footprint filters (FPFilters). For these two symbols they should be 
`TQFP*7x7mm*Pitch0.8mm*` and `SSOP*5.3x10.2mm*Pitch0.65mm*` respectively.

Could you still do that in this PR please?

Thanks,
JAN

PhirePhly:
Ah yes, I see the value in that if we were to make a hand-solder variant. Done.

jkriege2:
exactly.

Thanks for the contrib and work you put in.

Best,
JAN

PR #1310, opened 2017-06-13
BSS83P - added alias

dejfson:


jkriege2:
thanks for the contribution!
JAN

PR #1314, opened 2017-06-14
Added some of the S34ML0xG10xTFx00x family

suzizecat:
Added S34ML08G101TFx00x, S34ML04G100TFx00x, S34ML02G100TFx00x and S34ML01G100TFx00x.

Lots of pins are unused, so it looks like it's messy on the editor, but it should end up well on eeschema.

image

And the datatsheets : 
[S34ML08G101TFx00x](http://www.cypress.com/file/207441/download)
[S34ML04G100TFx00x](http://www.cypress.com/file/207521/download)


suzizecat:
I saw on some symbol that some unused pins aren't provided at all, should i do the same with that component since there is a lot of unused pins ?

evanshultz:
@suzizecat 
All pins on the physical device should be present in the symbol. This is so the pin numbers can be matched between the symbol and a footprint.

If you see other parts in this family missing NC pins or some other pin type, it would be fine to add them to this PR. If the parts missing pins aren't closely related to this part family, it's best to start a new PR.

Thanks for your contribution!

suzizecat:
If i find again this part, ill make a PR ;-)

suzizecat:
Hi ! 
It seems that this got forgotten somehow, since i think that the symbol is good to go, would it be possible to have a review ? 

Thanks ! =)

poeschlr:
Maybe move the footprint field a bit further outside such that it does not overlap the pins.

The default footprint for the tsop version is missing.
For tsop the first 6 pins should be NC pin 7 should be Rdy/Bsy, ... 
you have pin number 6 for Rdy/Bsy this means you have an off by one error for all pins.
(At least for pins 6..24)

The pinout of the bga symbol is ok.

suzizecat:
Thanks a lot for pointing this out.
I corrected it and changed the TSOP version a little to match both 1 CE (Chip Enable) and 2 CE versions.
Since the 2CE version isn't specified for the BGA version, i didn't changed it.

I also replaced the "F" (which indicate Lead-Free) by a wildcard in all versions. Finally, i let the same number of characters in all version name, since only 1 character is lose for the BGA version to match the full MPN and i think it's better to have a matching name lenght for both (TSOP and BGA) versions.

Here is a screen

image

SchrodingersGat:
@suzizecat we are moving the symbols to kicad-symbols for the new software release. Please resubmit a PR to https://github.com/kicad/kicad-symbols

Sorry that this PR has been neglected. The new system will allow us to address these much more efficiently

Issue #1271, opened 2017-05-30
MAX31855 thermocouple-to-digital converter: duplicate symbols


hackscribble:
`sensors.lib` has MAX31855 ...

![image](https://cloud.githubusercontent.com/assets/7076027/26578280/16f5fdbe-4527-11e7-871a-49b7f0f2269a.png)

`interface.lib` has MAX31855KASA plus aliases for other suffixes ...

![image](https://cloud.githubusercontent.com/assets/7076027/26578365/7db0c106-4527-11e7-96d4-dffd57abb806.png)

If the librarians think this should be rationalised, I'm happy to raise a PR.  Also noticed that the MAX31820 temperature sensor (plus aliases) is in `maxim.lib` not `sensor.lib`.

jkriege2:
If you have time: Go ahead and submit a PR with the changes. I would suggest using the symbol from interface.lib, but maybe moving the footprint field to the bottom right, besides the GND-pin. The symbol in sensors.lib can then be removed (it's more like an interface, as it connects toa thermocouple but does not contain one itself). You could also move those temp-sensors from maxim.lib to sensors.lib.

Best & Thanks,
JAN

hackscribble:
Thanks @jkriege2 I'll work on it this week.

hackscribble:
Closed by https://github.com/KiCad/kicad-library/pull/1595

PR #1316, opened 2017-06-15
Move Maxim temperature sensors

hackscribble:
### Changes

As discussed in this issue: https://github.com/KiCad/kicad-library/issues/1271

Moved MAX31820, MAX31820PAR and MAX31826 1-wire temperature sensors from `maxim.lib` to `sensors.lib`. Fixed KLC errors (visual style, footprint filters).

image
image
image

### Remaining KLC errors

All three symbols have rule 4.5 errors due to power pins being on the left.  The existing symbols all seem to have been inspired by a vertical pin alignment of the TO-92 package of the MAX31820.

I will change pin positions if the librarians think this needs fixing.  Note that the datasheet shows the devices as follows.  VDD is connected to GND when running on parasitic power.

image
image

### Datasheets
https://datasheets.maximintegrated.com/en/ds/MAX31820.pdf
https://datasheets.maximintegrated.com/en/ds/MAX31820PAR.pdf
https://datasheets.maximintegrated.com/en/ds/MAX31826.pdf


hackscribble:
@SchrodingersGat @poeschlr 

Since I shouldn't "mark my own homework", would one of you mind reviewing this PR, especially the query under "Remaining KLC errors" above.  Thanks.

SchrodingersGat:
Given the potential use-case of these parts, I think that the pins-on-left is fine, as long as GND is at the bottom-left and VCC at the top-left.

Otherwise I would suggest adding the same artwork as other temperature sensors e.g. 

image

hackscribble:
Thermometer added.

image
image
image

jkriege2:
What about our usual requirement to have power-pins at top/bottom? Maybe it makes sense to break that here (I did e.g. in opto.lib for the TSOP-ICs as it makes the symbols smaller). If so, we should maybe also rework the other temperature sensor ICs with the same "problem".

JAN

poeschlr:
If we start breaking this rule in nearly every lib, is it a rule anymore?

hackscribble:
Looking at the symbols with the thermometer graphic (recognising that there are probably other temperature sensors without it), most of them have the power pins on top and bottom.  I'm not sure that moving them to the side would reduce symbol size.

So I would prefer to change these three existing symbols to "top and bottom".

jkriege2:
I think we could break that rule for defined cases ... if it makes the symbols easier to use. For the TSOP-IR-Receivers that was the case ... and especially for many small devices that could be the case (I'm thinking of typical temperature sensors, some opto-sensors, so there would be only few exceptions).

But if you come up with a nice top/bottom symbol that's also fine ;-)

Remember: Rules are rules, but we can always make exceptions if there is a good reason for them.

Best,
JAN

jkriege2:
+ I think it is hard to find a set of rules that applies to ALL devices (see e.g. the 100mil grid for ICs and 50mil for small devices as rransistors, diodes, R, ...)

SchrodingersGat:
If we do decide to make an exception we should work out how to define it and add it to the KLC

hackscribble:
Here are the symbols with power pins moved to top and bottom. PR updated.

image
image
image

jkriege2:
Hi!

I think the DQ (output?) pins should be moved to the right ... we usually try to have outputs there ...

Best,
JAN

hackscribble:
Moved ...

image
image
image

hackscribble:
Hi @jkriege2 are these OK to be merged? Thanks.

jkriege2:
A few more things should be fixed:
1. set footprint of MAX31820 to `TO_SOT_Packages_THT:TO-92_Inline_Narrow_Oval`
2. MAX31820PAR: footprint as above and please add the NC pin 3 as invisible NC-pin
3. add the package-name to all description fields (e.g. add `, TO-92` to MAX31820 ...)
4. DS1822Z, DS18B20U, DS18S20Z, DS18S20Z are still in maxim.lib although other DS1822/...-variants have been moved ... is there a reason for that?
5. DS2401 has been moved, but is no temperature sensor ... please move back to maxim.lib and make sure to use a symbol that does not have a thermometer inside
6. would you also move DS28EA00 to sensors.lib, as it is the last remaining thermometer in maxim.lib as far as I can see?

Thanks,
JAN

hackscribble:
Wow, I really managed to mess up the DCM files :-( 

I will fix this PR for the MAX... devices with the latest changes above, then open a new PR for the DS moves.

jkriege2:
Fine with me ... I'll look out for you new PR ... do you close this one, or should I?

cu
JAN

hackscribble:
Closed for renovation :smile:

PR #1317, opened 2017-06-15
Added TI SN65HVD23x family

suzizecat:
Hi ! 
I just created CAN Transceivers SN65HVD230 231 232 233 234 and 235
Please, mind that the 232 have two unconnected pins but i kept the same pin placement to be able to switch to any other SN65HVD23x without problem. 

Please alos note that, unlike the scripts may think, Vref is an ouput ;-)
image

image

Here comes the datasheets : 
[SN65HVD23(012)](http://www.ti.com/lit/ds/symlink/sn65hvd230.pdf)
[SN65HVD23(345)](http://www.ti.com/lit/ds/symlink/sn65hvd234.pdf)

Thanks for your review ! 
Julien


jkriege2:
Nice work!

Thanks.

One question: Why did you also change the `MAX31855KASA` symbol? There the CS-pin should NOT be an inverted pin, because the label already indicates the inverted input (i.e. double-inversion). So please revert changes to that symbol. Then I will merge this otherwise perfect PR ;-)

Best,
JAN

suzizecat:
I didn't touched it myself... Since it's nothing, i'll correct that right away ;-)

jkriege2:
Thanks!
JAN

PR #1309, opened 2017-06-12
MAX31855: fix KLC error / remove duplicate symbol

hackscribble:
As discussed in this issue: https://github.com/KiCad/kicad-library/issues/1271

1. Fixed KLC error in MAX31855KASA by removing double inversion on pin 6. Moved footprint label.

image

2. Removed duplicated symbol MAX31855 from `sensors.lb`.

*Note:  I will raise a second PR for the other point raised in the issue about MAX31820.*

PR #1318, opened 2017-06-15
Fix footprint label position

hackscribble:
Revert change in footprint label location that happened in https://github.com/KiCad/kicad-library/pull/1309.


hackscribble:
Need to fix my error in branch.

suzizecat:
Hi, i might have messed something in #1317... I didn't touch the symbol myself (or i do believe it) but some modifications happened.
In particular, i just saw that the package label came back to 0,0...

hackscribble:
No problem :smile:

suzizecat:
Anyway, i don't understand where does this modification came from at all ^^'

hackscribble:
I think what happened was ...

- We both created local feature branches to make our changes based on the same version of KiCad master.
- My PR was merged first.  This updated KiCad master. 
- Your PR from your feature branch still had the previous version of MAX31855.  So, GitHub thought you intended to change MAX31855.

I think the way to avoid this is to check the PR before submitting it.  If there are changes that were not intended, the fix would have been to do a "pull upstream" to update your master branch to the latest version, then rebase your local feature branch onto this new master.  That would have left your feature branch with just the changes to the devices you worked on.

suzizecat:
Well well, i don't know since i updated master just before starting to work this morning, and GitKraken seems pretty straightforward about it...
image

Anyway, it's fixed so no problems 😛

PR #1319, opened 2017-06-15
Fix footprint label

hackscribble:
Fix footprint label location that was changed in https://github.com/KiCad/kicad-library/pull/1317


poeschlr:
Thanks for this fix.

PR #1320, opened 2017-06-16
ST LPS25HB MEMS Pressure sensor

suzizecat:
Hi ! 
Here is the LPS25HB Pressure sensor from ST.

image

And his datasheet : [ST LPS25HB](http://www.st.com/content/ccc/resource/technical/document/datasheet/58/d2/33/a4/42/89/42/0b/DM00066332.pdf/files/DM00066332.pdf/jcr:content/translations/en.DM00066332.pdf)


poeschlr:
Is there a reason for gnd to be power output? Normally we use power input for supply pins.
Should the reserved pin be removed from the gnd stack? (set to visible and passive, pin name = reserved)
SDA is bidirectional.
I think SA0 is an input. Which means the SA0/MISO pin is bidirectional as well.

suzizecat:
The reserved pin should be connected to GND (according to the DS)
GND changed to power input
Bidirectional pins updated

suzizecat:
Some real fail here...
I think i'll open a new PR...

PR #1323, opened 2017-06-18
Ea dogxl160 7

Misca1234:
> **I have a few questions (and suggestions.)**
> **Should the VBx+/- pins be power input pins? (What does the note voltage converter mean in this context. **
> **Am i right in assuming that these pins are for supplying the lcd itself?)**

Yes, I guess that to, I guess they are for some sort of volatage amplifier

> ** You have VSS2 as pin 25 the datasheet gives pin 24 for it (In other words pins 24 and 25 are exchanged)**

You are absolutely correct, I fix that 

> ** If i understand it correctly the serial interface is I2C/SPI. If this is correct, **
> ** the pin SDA should be bidirectional.**

It have been changed 

> ** SPI normally has the data output on a separate pin. **
> ** I'm not sure (from cursory reading the datasheet) **
>**how spi is implemented in this display.**

SPI and I2C share the same pins, you select via BM0 and D6 wich type of serial interface you want.

> ** You submitted a footprint. Would it make sense to fill out the footprint **
> ** field such that the symbol is normally connected with this footprint?**

It is fixed

>**Please move the value field to the outside of the symbol.**
Done

Almost all displays from the EA DOG serie require a backlight expect for just a couple few,
one should not have it and the rest may have it.

bild


suzizecat:
For a pure good-looking purpose (and IMHO), you should VSS1, VSS,VDD and VDD2/3 to make them centered (don't mind the hidden pins, you won't see them in eeschema)

By the way, (if i'm correct) all hidden pins should be named NC or not be hidden (KLC 4.7)

poeschlr:
For reference, old pull request: https://github.com/KiCad/kicad-library/pull/1275

Misca1234:
Well, this is a little problem for me, because A1,A2,A3C1,C2,C3 are used to power feed a background light, 
if this is mounted.
This mean that the EA DOG displayes always have these pins, even if you dont have a background light.

The EA DOG's displays are "piggy" mounted on top the background light package, if such is used.

My ambition is to make two schematics, one with background pins visible and one with without, 
This one is without background light however the pins A1,A2,A3,C1,C2 and C3 are still physical there.


So my dilema is 
1)
Make two schematics, one with visible and one with not visible "background light power feed pins"
Make one footprints that look the same for both schematics but loosing the 3D model with background light package
or
2)
Make two schematics, one with and without the "background light power feed pins" that match its specific foot print.
Make two footprints that should look the same but have different 3D models (one with the background light package)
or
3)
Make one schematic, with the "background light power feed pins" visible, the user have to "NC" them manually if no 
background light is used
Make one footprints that should look the same for both schematics, but loosing the 3D model with background light package
or
4)
Make one schematic, with the "background light power feed pins" visible
Make two footprints that should look the same but have different 3D models (one with the background light package)

Misca1234:
To start with we can do like his 

bild

Misca1234:
Or like this 

bild

suzizecat:
If i understand correctly, you got one component, but with an option that may (not) be mounted. Is the final user mounting the backlight or does the component comes with it already mounted ? Because, if the user can mount it, you shouldn't make two schematics.

For me, the most "KLC-like" would be the 4th. With this one you can create two footprint and just add something like `_Backlight` at the end of the footprint.

Misca1234:
You buy the back light component separately and mount it your self.

Well, once upon time, (for 14 days ago) I made 2 foot prints and 2 3D models, one with and one without the back light and the one with back light had the the suffix _back light ;)

evanshultz:
@Misca1234 

This is a confusing thread. First, in [this post](https://github.com/KiCad/kicad-library/pull/1323#issuecomment-309496491) you use the word "schematic". I believe you mean "symbol". You also wrote "background pins visible" but I think you mean "backlight pins visible".

Next, I believe [this datasheet](http://www.lcd-module.com/eng/pdf/grafik/dogxl160-7e.pdf) is the correct one. It helps me to have all information together.

Now for the conceptual/architectural discussion:
All pins for the backlight (1, 2, 3, 14, 15, and 16) should be visible. The KiCad user can mark unused pins as NC on the schematic if the user does not want the backlight function.

In KiCad, symbols are universal and can use multiple footprints in this case. So that leaves only options 3 and 4 above. And since a footprint can only link to a single 3D model, that leaves option 4. Just as @suzizecat suggested.

And review of the part:
- Pins 18-21 are the power supply for the LCD. Those should not be on the side. Move the VBx+ pins to the top and the VBx- pins to the bottom.
- It's unclear to me what pin 17 is for. What does VLCD do? I find a few schematics using this LCD on a Google search, and all but one of them show an RC to ground from pin 17.
- Separate the VDDx pins from the Ax+ on the top, and the VSSx pins from the Cx- on the bottom.
- Pin 29 isn't exactly related to the serial bus pins, so move it away.
- Because there won't be a single footprint, empty this field and add both relevant footprint names into the FPFilter section of the symbol.
- All pins must be present on the symbol so the footprint will be matched in KiCad. NC pins (4-13) must be invisible and follow all instructions in KLC for NC pins.

All of that being said... thanks for your contribution to KiCad and you time and effort to finish this symbol!

evanshultz:
Oh, and one more thing. If the footprints are different only by a suffix (like `_Backlight`), you will only need one entry for the FPFilter. The footprint field will still be empty.

Misca1234:
**This is a confusing thread. First, in this post you use the word "schematic". I believe you mean "symbol".**
In the main window, the editor used to create and edit symbols are called "schematic editor" so i guessed it 
is some internal langugae use within KiCad.

**You also wrote "background pins visible" but I think you mean "backlight pins visible".**
Yes 

**Next, I believe this datasheet is the correct one. It helps me to have all information together.**
Yes, that is the correct one, that is used as a base for the symbol/schematic

> Now for the conceptual/architectural discussion:
> All pins for the backlight (1, 2, 3, 14, 15, and 16) should be visible.
> The KiCad user can mark unused pins as NC on the schematic if the user does not want the backlight function.
> In KiCad, symbols are universal and can use multiple footprints in this case. 
> So that leaves only options 3 and 4 above. And since a footprint can only link to a single 3D model, 
> that leaves option 4. Just as @suzizecat suggested.
> 

Jee, that is probably best approach



> And review of the part:
> Pins 18-21 are the power supply for the LCD. 
> Those should not be on the side. Move the VBx+ pins to the top and the VBx- pins to the bottom.
> 

I dont agree here, those pins are for an internal volatage converter, and they will always connect to 
two capacitors so, if they are placed like this, the cpapacitors will fit smoothly and nicely be next to the symbol/schematic.


> It's unclear to me what pin 17 is for. What does VLCD do? 
> I find a few schematics using this LCD on a Google search, and all but one of them show an RC to ground from pin 17.
It is called "POWER LC DRIVER" so I guess it the capacitor feed for an LC filter.

> Separate the VDDx pins from the Ax+ on the top, and the VSSx pins from the Cx- on the bottom.

Dont agree, see above

> Pin 29 isn't exactly related to the serial bus pins, so move it away.
They are grouped together because these pins will most likely be the only one a user will connect so it 
will look smooth with a "bus" there.

> 
> Because there won't be a single footprint, empty this field and add both relevant footprint names into the FPFilter section of the symbol.
> All pins must be present on the symbol so the footprint will be matched in KiCad. NC pins (4-13) must be invisible and follow all instructions in KLC for NC pins.

Pins 4-13 does not exist on the physical device.


> All of that being said... thanks for your contribution to KiCad and you time and effort to finish this symbol!

Np, it is fun

Misca1234:
There are a lot of "cats" in this kicad community, so in honor to those, I have changed my profile image accordingly

evanshultz:
@Misca1234 
Thanks for the explanations. I would still like to separate VDD/VSS from the A/C pins on the top and bottom since they're unrelated functions (power and optional backlight). Having them unbalanced isn't a problem; that's done on a number of other symbols.

Change VBx+, VBx-, and VLCD to PASSIVE since they go to passive parts. Here is the current pin table:
image
image

Can you please provide a link to the footprint for verification? I don't see the footprint in Displays.pretty and I want to check to be sure the pinout matches.

Thank you!

Misca1234:
Hi

The footprint is here 
https://github.com/KiCad/Displays.pretty/pull/9

Something like this ?

bild

poeschlr:
@evanshultz care to take another look at this?

evanshultz:
@poeschlr Thanks for asking. I'm good to merge if you are. The symbol is a bit cramped, but that's subjective and doesn't hurt functionality. I don't think we can know if SPI or I2C will be more popular, but no matter what we have a compromise in one way.

poeschlr:
In this case: @Misca1234 Thanks for your contribution and thanks @evanshultz for the review.

Misca1234:
Thanks

PR #1324, opened 2017-06-19
New component TPS74401_VQFN in regul library

jjuyon:
Here is the new component TPS74401_VQFN.
From the old pull request #1254


poeschlr:
Sorry for the late review. You need to merge the master branch back in there again.
(If travis again shows changes to other parts you can ignore it. I will use diff locally to make sure you really only changed your symbol as compared to the master branch.)

The symbol it self is ok. The naming conforms to the current regul lib but not to the current klc. (But we are discussing the naming convention anyways. https://github.com/KiCad/kicad-library/issues/1398)

jjuyon:
Ok merge is done.

poeschlr:
Thanks.

PR #1325, opened 2017-06-20
NXP ARM: tidy up LPC2148

hackscribble:
Renamed LPC2148 to LPC2141FBD64 with four aliases:

```
LPC2142FBD64
LPC2144FBD64
LPC2146FBD64
LPC2148FBD64
```

Simplified symbol.  Fixed KLC errors.  

One KLC warning is left: suggesting that `VREF` should be an input rather than a power input. I will change this if librarians think it should be.

Note: datasheet says GPIO 0.31 is output only.

Datasheet:
http://www.nxp.com/documents/data_sheet/LPC2141_42_44_46_48.pdf

image
image
image


poeschlr:
Great work. Thanks for contributing

PR #1331, opened 2017-06-22
added LT3080

ilya-epifanov:


ilya-epifanov:
image

poeschlr:
This symbol is only compatible to the 5 lead package versions. Give this information in the description and add the footprint filters.
(show it also in the naming of the part. You might need to use aliases here)

maybe move set toward the bottom? (It is comparable to the adj pins in regul)

ilya-epifanov:
@poeschlr fixed

poeschlr:
For components with multiple power outputs (or even normal outputs) that need to be connected on the pcb, You should not give all of the pins the electrical type power output. ERC will complain if the user connects them together.

The solution is to give all but one of these pins the type passive.

Even better hide these other pins and stack them with the power output pin. 
Be careful here. The symbol editor does not really like this. If you edit the properties of a stacked pin, all other pins in the stack get the same electrical type and name as the current pin. The solution is to finalize everything about the pins before stacking them. Only moving them on top of each other when you don't plan to change anything about them anymore.

ilya-epifanov:
@poeschlr done. BTW, is there a way to make pin 1 to appear on top of pin 9? It appears kicad is displaying them in their pin number order, so 9 is always on top of 1 in preview.

evanshultz:
@ilya-epifanov You will make one pin visible and with the desired type, and then all other stacked pins are PASSIVE type and invisible. In the symbol editor and during preview before placing the symbol all pins will show, but once placed on the schematic only one pin will show up and life is good. :)

poeschlr:
In eeschema there is a button that controls if you see invisible pins or not.
invisible pin

poeschlr:
What is the difference between LT3080EST and its alias LT3080IST (If there is no functional difference i would suggest renaming the symbol LT3080xST and removing the alias.)
If there is a difference make sure the description of the aliases highlights this.

The same seems to be the case for all symbols.

TO220 version: should it be the vertical footprint? (TO_SOT_Packages_THT:TO-220-5_Pentawatt_Multiwatt-5_Vertical_StaggeredType1)

What about LT3080xQ? (The DD-Pak version = TO_SOT_Packages_SMD:TO-263-5_TabPin3)

evanshultz:
I've asked ADI (who bought LT) about the difference between those PNs. Stay tuned.

evanshultz:
It's in the datasheet, but I missed it (and searching for "ist" and "est" didn't help). See note 2 on page 4. "E" grade is tested less so it's cheaper, but they're the same part.

poeschlr:
@evanshultz Not only you missed it. But thanks for investigating.

@ilya-epifanov It seems some other pull request was merged that changed the regul lib in such a way that automatic merging is no longer possible. Could you manually merge the current master branch into this branch? (If you need help here just ask.)

poeschlr:
Could you remove the "I" aliases and replace the "E" in the names with "x"? (Example: remove LT3080**I**ST and rename LT3080**E**ST to LT3080**x**ST)
As we found out there is no functional difference between these parts.

Misca1234:
This push can be closed and deleted

a repush of LT3080 is here
https://github.com/KiCad/kicad-library/pull/1709

SchrodingersGat:
Thanks @Misca1234 for the final changes, and thanks @ilya-epifanov for the contribution

PR #1333, opened 2017-06-22
added IFX25401

ilya-epifanov:


ilya-epifanov:
image

poeschlr:
part naming should reflect that this is the adj version (use the manufacturer part number for that)
could you add the fixed output version as well? In this version the pin 4 should be NC

you have the wrong footprint filters.
(you use tsot and sot 23 but this part is not available in either of these)

The description field seems not quite right.
(same with keywords)

ilya-epifanov:
@poeschlr thanks for the review! I've fixed the points you've mentioned plus one from #1331 about the placement of ADJ (VA) pin. I really don't want to place the ADJ (VA) pin on the left side because it's supposed to be connected between GND and Q. What do you think?

poeschlr:
If you want the TBV and TEV versions in there, then i would suggest you make separate symbols for them and fill out the footprint field. In this case also remove the generic versions. (Yes i see now that the adj version is the one without any special ending in the manufacturer number.)
Or remove the specialized aliases. I would not use both.

More detail, i see two options how we make this symbols.
- option 1, two generic symbols
  - IFX25401-ADJ and IFX25401-FIX
  - both have the footprint field empty and the footprint filters
    - TO?252?4\*
    - TO?252\*TabPin3\*
    - TO?263\*TabPin3\*
- option 2, four specialized symbols
  - IFX25401TBV, IFX25401TBV50
    - footprint field: TO_SOT_Packages_SMD:TO-263-5_TabPin3
    - footprint filter:
      - TO?263\*TabPin3\*
  - IFX25401TEV, IFX25401TEV50
    - footprint field: TO_SOT_Packages_SMD:TO-252-4
    - footprint filter:
      - TO?252?4\*
      - TO?252\*TabPin3\*

ilya-epifanov:
@poeschlr done. I didn't update screenshots because there are no visible changes.

poeschlr:
Your symbols look good. But we have a merge conflict.

Could you manually merge the current master into this branch? (Something was merged that does not allow auto merging.)

ilya-epifanov:
@poeschlr rebased onto master, you can merge it anytime:)

poeschlr:
Thanks for contributing.

PR #1335, opened 2017-06-22
added LM27762

ilya-epifanov:


ilya-epifanov:
image

poeschlr:
Thermal pad should be pin 13 (to match other similar symbols)

You linked the wrong datasheet in the symbol properties.
Set the footprint field and footprint filter.

C1+/- should be passive.
FB+/- should be input
pgood is an output or more precise an open drain output (use open collector)

I would move CP towards Out- (it is the negative unregulated output)

Misca1234:
This push can be closed and deleted,

LM27762 have been re pushed here
https://github.com/KiCad/kicad-library/pull/1708

PR #1339, opened 2017-06-22
Schematic ads7830

Misca1234:
ADS7830
http://www.ti.com/lit/ds/symlink/ads7830.pdf

bild


Misca1234:
Removed the footprint 

bild

Misca1234:
Made the symbol centred along the Y-axis 

bild

poeschlr:
It seems this pull request would also change other stuff than what is intended. (Some other files change as well.)

Misca1234:
It comes from the master,  I did a 
git pull 
and had to merge, shall I redo the branch from start ?

poeschlr:
Ok i checked it with git diff and it looks like you really only changed the ac-dc lib (This is mainly a note to myself such that i know what i already checked.)

The symbol looks good.

The only problem i could find is the missing footprint field (This component only exists in one package)
It should be: `Housings_SSOP:TSSOP-16_4.4x5mm_Pitch0.65mm`

The footprint filter should not include the pin count. I would suggest:
`TSSOP*4.4x5mm*Pitch0.65mm*`

Misca1234:
It should be ok now

poeschlr:
Jup looks good now.
Good that you realized this symbol was in the wrong lib. I think i would not have noticed this.
Thanks for your contribution.

Misca1234:
Thanks

PR #1340, opened 2017-06-22
Schematic max5813 14 15

Misca1234:
MAX5813, MAX5814 and MAX5815
http://datasheets.maximintegrated.com/en/ds/MAX5813-MAX5815.pdf

bild


Misca1234:
Moving to adc-dac library

PR #1341, opened 2017-06-22
Txb0104

Misca1234:
TXB0104
http://www.ti.com/lit/ds/symlink/txb0104.pdf

bild

bild


Misca1234:
I guess it is the same here as in the 
https://github.com/KiCad/kicad-library/pull/1344

I should remove the aliases as well ?

poeschlr:
RGY version should have an exposed pad.

Part naming: Remove the R from the end 
(For example there are multiple order codes for SOIC packaging. They are: TXB0104D, TXB0104DG4, TXB0104DR and TXB0104DRG4.)

You have one symbol per footprint, fill out the footprint field.

ZXU should maybe have an alias GXU.
(or mark the Z/G as don't care)

Misca1234:
Hi

Added pin 15 to the RGY and called it Pad
I removed the R suffix 
Specified the footprint for each of them
Added Aliases (including the GXU)

bild

poeschlr:
Is there an electrical difference between for example TXB0104D and TXB0104DG4. (I can't find any mention of the differences between these versions. Neither in the datasheet nor on the ti webside.) If not i would suggest to remove the aliases.

The difference between TXB0104D and TXB0104DR (or TXB0104DG4, TXB0104DRG4) is that the R version is supplied on reel and the version without R is supplied in a tube.
I would suggest to remove these aliases. (If we would include such aliases for all components in the lib we would have a much larger lib.)

This should be the same for other packaging options.

Misca1234:
Oh, I thought with your comment 
"(For example there are multiple order codes for SOIC packaging. They are: TXB0104D, TXB0104DG4, TXB0104DR and TXB0104DRG4.)"

that i should add all the aliases, I remove them

poeschlr:
Sorry for the confusion.

Another question: do you know what the difference between gxu and zxu is? (Because the z and g are not at the end of the part name we either need to include the second one as an alias or maybe use a place holder there.)

Misca1234:
I do not know either the answer on your question, shall i add the gxu as an alias i nthe zxu ?

poeschlr:
I would suggest to add GXU as an alias. (I think this might give a better user experience. TI does a very bad job to detail the differences between the order numbers of this device.)

evanshultz:
I posted [this question on the TI E2E forum](https://e2e.ti.com/support/interface/voltage_level/f/391/t/607023).

Misca1234:
Nice, let us see what they say.
This component is made in way to many variants and odd sizes, 
some middle manager at Texas must have got hybrids at some point

Misca1234:
They made a response on 
https://e2e.ti.com/support/interface/voltage_level/f/391/t/607023

> You are correct. At one point they were both available, but now the ZXU is the available orderable
> package. They shared the same pinout as shown in the datasheet and the ZXU package landing 
> pattern is provided in the mechanical drawing at the end of the datasheet.

This should mean that GXU is no longer aviable, should be as an alias just in case someone have 
one in their drawer marked "GXU" ?

poeschlr:
If it is no longer available i see no reason to add it as an alias.
(It a user really has such a part, they will open up the datasheet and see that it is equivalent to the ZXU version.)

Thanks @Misca1234 for this contribution. Thanks @evanshultz for contacting TI.

Misca1234:
Great thx

PR #1343, opened 2017-06-22
Pic16 l f18344

Misca1234:
PIC16F18344
all four formfactors
PIC16F18344-P
PIC16F18344-GZ	is an PIC16F18344-P alias
PIC16F18344-SO	is an PIC16F18344-P alias
PIC16F18344-SS	is an PIC16F18344-P alias
http://ww1.microchip.com/downloads/en/DeviceDoc/40001800B.pdf

bild

PR #1311, opened 2017-06-13
NXP ARM: tidy up LPC11U12 and aliases

hackscribble:
Tidying up the LPC11Uxx (LQFP48 7x7mm) microprocessors in the `nxp_armmcu` library.

Simplified the existing symbol for LPC11U12FBD48 and fixed a KLC error.  

In line with other recent PRs, renamed it to LPC11U12FBD48-201 and changed the aliases to cover the following eleven devices:

```
LPC11U13FBD48-201 LPC11U14FBD48-201 
LPC11U22FBD48-301 LPC11U23FBD48-301 LPC11U24FBD48-301 LPC11U24FBD48-401
LPC11U34FBD48-311 LPC11U34FBD48-421 LPC11U35FBD48-401  LPC11U36FBD48-401 LPC11U37FBD48-401 
```
image

Datasheets:
http://www.nxp.com/documents/data_sheet/LPC11U1X.pdf
http://www.nxp.com/documents/data_sheet/LPC11U2X.pdf
http://www.nxp.com/documents/data_sheet/LPC11U3X.pdf

PR #1344, opened 2017-06-22
Pic16 l f18346

Misca1234:
PIC16F18346
all four formfactors
PIC16F18346-P
PIC16F18346-GZ	is an PIC16F18346-P alias
PIC16F18346-SO	is an PIC16F18346-P alias
PIC16F18346-SS	is an PIC16F18346-P alias
http://ww1.microchip.com/downloads/en/DeviceDoc/40001839B.pdf

bild


Misca1234:
Should I remove all variants, that is PIC16F18346-P, PIC16F18346-GZ, PIC16F18346-SO and PIC16F18346-SS 
and only have one variant, PIC16F18346, because the footprint  filter is in PIC16F18346-P anyway ?

poeschlr:
The aliases are ok. 
But i would shorten the pin names. (Remove the functional descriptions and use the main name only. Example RA5, RA4, ...)
Have a look at pull request https://github.com/KiCad/kicad-library/pull/1311 to see what i mean.

Misca1234:
Nice, I wanted to do that (otherwise they become so large) but when I looked at the already approved ones they had the full pin descriptions so I did the same.

Misca1234:
Now it looks like this 

bild

poeschlr:
The footprint filters should not include pin numbers.
It should match dimensional information though.

Also the uqfn version has different pin numbers compared to the other versions. (you need a separate symbol for this version.)
Maybe make the symbol atomic. (add the 4 symbols you where previously aliases and set the footprint field to the appropriate value.)

The Rxy pins should be bidirectional. (i assume these are gpio pins)
Why is RC6 a power input pin?

Misca1234:
Ohhh, you got an falk eye, what a practical move of Microchip to change the pin numbering for one of the capsules.

The schematic have been separated into four schematic
PIC16F18346-P  PDIP 
PIC16F18346-SS SSOP
PIC16F18346-SO SOIC
PIC16F18346-GZ which is the UQFN
In order to be able to specify a specific footprint for each of them

If I remove the pin number form the foot print filter it will look like this for the PDIP 
DIP*7.62mm*
and don't specify a specific footprint it means it will match almost ALL PDIP footprints


All Rxx pins are now bidirectional (including RC6) 

bild

bild

bild

bild

PS
If this is ok I will do the same for PIC16F18344
https://github.com/KiCad/kicad-library/pull/1343
DS

poeschlr:
The pin count is matched by a separate kicad filter. (In cvpcb the one with the # sign i think.)

Misca1234:
Hmm, I have been searching through a couple .lib files and did not find the # sign for the footprint filter.

I notice there are a lot of work to do among the existing schematics, for example PIC16(L)F1459-I/P will match all DIP foot prints.

Perhaps we should keep these four (PIC16F18346) as they are now?

poeschlr:
Sorry to confuse you. This button is what i meant:
screenshot from 2017-06-24 15-01-25
It toggles the option to use the number of pins in the symbol to filter which footprints will be listed.

poeschlr:
I checked the symbol and it looks good now.
Thanks for this contribution.

Misca1234:
Aha, ok, nice to know.

Thx, I will fix the PIC16F18344 in same way

PR #1288, opened 2017-06-05
NXP LPC82n: updated / added symbols

hackscribble:
Tidied up existing symbol for LPC822M101JDH20. Added alias for LPC824M201JDH20.

image
image

Created symbol for LPC822M101JHI33 with alias for LPC824M201JHI33.

image
image

Datasheet:
http://www.nxp.com/documents/data_sheet/LPC82X.pdf

PR #1345, opened 2017-06-22
NXP ARM: add two aliases for LPC83x

hackscribble:
Added two aliases that I forgot to include in this PR: https://github.com/KiCad/kicad-library/pull/1288

Datasheet: http://www.nxp.com/documents/data_sheet/LPC83X.pdf


poeschlr:
Thanks for this contribution.

PR #1347, opened 2017-06-23
Add STM6600/6601

suzizecat:
Added the `STM6600` and `STM6601` (alias) into `Power_Management.lib`
Please note that Vref is a 1.5V precise OUTPUT reference, so don't mind Travis on this 😉 
I'll provide the footprint in the afternoon.

Here is the DS : http://www.st.com/resource/en/datasheet/stm6600.pdf
Here is the symbol : 

image

Best,
Julien


suzizecat:
Footprint provided

poeschlr:
Should vref be of type power out?

Are you sure RST, INT, VCC_LO and PB_OUT are open emitter?
(open emitter would be open source. open drain is the same as open collector)

suzizecat:
Vref isn't a power output, it just give a voltage reference and can't supply current.
You're right on open emitter pins, i must have misread something 😅 

I'm pushing fix

poeschlr:
Thanks for your contribution

PR #1350, opened 2017-06-24
added TLE2426

ilya-epifanov:


ilya-epifanov:
image
image

poeschlr:
for the pdip/soic symbol you need to include the NC pins. (electrical type not connected, hidden, Do not stack these pins)

The 3 pin version has the wrong pin numbers. It should be: 1 = out, 2 = common, 3 = in

I would suggest you create 3 symbols with the following names (no aliases. The only difference in between C,I,M is the temperature range and the R, G4 versions differ only in the packaging information.)
- TLE2426xD
  - 8 pin symbol (current TLE2426 symbol plus 4 NC pins)
  - footprint field: Housings_SOIC:SOIC-8_3.9x4.9mm_Pitch1.27mm
  - footprint filter: SOIC\*3.9x4.9\*Pitch1.27mm\*
- TLE2426xP
  - 8 pin symbol (current TLE2426 symbol plus 4 NC pins)
  - footprint field: Housings_DIP:DIP-8_W7.62mm
  - footprint filter: DIP\*W7.62mm\*
- TLE2426xLP
  - 3 pin symbol (corrected pin numbers from the current TLE2426xLP)
  - footprint field: TO_SOT_Packages_THT:TO-92_Inline_Narrow_Oval
  - footprint filter: TO?92\*

evanshultz:
@ilya-epifanov Please respond to this case. You were given good feedback above which you need to implement so this case can be merged.

Misca1234:
This push have been replicated here
https://github.com/KiCad/kicad-library/pull/1650

This push can be closed and deleted

PR #1353, opened 2017-06-25
Mic5504  ym5

Misca1234:
This is a SOT-23 based LDO regulator
The base is the MIC5504-3.3V
http://ww1.microchip.com/downloads/en/devicedoc/mic550x.pdf
and the aliases
MIC5504-2.8V
MIC5504-2.5V
MIC5504-1.8V
MIC5504-1.2V

bild


SchrodingersGat:
As these devices only have a single compatible footprint (SOT-23-5) please add that as the default footprint.

Otherwise good to merge :)

Misca1234:
Like this ?
(It will "miss" the hand soldering variant of the SOT23-5 footprint)


bild

SchrodingersGat:
Exactly! Thanks :)

Misca1234:
awsome, thx

PR #1354, opened 2017-06-25
Added STMPE1600

Misca1234:
Component STMPE1600
http://www.st.com/resource/en/datasheet/stmpe1600.pdf

16 bits IO expander

bild


Misca1234:
This travis requirement on the text size is a little bit annoying, it differ from kicad's default setting and therefore it is necessary to manually edit the .lib file to change the F2 and F3 text field sizes.

SchrodingersGat:
@Misca1234 you can change the default setting in KiCAD to make it easier for yourself.

Misca1234:
In a way, it would be even easier for me if the travis script adapted to the kicad default or if kicad changed the default settings to match travis ^^

Misca1234:
Nice, thx

PR #1357, opened 2017-06-26
Added MAX5813, MAX5814 and MAX5815 in WLP format

Misca1234:
Added MAX5813, MAX5814 and MAX5815 in WLP format

http://datasheets.maximintegrated.com/en/ds/MAX5813-MAX5815.pdf

bild


evanshultz:
@Misca1234 
This is in the dc-dc.lib. Is that correct? This part appears to be a quad DAC and we have adc-dac.lib which is where DACs and ADCs typically live.

Misca1234:
Jee, you are right, i selected wrong library

PR #1363, opened 2017-06-28
ST LPS25HB MEMS Pressure sensor - Second try

suzizecat:
See #1320 for details.


poeschlr:
I think the interrupt pin should be next to the other communication pins.
(It signals to the communication partner that data is ready on either i2c or spi.)

I would put them on the right side of the symbol. (The input to this devise is the physical signal, output is the converted signal.)

suzizecat:
Is this way looking better ? I personally don't like not having any pin on a side of a component...
image

poeschlr:
Did you not push your changes? There is no new commit. (or did you wait for input?)

@suzizecat  wrote:
> I personally don't like not having any pin on a side of a component

In this case i think the user will connect all pins to the same micro controller. It makes more sense to have them together. (Left or right does not really matter because the user can always mirror the symbol.)

From a logic point of few the (interesting) dataflow is from sensor -> micro controller. (Yes there might be some configuration and control data flow in the other direction but i think the measurement data is what really interests the user.)

suzizecat:
I was waiting for a reply.
Ok, i'm mainly using hierarchical sheets so i just put hierarchic pins and don't mind the side where the pins are.

Let's commit !

poeschlr:
Looks great. Thanks for your contribution.

PR #1365, opened 2017-06-29
Fixed SW_DPDT_x2 pin numbering and filter

Ratfink:
Unit B had incorrect numbering, preventing the symbol from being usable with real DPDT switch footprints.  This gives the symbol the correct pin numbering so that no pins are shared between units.

Additionally, I added a footprint filter to the same symbol.
Fixed SW_DPDT_x2 symbol


jkriege2:
Thanks!

JAN

PR #1372, opened 2017-07-01
add stc89c52

zhangjun1996:


poeschlr:
This symbol is only compatible to one package (footprint). For this reason the footprint field should be filled out with the correct footprint.

Move vcc and gnd pins to the top/bottom
group the pins logically not by the position on the package

The "~" character can be used to put a bar above text in a pin name. If you want it only above part of the text, then put the text where you want a bar above between "~". Example: P3.6/~WR

Where in the datasheet is the information that the RST, PSEN, ALE and EA pins are part of P4.x?

poeschlr:
Thanks for all your fixes. I think there was a misunderstanding about the pin placement.
The GND pin should be on the bottom. The xtal pins should be on the side.

Could you answer my question about P4.x? Where is it stated that this port even exists? (The datsheet calls these pins RST, PSEN, ALE, EA without the P4.x part.)

Move the value field outside the symbol (Either to the top or bottom of the symbol)

Please set the footprint field to: `Housings_DIP:DIP-40_W15.24mm`
And the footprint filter to: `DIP*W15.24mm*` (The footprint filter should not include the pin count this is matched by a different filter that can be turned on/off separately in cvpcb.)

poeschlr:
@wuji2015 could you take a look at my suggestions and finish this pull request?

SchrodingersGat:
@zhangjun1996 we are moving the symbols to kicad-symbols for the new software release. Please resubmit a PR to https://github.com/kicad/kicad-symbols

Thanks

PR #1374, opened 2017-07-01
Schematic lcd 016 n002 l

Misca1234:
This is the updated symbol for LCD-016N002L1
See
https://github.com/KiCad/Displays.pretty/pull/13

bild


Misca1234:
Needed to update the LCD16X2 as well, becosue it used LCD-016N002L as an alias

Misca1234:
LCD16X2 looks like this now
Needed to update the LCD16X2 as well, becosue it used LCD-016N002L as an alias
It will get an overhaul later when this branch is in

bild

poeschlr:
If pin 15/16 are really for leds, the electrical type should be passive not power input.
(Same for A1,K1 and A2,K2)

LCD-016N002L1:
The Dx pins are named DBx in the datasheet.

Misca1234:
> If pin 15/16 are really for leds, the electrical type should be passive not power input.
> (Same for A1,K1 and A2,K2)

You mean "power input not passive" ?

I will make LEDX, AX and KX as power input on both schematics

Misca1234:
This push consist of two devices, the reason for this 
is that the old symbol "LCD12X2" had LCD-016N002L1" as an alias,
I could not just remvoe this alias becosue travis started to scan LCD12X2 (with errors)
so i had to "fix" LCD12X2 as well.
Which I discovered that LCD12X2 actuallay is actually WC1602A 

So I have 
Fixed LCD-016N002L1
Removed  LCD12X2
Added WC1602A

I have modified symbol, footprint, 3D models

LCD-016N002L1
http://www.vishay.com/docs/37299/37299.pdf

WC1602A
http://dl.btc.pl/kamami_wa/wc1602a0.pdf

bild

Misca1234:
Thanks

PR #1375, opened 2017-07-02
Texas TPL0401A-10-Q1 Added

alitekin2fx:
tpl0401x


hackscribble:
Hi @aliteke79 

Thanks for the contribution.  Here are some review points ...

1. **Name**. From the datasheet, it looks like the part comes in `A` and `B` variants, which are identical apart from the I2C address. If so, I suggest renaming your base symbol to `TPL0401A-10-Q1` (the full MPN) with one alias called `TPL0401B-10-Q1`.

2. **Footprint**.  There does not appear to be an SC-70-6 footprint in the KiCad library, but please set the footprint field in your symbol to `TO_SOT_Packages_SMD:SC-70-6`. Also, change the footprint filter to `SC*70*`.

3. **Keywords**. Change the keywords on base symbol and alias to `digital pot potentiometer I2C`.

4. **Graphics**. I like the internal design :smile:. To make the symbol easier to read, maybe the graphic lines should stop short of the pin names, rather than running behind the names.  If they still point at the pin names, I think they will still clearly indicate how the component works.

PS If you would like to add a footprint for the SC-70-6, then please open a second PR on the `TO_SOT_Packages_SMD.pretty` repo.

Many thanks

aliteke79:
Hi @hackscribble, thanks for your comment and suggestions. I have fixed and contributed it again. I will also contribute the SC-70-6 footprint soon.

hackscribble:
@aliteke79 There **is** a footprint already in the KiCad libraries. I missed it - many apologies.  I hope you haven't wasted time recreating it.  

image

aliteke79:
It's okay, i have changed it to TO_SOT_Packages_SMD:SOT-363_SC-70-6.

hackscribble:
I've been comparing the existing `SOT-363_SC-70-6` footprint with the TPL4010A datasheet: http://www.ti.com/lit/ds/symlink/tpl0401a-10.pdf  

The package dimensions and recommended footprint in the datasheet are wider than the existing footprint and wider than SC-70-6 footprints in a quick sample of other manufacturers' datasheets that I looked at.  Texas refer to their footprint as SC-70-6 but also R-PDSO-G6.

Here are the differences ...

image

Left hand side is existing footprint.  Right hand side is R-PDSO-G6 - the pads are wider and positioned further out from the package centre line.  The yellow lines show the range of pin end positions allowed for by the tolerance in the  R-PDSO-G6 package (as per datasheet).

The Texas package just about fits on `SOT-363_SC-70-6` but it is a bit tight at the wider end of the tolerance range.

@jkriege2 do you think `SOT-363_SC-70-6` is a good enough match here, or would it be better to have a new footprint `Texas_R-PDSO-G6`?  We could use the hand soldering variant already in the library but it seems too large.

image

hackscribble:
@aliteke79 just a couple of comments on the latest changes.

The keywords from the base need to be copied to the alias.

In both base and alias, add the common name of the package to the description: `128-TAPS Single-Channel Digital Potentiometer, I2C Interface, SC-70-6`

The `U?` label needs to be moved to the top of the outline.  And, not essential but something to consider, with just single digit pin numbers, you could shorten the pins to 100mil.

image

Many thanks

SchrodingersGat:
@aliteke79  @hackscribble I think it is definitely worth submitting a custom footprint for this .

hackscribble:
@aliteke79 Are you still OK to make the footprint as well?  

To confirm, footprint field in the symbol should change to `TO_SOT_Packages_SMD:Texas_R-PDSO-G6`

aliteke79:
Yes, i will make the changes :).

hackscribble:
The filter should match the footprint name but without including the number of pins, but with wildcards to match alternative punctuation marks and possible suffixes like "Handsoldering".  So I suggest `Texas*R*PDSO*G*`.

Also, the datasheet link used to be in the correct field - the "documentation file name" field in the component properties for base and alias.  In the latest changes, it has moved to the "datasheet" field in field properties.  That needs to be changed back.  It was OK to begin with.

aliteke79:
I have pushed it again.

hackscribble:
Thanks very much for the contribution :smile:

aliteke79:
You're welcome :)

PR #1380, opened 2017-07-03
Fixed pin labels on DRV8825 breakout

matthijskooijman:
The "A" en "B" motor outputs were switched, see https://www.pololu.com/product/2133.


poeschlr:
Thanks for this fix.

PR #347, opened 2015-12-15
Add linear power factor controllers

raiden00pl:
Datasheets:
http://www.linear.com/docs/3968
http://www.linear.com/docs/3129
http://www.linear.com/docs/1119

PR #1384, opened 2017-07-03
Add Linear Tech LT1248/1249/1509

evanshultz:
Inspired by https://github.com/KiCad/kicad-library/pull/347, I just couldn't let it go to waste.

Datasheets:
http://cds.linear.com/docs/en/datasheet/1248fd.pdf
http://cds.linear.com/docs/en/datasheet/1249fbs.pdf
http://cds.linear.com/docs/en/datasheet/lt1509.pdf

Images:
image

image

image

image

There is one KLC violation, but that's on VREF pins. Those are voltage output pins that can and have been used to power other circuitry (including opamps and such). They should be power output.


SchrodingersGat:
Thanks @evanshultz these look great.

PR #1388, opened 2017-07-04
Added RC1602A

Misca1234:
Pimped up RC1602A, both foot print and 3D model

See 
https://github.com/KiCad/Displays.pretty/pull/15

bild

bild


poeschlr:
The k pins are on the wrong position. (I did not check it with a cad program. Only looked at it in the 3d viewer.)
screenshot from 2017-07-08 01-36-11

Misca1234:
Hmm, it looks like this on the freecad drawing

Misca1234:
Yes, you are correct, i see it now

Misca1234:
I looks like this now 

bild

poeschlr:
Looks good. Thanks

Misca1234:
great, thanks

PR #1389, opened 2017-07-04
Change HIP210x gate drivers in intersil.lib to MPN style

evanshultz:
Just as it says. Go from the generic package name (`_SOIC`) to the MPN package suffix (`IB`).


SchrodingersGat:
Looks like Travis is throwing a silly-fit because the symbols got renamed.. That's a new one.

Thanks for the changes.

PR #1390, opened 2017-07-04
Schematic ag12864 e

Misca1234:
Updated the symbol for AG12864E

https://www.digchip.com/datasheets/parts/datasheet/1121/AG-12864E-pdf.php

bild


poeschlr:
The symbol looks good. But the footprint has the mounting holes as pin number 0. (The mounting holes should not have a pin number.) Would you care to fix this or should i do it?

Misca1234:
Np, i fix it

Here is the branch 

https://github.com/KiCad/Displays.pretty/pull/16

poeschlr:
Thanks

Misca1234:
Thanks

PR #1392, opened 2017-07-05
Added SX1273/73 LoRA transeivers

suzizecat:
I Added those two low power tranceivers.

Here is the DS : http://www.semtech.com/images/datasheet/sx1272.pdf
image


SchrodingersGat:
Missing **RESET** pin (6)

suzizecat:
Updated.
I also changed the `RF_MOD/NC` pin from input to output according the DS.
Since there isn't any reference to any use of this pin within the DS, should i put it as a hidden unconnected  NC pin ?

SchrodingersGat:
> Since there isn't any reference to any use of this pin within the DS, should i put it as a hidden unconnected NC pin ?

I don't think so - if it is useful to someone, making it NC and invisible means that they cannot use it.

Thanks for your work on this :)

suzizecat:
Thanks for the review =)

PR #1394, opened 2017-07-05
Fixed pin numbering and KCL warning for 2N7000/TP0610L/VP0610L

matthijskooijman:
All datasheets I could find for these parts showed SGD pin ordering,
while the part had GSD ordering. The related SMD 2N7002 part does have
GSD ordering, which is possibly why this was wrong.

2N700:
  http://www.microchip.com/mymicrochip/filehandler.aspx?ddocname=en571380
  https://www.fairchildsemi.com/datasheets/2N/2N7000.pdf
  http://www.taiwansemi.com/products/datasheet/TSM2N7000K_C15.pdf
  https://media.digikey.com/pdf/Data%20Sheets/ST%20Microelectronics%20PDFS/2N7000,%202N7002.pdf
  https://media.digikey.com/pdf/Data%20Sheets/NXP%20PDFs/2N7000.pdf

TP0610L/VP0610L:
  http://www.vishay.com/docs/70209/70209.pdf

Datsheet excerpt:
selection_003

Fixed part:
selection_004

In addition, I fixed the single KCL warning from checklib.py.


SchrodingersGat:
Thanks for this fix :)

PR #765, opened 2016-10-24
Added Template for Raspberry Pi 40 Pin Models.

ashtonchase:
Reference [Pi Hat Specification](https://github.com/raspberrypi/hats/blob/master/hat-board-mechanical.pdf) for dimensions.

For the mounting hole, I was surprised to not find a mounting hole/screw symbol in the repos already. Let me know if it exists and I missed it.

PR #1395, opened 2017-07-06
Raspberrypi40 pin template

Gansemarsch:
Fixed template per @jkriege2 's comments on #765.

Fixed

- Use a socket strip with F.Fab layer
- Mounted socket on bottom of board
- Fixed pinout
- Removed Extra libraries from .pro and .sch files


SchrodingersGat:
Hi @Gansemarsch looks like there are some conflicting files here. Are you able to fix those?

Gansemarsch:
This may not actually need to be merged anymore as #765 was already merged into master.

PR #1399, opened 2017-07-09
Added new library (driver.lib)

SchrodingersGat:
New library for driver ICs (e.g. pre-gate driver for FET)

First symbol: ZXGD3006E6


jkriege2:
May I suggest a lib-name `gate_drivers.lib` or `drivers_gate.lib` that is a bit more specific (we already have `motor_drivers.lib` and could extend on that naming scheme) ... also it is a plural as for many other libs ;-)

Best,
JAN

SchrodingersGat:
I'm happy with `gate_drivers.lib`. We should start adding library descriptions and categorisations e.g. to README.md

SchrodingersGat:
I have renamed to `driver_gate.lib` in preparation for the new library changes.

jkriege2:
Hi!


just scanning old PRs ... could you add the NC pin 5 to ZXGD3006E6? I think then I can merge ...

Best,
JAN

SchrodingersGat:
@jan done

jkriege2:
Thanks,
JAN

Issue #1211, opened 2017-05-01
many files have a CRLF DOS line ending


jonasstein:
a quick
find kicad-lib* -not -type d -exec file '{}' ';' | grep CRLF
on the kicad-library tree revealed that many files have a CRLF DOS line ending.

Mixing the file endings in large projects can cause trouble and most likely you want the unix line ending standard for all files.

```
$ find kicad-lib* -not -type d -exec file '{}' ';' | grep CRLF
kicad-library/modules/packages3d/Valves.3dshapes/VALVE-MINI_PENTODE_LINEAR.step: ASCII text, with CRLF line terminators
kicad-library/modules/packages3d/TO_SOT_Packages_THT.3dshapes/TO-8-3.stp: ASCII text, with CRLF line terminators
kicad-library/modules/packages3d/TO_SOT_Packages_THT.3dshapes/TO-8-2.stp: ASCII text, with CRLF line terminators
[..]
```

bobc:
Exactly what trouble? Given the fact there are lots of mixed line endings with no reported problems suggests that it is actually not a problem.
I get that some people obsess over details but the line ending "issue" is really not important. As soon as you "fix it" it will happen again, but it still won't matter.

jonasstein:
inconsistent file format can cause bugs, which consume time. Just a question of time. The gitclient is prepared to handle this automatically. https://help.github.com/articles/dealing-with-line-endings/ 
So it is just a misconfigured git client on a Win machine.

bobc:
> inconsistent file format can cause bugs,

Do you have proof for that statement? Programs processing text files are designed to expect different line endings and process them correctly.

I think there is no "question of time", we would have seen issues already if it was a significant problem. I fit hasn't causes a problem so far, it likely never will.

jkriege2:
This was already discussed and yes, line-ending SHOULD be UNIX on the GitHUB-Servers. I think there should be `.gitattributes`-files that give the rules for text-files. But this only works when the files are checked in (once) as Unix, then Git converts them to you local settings and converts them back when pushing.

But for now I would say we should have all the .lib and .kicad_mod-files consistent. About all the 3D models I'm not so sure ...

If you want/have time, you could make a PR to fix this problem ... (maybe start with the .lib/.dcm-files) ... and we go on discussing there.

Best,
JAN

BsAtHome:
The problem files are in the attached text file of .pretty libraries (simple clone and a git status).
[...]

PR #1400, opened 2017-07-09
Fix all line-endings to unix

gauravjuvekar:
https://github.com/KiCad/kicad-library/issues/1211


poeschlr:
Thanks.
I hope this does not break too many pull requests.

PR #1401, opened 2017-07-09
Add linear technologie libbs and LT3555 Led Driver componant

kro007:
Hello,
This is my first contribution to an open-source project.

I added New library for brand "linear technology", wich contain (for now...) one componant.

I have grouped some input (FB,ISP,ISN ...) pin on the right side wich is a KRC violation, to simplefy use in schematic. 

Datasheet here : http://cds.linear.com/docs/en/datasheet/37551fd.pdf
lt3755-qfn

LT3755-2 is an alias of LT3755. QFN and MSOP pinout differ.
LT3755-1 have another pinout and its also depend on package.
So there is 4 componant and 2 alias.

Gabriel.

PR #1404, opened 2017-07-10
Lt3755

kro007:
Hi,
I have close #1401 pull request because it was based on my  master.

I have fixed KRC error, and added LT3755 componant to Dc/Dc lib.

Gabriel.


jkriege2:
Hi!

I just had a llok at your symbols. Thanks for the contribution!

I have a few requests:
1. Please shorten pins to 100mil
2. move REFDES and NAME closer to the symbol, as e.g. here:
2017-08-18 20_49_37-part library editor -- d__kicad_kicad-library-review_library_sensors lib
3. Naming conventions have changed a bit: Please rename `LT3755-1-QFN` to `LT3755xMSE-1`, `LT3755-1-MSOP` to `LT3755xUD-1` ... i.e. we use the manufacturer naming now and escape some characters with `x`
4. please remove the commas from the tags (they should be space-separated)
5. please add apckage to description, e.g. `, MSOP-16` or `, QFN-16`

Best,
JAN

Misca1234:
How do you add the pin type in the screen shoot, that is "input", "Bidirectional" etc?
It is very useful when posting new components.

poeschlr:
New nightly builds show them. It helps a bit. (But you can just post a screenshot of the pin table. This gives the same information.)

kro007:
Hi,
.2 => ok

.3
Reading this : https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention
I don't understand why/how i should rename.

.1
If i reduce pin lenght to 100 mill, negate pin are not well printed : 
capture du 2017-09-07 23-16-25



.4 what's the tags ??

.5 here ???
capture du 2017-09-07 23-24-47

jkriege2:
Hi!

thanks for your changes so far!

1. There are two options: set the OPENLED-pin to normal (remove inverter-circle) but add a tilde `~` in front of the pin-name, this will add a negation-bar above the pin-name, or leave the pin-length at 150mil ... choose whichever option you like better ;-)
3. Please rename as follows:
  - `LT3755-1-QFN` to `LT3755xMSE-1`
  - `LT3755-1-MSOP` to `LT3755xUD-1`
  - `LT3755-2-QFN` to `LT3755xMSE-2`
  - `LT3755-2-MSOP` to `LT3755xUD-2`
  - `LT3755-QFN` to `LT3755xMSE`
  - `LT3755-MSOP` to `LT3755xUD`
  This replaces out the E/I/H temperature-rating with a wildcard `x`(we don't want ALIASes for temperature-ratings), but uses manufacturer-postfix `UD` and `MSE` for the packages.
4. About the "tags": Sorry wrong word on my side, I meant the keywords or "mots clés" in your screenshot. These should be space-separated (see KLC 4.10.iv)
5. yes

Hope that helps, let me know if not and I will try to explain better!

Best,
JAN

PS: Could you please add to the description field a short part explaining the diference between the three variants? (only if that's possible in a few words/numbers and does not blow up the desription too much

kro007:
Hi !
Except that i rename as follow :
LT3755-1-**QFN** to LT3755x**UD-1**
LT3755-1-**MSOP** to LT3755x**MSE**-1

I think it's ok.

jkriege2:
yes you're right ;-)

jkriege2:
Thanks,
JAN

Issue #1406, opened 2017-07-10
Hirose connectors


G33KatWork:
Hi,

Commit cf9ec51825a0a36b1d2495a3f25fd717bfe45081 added a bunch of Hirose connectors which I use all the time on my PCBs. And I just wondered where the 3d models that showed up by default in the past had gone.
I found that 13e8762dd361e8cb05795e032a171ca54e7a090c reverted a whole lot of stuff including the addition of the Hirose connectors in the earlier commit. Was this intended or by mistake? Because I'd really like to see these models to be readded so that I don't have to create my own library with all the models in them.

Thanks,
- Andy

evanshultz:
Hi Andy,

I think you will want connectors at https://github.com/KiCad/Connectors_Hirose.pretty. This library was added to break out the footprints since we had so many connections. Sorry for any confusion.

You'll need to update your fp-lib-table file (https://github.com/KiCad/kicad-library/tree/master/template) to see the new libraries in KiCad. I believe I've seen an automated utility to do it some time in the past, but I can't find it now. If you go to Preferences > Footprint Libraries Manager in Pcbnew you will see the path at the top of the dialog.

Please reply if you run into more trouble!

G33KatWork:
Thanks for your answer. I have the footprints, the 3d shapes are what is missing.
The footprints from https://github.com/KiCad/Connectors_Hirose.pretty even reference the old 3d shapes:

https://github.com/KiCad/Connectors_Hirose.pretty/blob/master/Hirose_DF13-02P-1.25DSA_02x1.25mm_Straight.kicad_mod#L45

evanshultz:
Andy, my mistake. I didn't catch that you meant the 3D models.

@SchrodingersGat It looks like you merged. Can you take a look?

SchrodingersGat:
@G33KatWork you have stumbled upon my shame.

I accidentally merged my personal libraries with the official KiCad libraries. I think perhaps you have been using the 3D models from those? The Hirose 3D models are not part of the official KiCad library dataset, due to licensing issues (all models must be submitted by the creator of the model).

If you are willing to develop these models for us? :) We have a powerful set of scripting tools for generating 3D models. Have a look at the examples in the [packages3d repo](https://github.com/kicad/packages3d)

G33KatWork:
[...]

PR #249, opened 2015-07-14
Added  Hirose DF12 and DF13 3D files (.wings and .wrl)

SchrodingersGat:
Downloaded from http://smisioto.no-ip.org/elettronica/kicad/kicad-en.htm (under CC)

A .pretty repository for the associated footprints is available here - https://github.com/SchrodingersGat/Connectors_Hirose.pretty

mod_conn_df12
mod_conn_df13

PR #1407, opened 2017-07-11
Added Hirose DF-13 shapes

SchrodingersGat:
This PR re-adds 3D shapes for the DF-13 connectors (Hirose)

These were originally [added here](https://github.com/KiCad/kicad-library/pull/249)

However a subsequent commit removed them (unintentionally).

An [issue was raised](https://github.com/KiCad/kicad-library/issues/1406) by @G33KatWork noting this.

These shapes **should not be added** to the packages3D repo as they are 

a)  Not MCAD compatible
b) Not submitted by the designer

They originally were sourced [from here](http://smisioto.no-ip.org/elettronica/kicad/kicad-en.htm)


poeschlr:
Ok the original source is under CC or GPL so we should be good.

poeschlr:
There is a problem with filenames. (A period is missing)
Example:
`Hirose_DF13-02P-125DSA_02x1.25mm_Straight` (3d file name)
`Hirose_DF13-02P-1.25DSA_02x1.25mm_Straight` (footprint file name)

SchrodingersGat:
@poeschlr thanks for spotting. OK now?

poeschlr:
Sorry for my long absence.
The alignment between the footprints and 3d models for the tht versions seems to be wrong.
The 13 pol smd version is missing. (Is this on purpose?)
There are models in this pull request that are not from this series. They do not match any footprint name.

screenshot from 2017-07-25 22-06-26

screenshot from 2017-07-25 22-06-49

PR #1408, opened 2017-07-11
4047+4543

piotr-wiszowaty:


hackscribble:
Thanks @piotr-wiszowaty I will review these tomorrow (no time today, I'm afraid).  I will also close the two PRs that are superseded by this one.

piotr-wiszowaty:
I've already closed the old PRs.

hackscribble:
The datasheet referenced for the 4543 is from NXP and shows only one footprint (SO). If the symbol is meant to represent a generic 4543, would this be a better datasheet? It has DIP, SO, SOIC and TSSOP.

http://www.ti.com/lit/ds/symlink/cd4543b.pdf

hackscribble:
Just a few review points ...

### 4047

1. Set `C`, `R` and `RC_COMMON` to `Input`.
2. Set description to `Monostable/Astable Multivibrator` and keywords to `CMOS monostable astable multivibrator`
3. There should be two footprint filters: `DIP*W7.62mm*` and `SOIC*3.9x8.7mm*Pitch1.27mm*`.

I was about to suggest reducing the height of the symbol by moving the left hand pins closer together.  But I guess you may be allowing space for the external components that go between `C`, `R` and `RC_COMMON`.

### 4543

4. Set keywords to `CMOS 7-segment latch decoder driver`.
5. Reverse the order of the outputs so that `Qa` is at the top.
6. Footprint filters will need to be updated depending on decision about which datasheet / packages (see my previous post).

Many thanks

piotr-wiszowaty:
Changes pushed.

hackscribble:
Thanks @piotr-wiszowaty 

Just one last change.  Please revert the 4543 description to `BCD to 7-segment latch/decoder/driver` and change the *keywords* to `CMOS 7-segment latch decoder driver`.

piotr-wiszowaty:
Changes made.

hackscribble:
Thanks for the contribution.

PR #1405, opened 2017-07-10
Hdsm 541 b hdsm 543 b

Misca1234:
I placed the 3D model into Displays.3dshapes because i guessed that the directory 
"Displays_7-Segment.3dshapes" would be removed and all displays goes into "Displays.3dshapes"

Foot print for 
HDSM-541B and HDSM-543B

Datasheet 
https://docs.broadcom.com/docs/AV02-1588EN

Foot print
https://github.com/KiCad/Displays.pretty/pull/20

Source code
https://github.com/KiCad/packages3D-source/pull/21

3D model in kicad-library
https://github.com/KiCad/kicad-library/pull/1405

3D model in package3D
https://github.com/KiCad/packages3D/pull/83

bild

bild

PR #1409, opened 2017-07-11
Hdsm 541 b hdsm 543 b 2

Misca1234:
Foot print for 
HDSM-541B and HDSM-543B

Datasheet 
https://docs.broadcom.com/docs/AV02-1588EN

Foot print
https://github.com/KiCad/Displays.pretty/pull/20

Source code
https://github.com/KiCad/packages3D-source/pull/21

3D model in kicad-library
https://github.com/KiCad/kicad-library/pull/1405

3D model in package3D
https://github.com/KiCad/packages3D/pull/83

bild

bild


poeschlr:
Looks good for now. We should move the footprints to the 7 segment lib. Maybe we can do this in one pull request (per repo) for this device and the 441/443 (moving the footprint, moving the 3d models and correcting the assignment of the footprint in the symbol.)

Misca1234:
Thanks
For the CA56 I already moved them to new "Displays_7_Segments.lib"
Alright, when we are done pimping up the existing ones we can move the already wrong ones

PR #1410, opened 2017-07-12
Removed deprecated library entries

SchrodingersGat:
- Libraries still exist on GitHub
  (won't break access for users)

issue raised - https://github.com/KiCad/Sockets_MOLEX_KK-System.pretty/issues/1


poeschlr:
Thanks

PR #1416, opened 2017-07-16
relays: add Panasonic JW2 relay

mignon-p:
Added a library symbol for [Panasonic JW2 relays](http://www3.panasonic.biz/ac/e_download/control/relay/power/catalog/mech_eng_jw.pdf), based on AZ850-x symbol.

jw2-schematic


hackscribble:
Hi @ppelleti 

Just one comment.  Please move the whole symbol left so that it is centred about the origin.  This can be done without changing the symbol size.

image

ppelleti:
Thanks!  I have now moved the symbol so it is centered around the origin.

jw2-centered

hackscribble:
Thanks for the continuing contributions :smile:

PR #1417, opened 2017-07-17
added 3d model for USB_Micro-B_Molex_47346-0001

Shackmeister:
Model is drawn by @Obijuan (who gave the permission to use it)
Models is adjusted for Kicad by Shack
usb_micro-b_molex_47346-0001
usb_micro-b_molex_47346-0001-1


poeschlr:
Thanks

PR #1418, opened 2017-07-17
Add MAX II devices

krtkr:
Hi All!

I have generated Altera MAX II devices from pin files downloaded from altera.com.

My generator is available [here](https://github.com/krtkr/altera_kicad_gen).

Package and device information for MAX II is available [here](https://www.altera.com/content/dam/altera-www/global/en_US/pdfs/literature/hb/max2/max2_mii5v1.pdf).

I plan to add whole Altera MAX 10 series next. So once we adjust appear of MAX II you should get MAX 10 in the same appearance.

Here is how it looks like:

EPM240F100:
epm240f100

EPM240M100:
epm240m100

EPM240T100:
epm240t100

EPM240ZM68:
epm240zm68

EPM240ZM100:
epm240zm100

EPM570F100:
epm570f100

EPM570F256:
epm570f256

EPM570M100:
epm570m100

EPM570M256:
epm570m256

EPM570T100:
epm570t100

EPM570T144:
epm570t144

EPM570ZM100:
epm570zm100

EPM570ZM144:
epm570zm144

EPM570ZM256:
epm570zm256

EPM1270F256:
epm1270f256a
epm1270f256b

EPM1270M256:
epm1270m256a
epm1270m256b

EPM1270T144:
epm1270t144a
epm1270t144b

EPM2210F256:
epm2210f256a
epm2210f256b

EPM2210F324:
epm2210f324a
epm2210f324b


poeschlr:
Just from looking at your screenshots i have the following suggestions.

Include the "pin number" in the pin name field. (Currently there is no information added by it.)

The symbols look unnecessary large to me. could you move some pins to the left side?
Maybe stack the power pins? (hidden pins should be of type passive. visible of the type power input)

Also set the footprint field. (If the symbols are intended to be used for one package only)

Move the reference and value field to the outside. (reference on top, value on bottom)

The pin ordering looks a bit random to me. Could you introduce some order here? (Or enlighten me how you got your current pin ordering.)

Have a look at the recently updated symbols in the nxp_armcpu library for inspiration. (Yes these devices are a lot smaller) Example pull request: https://github.com/KiCad/kicad-library/pull/1209

krtkr:
Good day!

I will be able to get back to this patch later on. For now I will answer some of your questions:

> Include the "pin number" in the pin name field. (Currently there is no information added by it.)

Sure, I will do it.

> The symbols look unnecessary large to me. could you move some pins to the left side?
Maybe stack the power pins? (hidden pins should be of type passive. visible of the type power input)

Currently I have a single unit per IO Bank. I see two options here:
 1. Distribute IO Bank's pins between left and right sides. For me it is not obvious how to distribute them.
 2. Place 2 banks on a single unit (one on the left, another on the right). All MAX II devices have either 2 or 4 banks, so most of the symbols will be reduced to a single unit.
 3. Place all banks on a single unit. In this case 324 pins device symbol is going to be huge!

I like 2-nd option more: place 2 banks per unit. What do you think? If it is ok, I will do this way.

> Also set the footprint field. (If the symbols are intended to be used for one package only)

Indeed, symbols are intended to be for one package only.

Footprint is set for TQFP packaged devices. For BGA I left footprint blank while I did not find any BGA in the kicad-library. Did I miss them? I planed to add required BGA housings to kicad-library later on. If you want we can do it within this pull request.

> Move the reference and value field to the outside. (reference on top, value on bottom)

Please, correct me if I am wrong. In this case I have to make all units same size? It seems that for kicad reference and value are per symbol and share same position between units. So to avoid collision between text and symbol body all units have to have same size?

> The pin ordering looks a bit random to me. Could you introduce some order here? (Or enlighten me how you got your current pin ordering.)

Symbol are automatically generated. The pins order for IO comes from list introduces by Altera. Take a look at this [document](https://github.com/krtkr/altera_kicad_gen/blob/master/docs/max2/epm240-io_s.txt), for example.

But we can apply any kind of sort if you would like to. I have ordered VCC* pins.

> Have a look at the recently updated symbols in the nxp_armcpu library for inspiration. (Yes these devices are a lot smaller)

It seems similar to my proposal 2: place two IO banks on a single unit.

hackscribble:
Might The pin order in the Altera documentation be related to BGA escape routing? See discussion here:

https://github.com/KiCad/kicad-library/pull/630#issuecomment-312824679

krtkr:
> Might the pin order in the Altera documentation be related to BGA escape routing?

Not sure. It might be related to internal connection of the pins to FPGA matrix. In the table pads are ordered by 'Pad Number Orientation' column. No clue what is it.

I do not see any problem in ordering pins by pin position. Items are automatically generated and we can apply any kind of ordering.

What I would suggest: keep dedicated (JTAG, configuration etc) pins separate from IO.

krtkr:
Hi all,
I have updated the components. What should we do with BGA packages?

poeschlr:
BGA packages should be in https://github.com/KiCad/Housings_BGA.pretty

SchrodingersGat:
@krtkr what's the status of this?

krtkr:
Hey SchrodingersGat,

I did not change MAX II since 20 Jul. Only add MAX 10 devices to my generator. I guess both still need some work to be finished. I am ready to continue.

So, what's next?

1. BGA footprints.
2. ...?

SchrodingersGat:
1. BGA footprints - let's start with that! :)

krtkr:
Hi all!

Seems I have all required footprints for this pull request in new [footprint libraries](https://github.com/KiCad/kicad-footprints) and ready to regenerate all components here with footprints.

Maybe it is reasonable to migrate this pull request to new [symbol library](https://github.com/KiCad/kicad-symbols) instead?

SchrodingersGat:
Yes please! We will soon be closing out all these PRs so please submit a new one to the new library! Thanks :)

Issue #1402, opened 2017-07-10
RFC - Rearranging Symbol Libraries


SchrodingersGat:
Following on from the work by @jkriege2 [here](https://github.com/KiCad/kicad-library/issues/1398#issuecomment-313969813)

| Library Name | Description | Previous Name | Notes |
| ------------------ | --------------- | --------------------- | ------------ |
| **Basic Devices** 
| device.lib | Symbols for simple devices e.g. Resistors / Capacitors / etc. | device.lib |
| elec-unifil.lib | ??? | | | 
| led.lib | LED | device.lib |
| motor.lib | Electric motors | | |
| power.lib | Power symbols (global) | | special symbols library |
| pspice.lib | Spice symbols | | |
| relay.lib | relays |
| switch.lib | Buttons and switches | | | 
| **Connectors** |
| connector.lib | Connectors, general purpose | conn.lib | Could be expanded in future e.g. *connector_usb.lib* / *connector_JST.lib* |
| **Virtual** |
| graphical_symbol.lib | | | |
| logo.lib | | |
| mechanical.lib | | | 
| **Semiconductors (simple)** |
| diode.lib | Diodes | | |
| transistor.lib | Transistors | infineon.lib | |
| thyristor.lib | Triacs, thyristors, SCR, etc | | |
| valve.lib | Valves | | |
| **Displays** |
| display_character.lib | Character displays e.g. 16x2 text display | display.lib |
| display_dot-matrix.lib |
| display_graphic.lib | Pixel-based displays, etc | display.lib |
| display_segment.lib | 7-segment, LED bar displays, etc | display.lib | |
| **Transformers** |
| transformer.lib | General purpose transformer |
| transformer_power.lib | Power transformers | 
| transformer_pulse.lib | Pulse transformers |
| transformer_signal.lib | Small signal transformers |
| **Sensors** |
| sensor.lib | Sensors that do not fit other categories | sensors.lib |
| sensor_audio.lib |
| sensor_current.lib | Current shunts / transducers / etc | sensors.lib / ir.lib / allegro.lib |
| sensor_gas.lib | Gas sensors, various | sensors.lib |
| sensor_humidity.lib | Humidity sensors | sensors.lib |
| sensor_magnetic.lib | Hall effect, etc | sensors.lib |
| sensor_motion.lib | Accelerometers / gyro | sensors.lib |
| sensor_multi-function.lib | Sensors that sense multiple things e.g. temperature / pressure | sensors.lib |
[...]

PR #1420, opened 2017-07-18
ams: add CCS811 indoor air qualtiy sensor

ChristianHirsch:
@SchrodingersGat added CCS811 indoor air quality sensor. I don't know why the rule 4.6 is violated, any hints?
More ams sensors are coming..


suzizecat:
@ChristianHirsch It seems that you used the type `INPUT` for an `SDA` pin (i believe it's an I2C data pin). If this is the case (which Travis assume) you should use the pin type `BIDIRECTIONAL` since you got data in both input and output.

BTW, you should link the datasheet and add a picture of your symbol to help a quick review 😉

ChristianHirsch:
Thanks @suzizecat for your quick reply! I fixed the pin type, no errors here. And thanks for the hints on the reviews, I'll add an pic and the datasheet the next time :)

suzizecat:
You're welcome !
You should add the pic and the DS anyway, since the librarian may want to check if the pins length is right, the position between pins seems good and so on. They may also check your symbol against the actual manufacturer specifications. 
Since it's pretty easy to add, i would strongly recommend you to do so 😉 
(You can add a pic by copy-pasting it into github and the same for the link, just paste the URL and it'll do the job)

ChristianHirsch:
Hi @suzizecat,
here is the image:
image
Here is the link to the datasheet: [http://ams.com/eng/content/download/951091/2269479/471718](http://ams.com/eng/content/download/951091/2269479/471718)
Hope that helps :)

suzizecat:
Thanks =)

It looks fine, but since you should group pins by function ([KLC 4.4.ii](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention#4-general-rules-for-symbols)), IMO it would be better to : 
- put `INT` with I2C pins,
- put `ADDR` near I2C pins,
- put `AUX` near `SENSOR`, with a space,
-  add a space between `RESET` and I2C pins,

And maybe reduce a little the width of the component. 

Good job !
julien

ChristianHirsch:
Thanks @suzizecat, I applied your suggestions:
image
Is this better?

suzizecat:
After looking at it, i find it a little bit squashed in... I often try to make my symbols as narrow as possible but i don't quite like the look... 

Anyway, for me, it look fine, let's wait for an actual librarian to check it out ! 😉

ChristianHirsch:
Hi @SchrodingersGat, is there anything I need to change?

poeschlr:
I think this symbol might be better suited in the sensors library.

This part is supplied in only one package.
Please set the footprint field to `Housings_LGA:AMS_LGA-10-1EP_2.7x4mm_Pitch0.6mm`
and the footprint filter to `AMS?LGA*EP*2.7x4mm*Pitch0.6mm*` (Exclude pin count, include size specification)

Otherwise nice symbol.

SchrodingersGat:
@poeschlr I'm fine with this either staying in AMS library or moving to sensors.lib, given that it's all [going to be rearranged anyway](https://github.com/KiCad/kicad-library/issues/1402).

@ChristianHirsch if you make the changes suggested above, and want to leave it in this library, I'm happy to merge.

ChristianHirsch:
@poeschlr @SchrodingersGat I'll add two more sensors from ams: [AS7276](http://ams.com/eng/Products/Spectral-Sensing/Multi-spectral-Sensing/AS7262) and [ENS210](http://ams.com/eng/Products/Environmental-Sensors/Relative-Humidity-and-Temperature-Sensors/ENS210), and propably some more in the future. I don't mind putting them in the sensors lib, you guys need to tell me :)

poeschlr:
In future we will go away from a manufacturer based library system towards a function based system.
So please put them into sensors.lib

More details about the future see issue https://github.com/KiCad/kicad-library/issues/1402 (The new libs will be created in a new repo. This library will be the legacy lib that will support kicad 4, the new system will be used in kicad 5.)

ChristianHirsch:
@poeschlr @SchrodingersGat fixed the footprint field/filter and moved it in the sensors lib.

SchrodingersGat:
Thanks for the contribution @ChristianHirsch

PR #1421, opened 2017-07-19
Pin 7 on 4029 (Cout) should be marked output instead of input.

DavidGriffith:
Here's a simple fix for a rather serious little bug if you happen to need a CMOS 4029 up/down counter.  Somehow the Cout pin (pin 7) was mistakenly marked as an input rather than an output.


SchrodingersGat:
Thanks for the fix :)

PR #1424, opened 2017-07-20
Capitalize instances of Vcc, Vss, Vdd, and Vref.

DavidGriffith:
Some symbols use Vcc, Vdd, Vss, and Vref.  Others use VCC, VDD, VSS, and VREF.  This leads to hangups of power pins not getting connected, ERC complaints, and lots of frustration.  This commit capitalizes all instances of Vcc, Vdd, Vss, and Vref in the interest of avoiding such confusion.


DavidGriffith:
I'll wait for the #1437 pull to be complete before revisiting this problem, if necessary.

Issue #1429, opened 2017-07-20
Wrong pin numbering - *78M05_TO252


bumcyk:
Pin numbers are inproperly assigned for the following parts:
- KA78M05_TO252
- LM78M05_TO252
- MC78M05_TO252

Function - current pin number - correct pin number
Input - assigned pin 3 - should be pin 1
Ground - assigned pin 1 - should be pin 2
Output - assigned pin 2 - should be pin 3

The part is incorrectly bound with LM1117-x, where the pin numbering is correct.

Sources:
http://www.ti.com/lit/ds/symlink/lm341.pdf
http://www.farnell.com/datasheets/1855685.pdf

evanshultz:
@bumcyk 
Thanks for pointing this out. Yes, indeed, this is an error. I'm quite familiar with both parts. Generic 78xx in DPAK have the same pinout as in TO220.

Are you up for making this change and generating a PR to merge it into the library?

bumcyk:
@evanshultz 
Sure, here you go. I'm new to Github and I don't know yet how to publish changes, so I attach both modified library and the diff file.
[regul.txt](https://github.com/KiCad/kicad-library/files/1163034/regul.txt)
[regul.zip](https://github.com/KiCad/kicad-library/files/1163035/regul.zip)

evanshultz:
I had forgotten that the regulator library has been updated recently by @jkriege2. If you're looking at the library that came with KiCad 4.0.6 you may see something different than in the latest library hosted here.

Anyway... all the `*M05_TO252` parts are an ALIAS of `NCP1117-12_TO252`.

[NCP1117 datasheet](https://www.onsemi.com/pub/Collateral/NCP1117-D.PDF):
![image](https://user-images.githubusercontent.com/1936989/28428681-eaa46ec8-6d2e-11e7-9acf-64b67bd4db62.png)

[MC78Mxx datasheet](https://www.onsemi.com/pub/Collateral/MC78M00-D.PDF):
![image](https://user-images.githubusercontent.com/1936989/28428691-f48f53e4-6d2e-11e7-902d-f4d48dbd30dd.png)

So yep, definitely a different pinout.

Here's the schematic symbol:
```
# NCP1117-12_TO252
#
DEF NCP1117-12_TO252 U 0 10 Y Y 1 F N
F0 "U" -150 125 50 H V C CNN
F1 "NCP1117-12_TO252" 0 125 50 H V L CNN
[...]

PR #1430, opened 2017-07-20
Correct LM78M05 TO252 regulators

evanshultz:
See https://github.com/KiCad/kicad-library/issues/1429#issuecomment-316747708.

Moved these regs from an ALIAS of NCP1117, which has a different pinout, to their own symbol. [ST's LFxx](www.st.com/resource/en/datasheet/cd00000546.pdf) and [TI's LP2950](http://www.ti.com/lit/ds/symlink/lp2951-n.pdf) also share the LM78M05 pinout.

I didn't add any regulators, but there are a number of other voltages (3.3V, 6V, 8V, 10V, 12V, 15V, 24V, etc.) and varieties/manufacturers (UA78Mxx, L78Mxx, NJM78Mxx, etc.) that are compatible.

@jkriege2 I'd be especially grateful if you can take a look at this since you're familiar with this library.


evanshultz:
Since this is correcting an error in the library, can someone please review and merge?

SchrodingersGat:
@evanshultz thanks, looks good to me :)

PR #1436, opened 2017-07-22
Add PIC16(L)F18325-I/SL to pic16mcu library

antoniovazquezblanco:
[Datasheet](http://ww1.microchip.com/downloads/en/devicedoc/40001795b.pdf)

imagen

imagen


SchrodingersGat:
Hi Antonio, thanks for the contribution. A few notes to get started:

1. Prohibited characters - please do not use `( ) / \` characters in part names (as per the [KLC](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention#1-general-rules)). 

**Note: Yes, this goes against a lot of the Microchip MCU parts, but we are working to improve them**

So, I would rename as thus:

`PIC16F18325-ISL`

2. Add LF part as an alias. The LF part has the same pinout, so it can be added as an alias to the part.
3. Add description / tags / datasheet to the part (see other parts in the library)

Let me know if any of the above points require further clarification

antoniovazquezblanco:
Finally! Sorry for the number of commits!
Final result:

imagen

imagen

SchrodingersGat:
Thanks for your contribution

PR #1439, opened 2017-07-23
Added Murata CRE1S0505S3C & CRE1S0305S3C

Matyro:
schematic

I added Muratas low cost isolated DC-DC transformer

http://power.murata.com/data/power/ncl/kdc_cre1.pdf

The footprint is currently located under
https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/15


hackscribble:
Hi @matyro 

Some comments on these symbols ...

1. Please change the footprint field to match the change you made in your other PR (from "...NMA..." to "...CRE...")

2. Change the footprint filter to be `DCDC-Conv*muRata*3C*`.

3. Picking up some more of the key terms from the datasheet, consider changing the description to `3.3V to 5V 200mA DC-DC Converter 3kV Isolation` and the keywords to `isolated isolation dc-dc converter transformer`.

4. When you make changes 2 and 3 above, make sure you apply them to the alias as well as the base component (different input voltage in alias).

5. Move the reference and value to the top of the symbol as shown ...

image

6. You could make the symbol smaller.  However, I appreciate that the current size is consistent with the other Murata converters in the library, and that there will typically be components connected between the output pins, so it may be sensible to leave the current spacing as it is.

Many thanks

hackscribble:
Thanks for the updates.

To be consistent with your footprint PR, please change the footprint field in the symbol to `Converters_DCDC_ACDC:DCDC-Conv_muRata_CRE1xxxxxx3C`.

I made a mistake in my earlier comment about the footprint filter.  Apologies.  It should be `DCDC*Conv*muRata*3C*`.

WIth these changes, it will be OK to merge.

hackscribble:
Thanks for the contribution.

PR #1440, opened 2017-07-23
remove number of positions from terminal block footprint filter

ppelleti:
According to rule 4.12.iii, "Filters should not match the pin count."

Also, filtering for 2pol, 3pol, etc. causes the filter to miss the newly added footprints that use a [different naming convention](https://github.com/KiCad/Connectors_Terminal_Blocks.pretty/pull/11) (e. g. `TerminalBlock_Philmore_TB132_02x5mm_Straight`).


SchrodingersGat:
Thanks @ppelleti

PR #1441, opened 2017-07-24
Add EuroCard 160mm x 100mm outlines with and without holes.

DavidGriffith:
Inspired by @bobc's comment in https://github.com/KiCad/EuroBoard_Outline.pretty/pull/2, here are two simple templates manually converted from from https://github.com/KiCad/EuroBoard_Outline.pretty (which is supposed to be soon deleted).  These two templates are for 160mm x 100mm EuroCard boards with and without corner holes.


poeschlr:
I'm not sure about the mounting hole positions.
I think the x distance between the holes is not quite right.

I found this specification: Do you have a specification that gives your mounting hole distance?
https://www.elma.com/-/media/general-web-content-files/resources/pdfs/us-corporate-dimensions-of-pcbs-d.ashx?la=en&hash=13B1EAF4FF743087F8B7E87BB10A1D20C44C40BD

It also seems the board should have a maximum size of 160x100 and a tolerance of 0.3mm towards a smaller board for both dimensions.
I would make the board 159.85x99.85mm with a +/-0.15mm tolerance for the manufacturer.

screenshot from 2017-07-26 11-21-25

DavidGriffith:
Here's where I got the specs for the holes:

http://www.alliedelec.com/m/d/599e795fb285c14ee0305bd1cd4f01a0.pdf

bobc:
That pdf is a little confusing, but if you look closely the hole offset is
3.57 on one side, and 2.76 on the other (total of 6.36), the holes aren't
symmetric.

ref
https://www.elma.com/-/media/general-web-content-files/resources/pdfs/us-corporate-dimensions-of-pcbs-d.ashx?la=en&hash=13B1EAF4FF743087F8B7E87BB10A1D20C44C40BD
page 2.

DavidGriffith:
I moved the holes on the connector side to comply with the specs.  I also added some lines 2.5mm in from the long edges as "keep outs" to avoid conflict with card guides.

Where is it suggested to make the board slightly undersized from the spec?  I thought the bearing surfaces on card guides were slightly wider than spec to account for this.

poeschlr:
In the pdf you found tolerances are not given. The pdf i found listed these tolerances. (on page 2)
screenshot from 2017-07-27 10-24-29

poeschlr:
Still not 100% correct.

screenshot from 2017-08-03 12-12-08

You can use freecad + [kicad stepup](https://sourceforge.net/projects/kicadstepup/) to import the pcb and then use the [drawing dimension](https://github.com/hamish2014/FreeCAD_drawing_dimensioning) workbench to check your work.

DavidGriffith:
How's this?

poeschlr:
Thanks

PR #1442, opened 2017-07-24
Added NXE2S0505MC

Matyro:
Added Low Cost 2W isolated power supply

http://power.murata.com/data/power/ncl/kdc_nxe2.pdf

You can find the foodprint under:
https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/16

schematic


hackscribble:
Hi @matyro I will review this PR and your related footprint PR.  But I would like to check something ...

@SchrodingersGat This device package has five pins numbered non-consecutively.  Should these numbers be used in the symbol and footprint, rather than 1 .. 5?

image
image

poeschlr:
I vote to use the non consecutive numbers. (The datasheet numbers should be used.)

hackscribble:
Thanks @poeschlr  Also, is it OK to name the unused pin `NA` as per datasheet? KLC 4.7.iii specifies `NC`.

hackscribble:
Many thanks for the further contribution :smile:

PR #1444, opened 2017-07-25
Added Texas 7331

matyro:
Added missing Texas Instrument ISO7331.

http://www.ti.com/lit/ds/symlink/iso7331c.pdf

I copied the design from the existing ones, because i do not have enough time to redesign all of them.

unbenannt


jkriege2:
nice work!
Only one thing remains to change:
The FPFilter should be `SOIC*7.5x10.3mm*Pitch1.27mm*`.

BTW: As that is also in the DS, would you be willing to also add ISO7330?

Best,
JAN

matyro:
unbenannt

jkriege2:
nice work!

Thanks,
JAN

PR #1448, opened 2017-07-26
sensors: add ams ENS210 sensor

ChristianHirsch:
Hi @SchrodingersGat @poeschlr 
added ams ENS210 sensor based on its [datasheet](http://ams.com/eng/content/download/950231/2267959/483138).
image

I'm not really satisfied with the symbol.. do you have any concerns/suggestions?


evanshultz:
@ChristianHirsch 
Pin names should not overlap. You can also stack pins, which would require leaving only one of the GND pins visible and stacking all pins of the same net connection on top of each other (since only one is visible, all others are invisible). If this is done, you'll have room to avoid the crash between "SDA" and "GND".

What about the symbol doesn't satisfy you?

ChristianHirsch:
@SchrodingersGat @poeschlr Anything I need to change?

SchrodingersGat:
Seems that there is a file conflict here - can you please address this? :)

SchrodingersGat:
Thanks :)

PR #1452, opened 2017-07-27
Manxxxx

Misca1234:
Here is the replacement for the old MAN3640A and MAN71A

Beside the old MAN3640A, I added all the MANXXXX from the datasheet

Common datasheet
https://www.digchip.com/datasheets/parts/datasheet/161/MAN3640A-pdf.php

The shape of the digit is not exactly as the datasheet, 
it is reused from another model.

All MANXXXX uses three basic symbols, 3 footprints and 3 3D models

MAN3410A
MAN3420A
MAN3440A
MAN3610A
MAN3620A
MAN3630A
MAN3640A

MAN71A
MAN72A
MAN73A
MAN74A

MAN3810A
MAN3820A
MAN3840A

kicad-library push branch is 
https://github.com/KiCad/kicad-library/pull/1452

Display_7-Segment.pretty push branch is 
https://github.com/KiCad/Displays_7-Segment.pretty/pull/7

packages3D push branch is 
https://github.com/KiCad/packages3D/pull/93

https://github.com/KiCad/packages3D-source/pull/27

bild

bild


poeschlr:
The common cathode versions have common anode in the description (3440A, 3640A, 714A, 3840A)
3440A, 3640A, 714A, 3840A have the wrong footprint assigned. (The CC has no pins where CA has the anode pins, The CA has no pins where CC has its cathode pins.)

Some of the pins you list as NC do not exist in the packages. (This is the reason why you might need more footprints than you have right now.)
For CA digit with dot left and right (example 3410, 3420), the pins 4,5 and 12 do not exist. (You have them as NC)

For the overflow (example 713) the pins 2,4,5,6,12 and 13 do not exist

For the cc (example 3440) the pins 3,5,10 and 11 do not exist

Misca1234:
I strongly think that they have the same physical package for all of them, but some pins are not used.
They claim it is "No pin" but surely it should be NC, why have different physical packages?
Look at the model A (the left most) according to the mid image row, it have only 1 pin

poeschlr:
I just found a webside that has pictures of these displays.
https://picclick.com/MAN73A-7-Segment-LED-Display-RED-Common-anode-General-262447129622.html
They really don't have the pins that are marked as no pin.

Ok i think we should not care to much about these. They are marked as obsolete.
I will take another look at them and merge as is (If you don't want to invest more time into them.)

Misca1234:
Yes, they are very old, fro 2000 if i don't remember wrong, 
No, i really don't want put more effort into them unless you want, 

1) They are very old
2) They are "only" part of my work "fix existing ones" 
3) I want to move on to add  new ones that are more modern and available (when all the currently existing ones in kicad are done)
4) More holes than necessary is better than missing holes

poeschlr:
Ok for me. (Will take a look later today. I am correctly on the wrong pc. Kicad on this pc is not setup for the purpose of checking pull requests.)

poeschlr:
Thanks for this contribution.

Misca1234:
Thx

PR #1455, opened 2017-07-27
Eld 426 xxxx

Misca1234:
Here is the replacement for the old ELD-426x

ELD-426SYGWA
http://www.everlight.com/file/ProductFile/D426SYGWA-S530-E2.pdf

kicad-library push branch is 
https://github.com/KiCad/kicad-library/pull/1455

Displays_7-Segment.pretty push branch is 
https://github.com/KiCad/Displays_7-Segment.pretty/pull/9

packages3D push branch is 
https://github.com/KiCad/packages3D/pull/95

https://github.com/KiCad/packages3D-source/pull/29

bild


poeschlr:
I would suggest to order the pins a, b, ..., g, DP
You have only one footprint so maybe rename the symbol to use wildcards for the options in the manufacturer part number (similar name to the footprint)

Misca1234:
Here you go
I rearrange the pins as well so it start with a at top and then sorted downwards




PS
When the last push is in, I will make a clean up push for Displays and Displays_7_Segments,
hopefully I remeber to change all symbols that have to upper-case letters to lower-case
DS

bild

Misca1234:
Strange, i fetched the master and merged and got some errros from previous models, i check into it

Misca1234:
Branch have been re push in this branch
https://github.com/KiCad/kicad-library/pull/1470

PR #1457, opened 2017-07-27
Major update to Power Integrations library

evanshultz:
I think the diff shows what happened, so I'll just describe it.

Went though the library, symbol by symbol, and made the following changes:
- Stacked D pins and moved other pins on CAP002DG
- If symbol had multiple S pins, stacked pins
- Made all S pins Open Emitter pin type
-  Shrunk body to align with the family (typically because they now all only have one visible S pin)
- Replaced "-" in FPfilter with "?"


evanshultz:
Here's one example.

Original:
image

Updated:
image

So the updated one matches the size and look of ones with a single S pin. Here is a collection of symbols to show the homogeneity in the library:
image

@SchrodingersGat @jkriege2 @poeschlr @hackscribble 
As you can see, now no KLC errors and a nice visual improvement. Merge please.

SchrodingersGat:
Thanks for this fix, this is a great improvement to the library.

PR #1459, opened 2017-07-27
Added Texas PECL Driver/Reciever

matyro:
I added PECL Driver and reciever from Texas Inst.

http://www.ti.com/lit/ds/symlink/tb5d2h.pdf

http://www.ti.com/lit/ds/symlink/tb5r1.pdf

tb5d1
tb5r1


evanshultz:
@matyro 
Please see a number of checks above using the Details link. Please correct those as a first step.

Also, I see that all signal pins are of type Input. The pins on the right should be of type Output.

SchrodingersGat:
Thanks for your contribution

PR #1461, opened 2017-07-28
Ea t123 a i2 c

Misca1234:
Replacement of DISPLAY EA7123-12C placed in kicad-library,
no foot print connected, 

I assume it is EA T123A-I2C
http://www.lcd-module.com/deu/pdf/doma/t123-i2c.pdf

EA T123W-I2C
Is an alias to EA T123A-I2C

kicad-library push branch is
https://github.com/KiCad/kicad-library/pull/1461

pretty push branch is
https://github.com/KiCad/Displays.pretty/pull/23

packages3D push branch is
https://github.com/KiCad/packages3D/pull/98

https://github.com/KiCad/packages3D-source/pull/32

bild


poeschlr:
I think VLCD should be a normal input not a power input.

Why the alias? The only difference is in the color. If there no different 3d model (and therefor footprint) i would simply use one symbol with the name EA_T123x-I2C. In that case also give the footprint and 3d model this name.

Misca1234:
The VLCD have been changed to input,


bild

poeschlr:
Thanks

Misca1234:
Thx

PR #1437, opened 2017-07-23
Complete overhaul of linear library

evanshultz:
The vast majority of the information is at https://github.com/KiCad/kicad-library/pull/1081, and I suggest y'all go there for details. This has been a long saga and doesn't bear repeating here.

The gist of the update is this:
- Updated the entire library, standardizing and fixing and adding parts
- Pulled in one open PR
- Moved some parts from other libraries that belong with the parts in this library

And there you go! Whew!

Notes:
1. I accidentally got this fork's branch out of sync with master for regul.lib when I added the fix at https://github.com/KiCad/kicad-library/pull/1430. I just coped master from the original KiCad lib here so it shows there is a new commit for that lib but really it's the same. There is no diff below.
2. https://github.com/KiCad/kicad-library/pull/1424 will certainly get broken due to this PR because I'm changing so much. However, I believe this is for the greater good.

There is one KLC violation and that is for LM311. It has a pin named GND which is really an open emitter. Conversation about this, and agreement to do it this way, is in the PR linked above.

PR #1462, opened 2017-07-28
Capitalize instances of Vcc, Vss, Vdd, and Vref.

DavidGriffith:
Some symbols use Vcc, Vdd, Vss, and Vref. Others use VCC, VDD, VSS, and VREF. This leads to hangups of power pins not getting connected, ERC complaints, and lots of frustration. This commit capitalizes all instances of Vcc, Vdd, Vss, and Vref in the interest of avoiding such confusion.

This pull was originally submitted as #1424, but canceled in deference to #1437, which has now been merged.


evanshultz:
Hi @DavidGriffith. Thanks for continuing to help KiCad. I've not run across this, so can you give some examples using screenshots or other files? I made a quick example design to try and understand what you're seeing, but I'm not finding the issue. Here's my schematic:
image

U1 has a hidden power pin named `Vcc`. X1 also has `Vcc` but it's a visible pin. And U2's power pin is named `VCC`. I can name the power net `Vcc` or `VCC` without seeing any ERC issues.

Can you show me what to do with my test design, or provide a design of your own, so I can understand the problem(s) you see?

With regards to power pins, we're moving towards only visible power pins (but the work may take some time). So hopefully hidden power pins aren't involved.

DavidGriffith:
Clone https://github.com/DavidGriffith/g3ruh-modem.  Open up the .sch file, ignore the complaint about missing dg_misc, and don't do the rescue thing.  Run ERC and you'll see this:

vcc-trouble

DavidGriffith:
I couldn't get a simpler 2-component demo schematic to show this problem.

poeschlr:
In kicad 4.0.2 i only get one erc error. It is about pin 5 of component u3 not being driven. This is because your 4029 symbol has the electrical type input for the pin Cout.

I manually checked the supply pins of U7 and U3 in pcb_new. They get connected correctly in 4.0.2 (U7 uses capitalized pin names, U3 does not.) So it seems the netlist you created did connect the pins even thought they are capitalized differently.

After that i checked the complaints in your erc report. (I opened your pcb_new file without changing it.)
- pin 14 of U12. It is connected to +5V
- Pin 7 of U7 is connected to GND

Are you using a nightly build?
It might be that nightly build ERC is a bit more strict. (But it seems the netlist generator is not.)

DavidGriffith:
I'm using e797af331

poeschlr:
A short review of the project you linked:

Your project might benefit from a deeper hierarchical structure.
(Every sub unit where you draw a box around it could be placed into a sub sheet.)
Stuff like the zero crossing detector, the rx filter and lock detect have very few connections to the rest of the schematic. (very few hirarchical pins would be needed to connect them to the rest of the schematic if they would be drawn as a hirarchical sheet.) 
It makes the schematic a lot more readable. (You can than look at it in multiple levels of abstraction.)

If you do that you would have enough space to make the power input pins visible and to move the decoupling capacitors near their ics which they should decouple. (Yes this would require you to create your own symbols. But i bet it would be worth it. As @evanshultz already told you, the official lib will hopefully eliminate invisible power pins in symbols other than power symbols.)

poeschlr:
If indeed the nightly build erc is more strict, we could capitalize all invisible power pins the same way. (for now until someone reworks these libs and finally removes this invisible pin nonsense.)

DavidGriffith:
What does the invisible pin stuff mean for working with individual gates?  Will each gate symbol have power pins?

poeschlr:
The idea would be to have a separate unit for the power pins. There is an open issue where this is discussed. I'm on mobile can't link it. (@bobc wrote a script to rebuild the affected libs.)

bobc:
> It might be that nightly build ERC is a bit more strict.

If that is true, it is a regression and should be fixed. Wayne's policy is that all previous designs should work without breakage.

For tidiness, it might be nice to have consistency, but it doesn't bother me. Vss= VSS however you capitalize it. VSS is the most common version currently, 47 libraries use it. 13 libraries use Vss.

jkriege2:
Hmmm ... usually we try to orient the symbols on what is given in the datasheet ... 

I'm not clear about the whle problem:
Is this a problem solely with the connect hidden pins with the same name hack/feature? If so I would be against a "fix", as we try to abandon hidden power-pins anyways, as those are non-obvious and you don't really see what's going on + it is hard to have e.g. differen power-lines for logic (say some 3V3, some 5V, ...).
If it's a problem of net names ... that doesn't have to do with the symbols, right? The user can give any name and is also repsonsible for giving them correctly ...

Best,
JAN

bobc:
Damn browser lost my reply! I couldn't be bothered to type the whole thing again.

tldr; KiCad used to be case-tolerant of net names, post-2015 kiCad is case sensitive, a lot of old symbols now screw up. This is a good and useful bugfix patch, we should merge it regardless of what we plan to do with power pins.

bobc:
I created a simplified test case for this issue (attached) and did some testing.
[test.zip](https://github.com/KiCad/kicad-library/files/1203180/test.zip)

It turns out, whatever JP Charras did in 2015 didn't "stick", that is, KiCad 4.0.6 does not have case-sensitive net names. However, the nightly builds (according to one I tested) DO have case sensitive net names.

There is a quirk in KiCad 4.0.6, ERC generates warnings for "unconnected" global labels, but then connects them anyway.

For example, KiCad 4.0.6 happily connects Vss to VSS, and Gnd to GND. But nightly builds regard these as separate nets, so Vss only connects to Vss, and VSS only to VSS. KiCad nightly warns for global labels (including power labels) but not for local labels that might differ in case.

I am assuming that the behaviour of 4.0.6 is typical of 4.0.x series, maybe it isn't, and the behaviour changed somewhere in 4.0.x series.

Personally, I think this is a regressive change and is a bug, designs from KiCad 4.0.x might break without warning and the user won't realise.

Anyway, the current library symbols do not cause problems for 4.0.x users, but do for nightly builds.

SchrodingersGat:

PR #1466, opened 2017-07-29
reexported Disc caps with FC0.16 due to bug in FC0.17

Shackmeister:
https://forum.freecadweb.org/viewtopic.php?f=10&t=23400&start=10#p182407


hackscribble:
Many thanks for the contributions :smile:

Issue #1358, opened 2017-06-26
More deprecated libraries to hide


evanshultz:
Along with https://github.com/KiCad/kicad-library/pull/1349...

I believe these can be safely hidden (removed from fp-lib-table):
https://github.com/KiCad/EuroBoard_Outline.pretty
https://github.com/KiCad/Filters_HF_Coils_NEOSID.pretty
https://github.com/KiCad/NF-Transformers_ETAL.pretty
https://github.com/KiCad/Transformers_CHK.pretty
https://github.com/KiCad/NF-Transformers_ETAL.pretty
https://github.com/KiCad/Transformers_SMPS_ThroughHole.pretty

https://github.com/KiCad/Sockets_MOLEX_KK-System.pretty has parts that appear to be the same as the KK 6410 parts at https://github.com/KiCad/Connectors_Molex.pretty. It seems that the same physical parts are covered in the KiCad library, but the deprecated library is an older footprint version. Can someone please check me on that?

https://github.com/KiCad/Oddities.pretty has a few footprints for net shorting that may be useful. Go ahead and hide this too if you want, but I'd like to submit a PR for matching schematic symbols and find a nice place for these footprints (after appropriate updating to meet KLC).

Lastly, the connectors/pin headers/terminals/etc. seem overly confusing. Perhaps there's a better way to organize all of those similar parts?

evanshultz:
@SchrodingersGat  @poeschlr @jkriege2 
Forgot to alert you fine folks.

poeschlr:
I'm not sure deleting repos is a good idea.
Because of the current setup of the lib system cvpcb crashes for all users!
https://forum.kicad.info/t/cvpcb-crashes-on-deprecated-footprint-library-files/6872/2

It might be a better idea to just remove them from the fp-lib-tables and let them slowly die. (They can be removed a few releases in the future.)
I would even suggest to recreate the 3 libs that have already been deleted. (maybe only have one dummy footprint in them that is call this_lib_is_deprecated_and_will_be_deleted_soon)

evanshultz:
Uh oh. See the PR I linked above where Olivier only removed them from the fp-lib-tables. That was only my intent, to continue that work. Actually deleting the repo was not my intent, though I see my wording was confusing if one wasn't thinking about continuation from the linked PR.

poeschlr:
Yea but these 3 repos have already been deleted. (This is why i brought it up. Just wanted to make sure it will not happen again.)
[...]

PR #1472, opened 2017-07-30
Add net ties to schematic library

evanshultz:
This is intended as a discussion point. I do not expect this PR to be merged as-is.

As mentioned at https://github.com/KiCad/Oddities.pretty/issues/2, I'm submitting the net ties symbols I've made. While it's understood that these are required, no schematic symbols existed that I could find. For ease of following past discussion, here's a copy of that issues' post by me:

> Tying nets together at a specific point on a board is a key part of any PCB design program, which I wrote up as an issue for the devs at https://bugs.launchpad.net/kicad/+bug/1571930. While I haven't seen any progress on a new/different solution, I strongly believe these footprints should be retained to allow net ties (as mentioned at KiCad/kicad-library#1358 (comment)).

> I've already made matching schematic symbols that I will add to a PR this weekend.

> The name Oddities is awful, though, and something needs to be done about that. I started writing down some ideas about footprint library naming at KiCad/Connectors_WAGO.pretty#1, but I don't have a good idea where they should go. Even Miscellaneous is better than Oddities.

Image:
image

Ideas for discussion:
- Buy-in for this approach: Other PCB design tools have even more elegant ways which work better in some situations, but I think this style can work well for 98% of the situations if executed well.
- Symbol names: Current I think it's dumb to use `I` for the version with 2 pins, `II` for 3 pins, etc. Also, should the symbols use mixed case instead of all capitals?
- Ref des prefix
- Symbol shape: It could be a single line at the top with multiple vertical post, like this:
```
 ________
|  |  |  |
```
- Symbol description and keywords
- Footprint library name and footprint names: As noted above I followed the footprint names but I dislike it. Also, I selected the existing footprints in Oddities.pretty but that certainly will need to change.

@SchrodingersGat @jkriege2 @poeschlr @hackscribble 
Let the discussion begin!

EDIT: Copy contents of my post referenced above.


hackscribble:
> Symbol names: Current I think it's dumb to use I for the version with 2 pins, II for 3 pins, etc. Also, should the symbols use mixed case instead of all capitals?

How about `Net-Tie_2Nets`, `Net-Tie_3Nets`, etc?.

> Symbol shape: It could be a single line at the top with multiple vertical post ...

I prefer the current shapes.

SchrodingersGat:
I prefer the shapes as drawn above, also.

Why not simplify naming e.g. `Net-Tie`, `Net-Tie_3`, `Net-Tie_4`

evanshultz:
@hackscribble Thanks for responding.

@SchrodingersGat What I dislike about that is that there must be at least two nets connected, so having no number or `I` doesn't make sense. I think we should start with at least 2. I prefer Net-Tie_2Nets, Net-Tie_3Nets, etc.

One other idea about shape: the biggest version could be like a "+" sign, starting out like "-" and then adding legs sticking up and down. But I'm fine sticking with the symbols I drew if everybody likes them.

What about ref des prefix? Is `NT` fine?

SchrodingersGat:
I'm fine with `Net-Tie_2` but I don't think the suffix `Nets` is needed.

`NT` seems like a good choice for RefDes.

I kind of like the idea of a cross, it's very compact.

e.g.

image

evanshultz:
OK. I made the changes to date:
- Update symbols names
- Removed the default footprint and made the FPfilter more generic
- Worked on the description and keywords a little bit

Is there a reason to have more than 4 nets tied together? I've not used more than 4, but if there is the above looks ugly. It will have to turn into something like `++`.

Still looking for:
- Agreement with Oliver's symbol above, or other ideas if we can find something better
- Feedback on description and keywords
- Footprint names: I suggest something like `Net-Tie_2_1mm-Rd_THT`, `Net-Tie_3_o.5mm-Sq_SMD`, etc.

I've mostly used this as a schematic device, to be honest. I'm going to make a project and play around with this using net connections, dropping it into shapes and whatnot. More testing would be helpful.

evanshultz:
Any more thoughts on the above?

Since the Oddities library has been removed from all fp-lib-table files at https://github.com/KiCad/kicad-library/tree/master/template, we need a new home for the net tie footprints. I don't see an obvious fit, but if we find a new existing library or create a new one for them (named better than Oddities), I'll raise a PR to add updated and KLC-compliant net tie footprints.

SchrodingersGat:
We should think of a better name that encompasses net-ties and other board-related parts that do not have associated components

evanshultz:
I don't like the cross above because it will often be used to connect to grounds. When there is a pin pointing up that goes to ground it always looks awkward to me. My opinion.

I have an example project with a few permutations. Where can I put it so you all can see it? Is there a place on GH to upload files which doesn't become part of the KiCad repo? Or it is easiest to just put it in one of my repos?

To me, fiducials are similar to this. It's not a physical component but it's part of the board. What about a library called `Board_Features` or something like that which holds net ties and the existing Fiducials and Measurement_Scales libraries. Symbols is also a related library, but it has a lot of footprints already so perhaps it's best to leave separate? But perhaps rename to Board_Symbols so it would be grouped with Board_Features (or whatever name we choose).

What about the remaining questions above? Any thoughts?

@SchrodingersGat @jkriege2 @poeschlr @hackscribble @Shackmeister

evanshultz:
@SchrodingersGat @hackscribble Thoughts on my comment above?

SchrodingersGat:
We could add both variants of the `NT?` symbols, the linear ones and the cross ones. I think that should be fine.

I think we can do better than `Board_Features`, unfortunately I can't offer a better option at this stage :p

jkriege2:
I would also be in favour of having both variants.

As a name, what about `PCB_Structures` or `PCB_Features` or `PCB_Specials` or something like that?
JAN

SchrodingersGat:
Are we thinking that this lib would also contain

* Fiducial marks
* Test point

antoniovazquezblanco:
I would love to make use of net ties in one of my boards (for reasons explained in https://forum.kicad.info/t/erc-power-pin-conflict-resolution/8323/7).

Any progress on library naming? It seems it is the current issue for merging.

Also, is there any way travis can be silenced for generic components that do not require a datasheet?

Thank you guys!

evanshultz:
@antoniovazquezblanco Thanks for bringing this back up!

I've added the schematic symbol for the '+' 4-pin version to this PR. Footprint names (and thus FPfilters) are an open issue, and there's probably a better symbol naming scheme if anyone has ideas.

We still would need to decide on the corresponding net tie footprints and their names. I would vote for having small and large sizes in both SMT and THT for each one. So that would be 4 footprints for each schematic symbol. I don't see the point in a net tie footprint that doesn't have copper between the pads, so perhaps Net-Tie_2_SMD_0.5mm, Net-Tie_2_SMD_2mm, Net-Tie_2_THT_0.3mm, Net-Tie_2_THT_1mm? That would describe a net tie footprint with 2 pins, either SMT or THT pads, and the pad diameter. In add cases copper would connect all the pads. The footprint naming convention would be `Net-Tie_{pin count}_{SMD|THT}_{pad diameter}`. To match with the symbols created above we would need versions with 2, 3, and 4 pads. I'm happy to make up the footprints if we agree on what we want.

@SchrodingersGat Yes, I'm considering a footprint library name to capture elements that are for PCB design/construction but do not have a physical part placed on the footprint. Net ties, measurements scales, fiducials, board targets, layer coupons, test structures, etc.

evanshultz:
@hackscribble @jkriege2 @SchrodingersGat 
Any thoughts on my above comment? How can we wrap this up?

jkriege2:
Generally all fine for me ... I would only suggest to work a bit more on the design. I think we should keep this in mind:
1. pin-numbers are not really necessary, but if they should be there ... we should move pins further apart so they don't intersect, or make the smaller 
2. the REFDES/name should not be where wires will be (e.g. directly below the 4-way cross)!

I made 3 small mock-ups of redesigns keeping tehse points in mind:

2017-11-05 10_04_42-part library editor -- d__kicad_kicad-library-review_library_device lib
2017-11-05 10_03_56-part library editor -- d__kicad_kicad-library-review_library_device lib
2017-11-05 10_07_05-part library editor -- d__kicad_kicad-library-review_library_device lib

evanshultz:
@jkriege2 Good points! I made those changes and made the FPFfilter generic so it will catch the pin number. I also spelled out `Cross` in the 4 pin variant because, even though `+` is legal, it's less descriptive (at least in English). Here are all of them:
image

Can we merge?

jkriege2:
Hi!

I think for the NT-3 we should also have a T-shaped variant???

Maybe one can also make the connecting lines thicker to make them stand out more ...

Best,
JAN

jkriege2:
PS: You can get rid of the datasheet-error-message by setting it to `~`, which is read as "consciously set to empty" by KiCAD

evanshultz:
I fattened all the connecting lines to 20mil and added a "T"-shaped 3 pin net tie:
image

Are we requiring to use the tilde to eliminate the datasheet error? It's no problem, but not done on the discrete parts I see so is this a new KLC requirement?

evanshultz:
Anything else here? Merge as-is, or add the tilde for the datasheet field?

jkriege2:
I would make the name invisible (as we do for graphics) and I would set the datasheet field to `~` to switch off the KLC-error that there is no datasheet.

2017-11-14 07_14_47-part library editor -- d__kicad_kicad-library-review_library_device lib

jkriege2:
... otherwise: Looks good to me. Merge?

SchrodingersGat:
Thanks @evanshultz  :)

antoniovazquezblanco:
Thank you!

evanshultz:
Thanks for merging!

There are some lingering points here:
1. Net tie footprint names? I vote for having small and large sizes in both SMT and THT for each one. So that would be 4 footprints for each schematic symbol. I don't see the point in a net tie footprint that doesn't have copper between the pads, so perhaps Net-Tie_2_SMD_0.5mm, Net-Tie_2_SMD_2mm, Net-Tie_2_THT_0.3mm, Net-Tie_2_THT_1mm? That would describe a net tie footprint with 2 pins, either SMT or THT pads, and the pad diameter. In add cases copper would connect all the pads. The footprint naming convention would be `Net-Tie_{pin count}_{SMD|THT}_{pad diameter}`. To match with the symbols created above we would need versions with 2, 3, and 4 pads. I'm happy to make up the footprints if we agree on what we want.
1. Footprint repo name. This repo should include net ties, measurements scales, fiducials, board targets, layer coupons, test structures, etc. Suggestions are are `Board_Features`, `PCB_Structures`, `PCB_Features`, `PCB_Specials`. I also thought `Fabrication` might be a good word for some of these elements, like `Fabrication_Features`.

SchrodingersGat:
1. `NetTie` not `Net-Tie` to fit the **CamelCase** naming scheme of KLC-3 :)
1. Now that we have a single repository for footprints, we can just have `NetTie.pretty` store all the net ties
1. This way, we can separate out `MountingHole`, `Fiducial`, `TestPoint` without worrying about a gotta-catch-them-all naming scheme

PR #1474, opened 2017-07-30
Add MCP45HVx1 digital potentiometer.

FabianInostroza:


hackscribble:
Hi @FabianInostroza 

To help with review of your PR, could you please post an image of the symbol and a link to the datasheet.

Thanks

FabianInostroza:
Hi, here is a screenshot of the symbols and the datasheet can be found here: http://ww1.microchip.com/downloads/en/DeviceDoc/20005304A.pdf

captura de pantalla de 2017-08-01 00-04-44

hackscribble:
Thanks @FabianInostroza 

@SchrodingersGat Can you advise on symbol and alias naming for these atomic parts?  

Taking two examples of the MPN from the datasheet ...

`MCP45HV31T-104E/MQ`
`MCP45HV51T-103E/ST`

- `31` vs `51`: bit resolution of the pot
- `103` vs `104`: resistance of the pot
- `MQ` vs `ST`: package type.

The PR has two symbols for the different packages, with aliases for the resolution variants.

Q1. Should aliases be added to cover the resistance value as well? That would lead to seven aliases per symbol.

Q2. If not, should the name be changed to full MPN with wildcards?  For example `MCP45HV51T-xxxE-ST`.

poeschlr:
I don't think resolution and resistance are parts of the mpn that are important for the library. We can wildcard them. (The user then can fill out the value field accordingly.)
The T and E are unimporant as well. (page 97 of the datasheet, Packaging and Temperature range)
`MCP45HVxx-xxxx-MQ` or simply `MCP45HV-MQ`

hackscribble:
Thanks @poeschlr 

@FabianInostroza Could you please rename the symbols to `MCP45HV-MQ` and `MCP45HV-ST` and remove the aliases.  I will start to review the other aspects of the symbols.

SchrodingersGat:
Beat me to it @poeschlr ,  we should adhere to the "rule of fewest symbols".

hackscribble:
@poeschlr @SchrodingersGat I need another "consult" :smile: 

The datasheet says of the `NC` pins ...

> Pin not internally connected to die. To reduce noise coupling, connect pin either to DGND or VL [positive digital power supply input]

So I'm thinking the `NC` pins should be of type `passive` and stacked as per KLC 4.3.  

However this would force **all** these pins (there are seven in the QFN package) to be connected either to `DGND` or `VL`.  Alternative would be unstacked `passive`: gives flexibility for PCB routing but keeps the symbol larger than maybe it needs to be.

@FabianInostroza Have you used the QFN version in a layout?  If so, how did you handle the NC pins?

image

evanshultz:
Page 24 of the datasheet linked in the 3rd comment above says this:
image

That will be the bulk of the heat. The NC pins will carry away far less heat since they're not connected to the substrate.

If the layout is easy, forcing pin connections doesn't seem so obtuse to me. Another way to go would be to stack pins 8-10 on DGND and pins and pins 16-20 on VL, since that's easiest for layout. Those pins won't help thermal too much and otherwise you need to get across the device, and around the thermal pad, to connect all NC pins together. Just a thought...

jkriege2:
NC pins should never be stacked! -> KLC 4.3.i

hackscribble:
> NC pins should never be stacked! -> KLC 4.3.i

But in this case they have to be connected :smile: so maybe we have to change the name too?

evanshultz:
Right. If we have no information about NC pins, then the pin should be floating. It may be used for production test, for example. But in this case, we're given explicit instructions of how to connect the pins:

> 3.15 Not Connected (NC)
> This pin is not internally connected to the die. To reduce noise coupling, these pins should be connected to either VL or DGND.

"NC" in this case seems to mean more than there's no logical connection to the pin, but it does not mean the pin should be floating.

jkriege2:
Then I would suggest having them as visible pins, as the DS avctually proposes them to be connected somewhere ... one might alo stack than in this specific example, IFF the DS is interpreted to say that they should always be tied together

JAN

hackscribble:
Just posting an image to visualise @evanshultz's suggestion of grouping the (so called) `NC` pins into two stacked groups for shortest routing to either `VL` or `DGND`.

image

jkriege2:
If I see that correct, the connection indicated in the image would "short-circuit" DNG and V- via the substrate?

Best,
JAN

jkriege2:
I would treat them as any other NC pin (this is IMHO not the only device, where they can be connected to GND to reduce noise ...), i.e. hide them and have them as NC. IFF somebody wnats them connected, they can always unhide and connect them!

JAN

hackscribble:
> If I see that cnorrect, the connection indicated in the image would "short-circuit" DNG and V- via the substrate?

The datasheet says that these pins are not connected internally, so there would not be a short.  But I take your point about treating them as normal non-stacked `NC` pins.

hackscribble:
@FabianInostroza while we resolve the question of the NC pins, there are some other review comments that you could be working on ...

1. Rename the symbols as per earlier comment above.

2. Since this device does not do I2C clock stretching, SCL should be of type input rather than bidirectional.

3. The datasheet is inconsistent about the names of the pot pins.  I would go with the pin table, so the three pins are `P0B`, `P0W` and `P0A`.

4. Small changes to the description and keywords: suggest `7/8-bit single +36V (+/-18V) digital pot, I2C serial interface, volatile memory, TSSOP-14` and `I2C potentiometer pot digital`.

There may be scope to reduce the length of the pins and the size of the symbol outline.  But let's see what the conclusion is on the NC pins.

FabianInostroza:
Hi guys, I have been busy.

@hackscribble I agree with all the points except (1). I agree that the resolution is not big differentiating factor, it seems less relevant that the potentiometer value but think about the users. I believe nobody would say "I'm using the MCP45HV", also if a user is looking for the MCP45HV31 or MCP45HV51 he would not find it, if he finds the "MCP45HV" first would have to check if it is referring to the part he is looking and then check if everything is correct. AFAIK there should not be part numbers in part descriptions so we can't add the part numbers to the descriptions.
I think that by using just MCP45HV you are "simplifying" the file (the computer doesn't care about it) but complicating --a little-- the users life.

hackscribble:
Thanks for the changes @FabianInostroza 

@SchrodingersGat @poeschlr @jkriege2  What are your thoughts on @FabianInostroza's comment about review point (1) https://github.com/KiCad/kicad-library/pull/1474#issuecomment-320469175 ?

jkriege2:
About how many pot-values and resolutions do we talk?

For the fixed voltage regulators we explicitly added all voltages as ALIAS and for ADCs/DACs that are available (with the same pin-out) in different resolutions we also added all resolution-variants (possibly as ALIASes).

I would say, if we stick to this scheme, we should add all variants (resolution and value). At least I would vote for adding the resolution (naming argument is valid IMHO).

Best,
JAN

jkriege2:
A comment on the xx-wildcards: As the string-matching in the symbol selector does not care about these, they are relatively invonvenient, when you try to find a symbol with a specific part-number, as you e.g. have to paste the full number and then delete from the back until you get a match, but then have possibly several matches with different extensions and have to go into the other direction then ... if you know that you have to do it that way that is!

JAN

hackscribble:
> About how many pot-values and resolutions do we talk?

For each package, there are two resolutions x four resistance values, so base symbol plus seven aliases.

FabianInostroza:
So what is the veredict?

SchrodingersGat:
@jkriege2 are you able to provide some resolution here? :)

jkriege2:
Hi @FabianInostroza!

I had a quick look at the symbols. I only found a few things that still need fixing:
1. pins should be shorter: 100mil is usually enough.
2. You could set the pin text offset to 20mil (justa propsal, I feel that looks better)
pinpadtext
3. About the NCs ... maybe here it's best to have them separate, but visible ...

Best,
JAN

FabianInostroza:
Hi @jkriege2 

I made the changes you requested but I'm unsure about the NC pins, do you think this look right
captura de pantalla de 2017-10-22 19-24-01

jkriege2:
Hi!

please move the symbol name to the top-right, the package to the bottom-right or a bit further down and maybe add teh drawing of a small potmeter inside, as done for many other digital pots, see
2017-10-23 07_13_59-clipboard

FabianInostroza:
@jkriege2 Done, but visible NC pins violate KLC

jkriege2:
Hi @FabianInostroza ,

I'm on holidays right now, but will have a look at the weekend or early next week.

JAN

jkriege2:
PS: The errors in visible NCs, we can ignore here as discussed above

jkriege2:
nice work!

Thanks,
JAN

PR #1476, opened 2017-08-01
Added lib-table-entry for HDMI connectors

SchrodingersGat:
Added entries to library tables for HDMI connectors


SchrodingersGat:
Thanks Jan

Issue #1398, opened 2017-07-08
RFC - Symbol Naming Guidelines


SchrodingersGat:
This issue follows on from https://github.com/KiCad/kicad-library/issues/1386

Some references:

* https://github.com/KiCad/kicad-library/pull/1081
* https://github.com/KiCad/kicad-library/issues/1082

SchrodingersGat:
> @evanshultz -  I proposed and pushed for the generic scheme that Jan is professing above for the same reasons. It has the fewest number of symbols and gives the greatest flexibility in matching a given symbol with a footprint. 

Maybe I'm being dumb but I still fail to see how using the `MPN_SOIC` / `MPN_SOT-23` nomenclature actually reduces the number of symbols. The other option is using the full MPN e.g. `MPN-xxyy` to designate symbols, and you have to draw the same number of symbols anyway.

I think we are talking about different things. When I say `symbols` I am talking about the actual graphical drawings - which may have dozens of associated aliases.

SchrodingersGat:
> @evanshultz - Is there a desire by Wayne or someone else to make the KiCad library atomic? If there's a future plan then it might be good to get ready now. The references above don't talk about that and I did hear Wayne mentioning at FOSDEM that there could be plugins for Digi-Key or other vendors in the future. In that case, the KiCad library would only need to be (but wouldn't have to be) generic parts.

AFAIK there is no such push from above to have atomic libs. There is a push from all of us to have libs that are at least consistent and have a solid policy behind them.

To me, a fully atomic library part would have exact order codes, temperature rating, etc. This would quickly bloat the libraries.

Perhaps with the new library format we could support this using "VARIANT" functionality. i.e. we have a part say MCP2551_SOIC that has underneath it all the possible variants.

If we look at the basic requirements for a fully-defined library part, we need the following:

1. A schematic drawing for the part
1. A PCB footprint for the part
1. The pinout must match the footprint exactly
1. The symbol name must indicate the footprint type in some fashion, if options exist

There are some symbols that do not need to indicate the footprint style as there is only a single option. The [ADM3053](https://www.digikey.com/products/en/isolators/digital-isolators/901?k=ADM3053) only has one compatible footprint. This could be added to the library simply as **ADM3053**

Conversely, a symbol such as the [MAX232](https://www.digikey.com/products/en/integrated-circuits-ics/interface-drivers-receivers-transceivers/710?k=MAX232) is available in different footprints (and different manufacturers!) This presents a more tricky case.

[...]

PR #1480, opened 2017-08-02
Add TL494/UC253x/UC384x/UC380x/UCC3808 PWM controllers

evanshultz:
This adds some of the basic PWM controllers into the library. Image:
image

TL494: http://www.ti.com/lit/ds/symlink/tl494.pdf
UC352x: http://www.ti.com/lit/ds/symlink/uc2525b.pdf
UC384x: http://www.ti.com/lit/ds/symlink/uc3842.pdf
UC380x (ALIAS of above): http://www.ti.com/lit/ds/symlink/ucc3800.pdf
UCC3808: Datasheet: http://www.ti.com/lit/ds/symlink/ucc2808-2.pdf


evanshultz:
TL494 has an FPfilter that matches the footprint at https://github.com/KiCad/Housings_SOIC.pretty/pull/26.

jkriege2:
Hi!

a few questions/spotted bugs on these:
1. Why did you choose AC-DC-lib? Wouldn't linear.lib or texas.lib be better (before we reorganize everything)?
2. UC3525, DISCH should be OpenCollector output, I think
3. UC3842_*: FB should be renamed to VFB  and REF should be VREF (as in datasheet)
4. UC3842_SOIC14: CS should be ISENSE RC should be RT/CT
5. When correcting names above, please ensure that the names here stay as they were
6. UC3842_DIP8 does not have a set footpritn field, but several FPFilters? If it is `_DIP8` it should have DIP8 set as footprint and FPFilter accordingly

Other wise this looks fine!

best,
JAN

evanshultz:
@jkriege2 Thanks for the review.
1. There is no suitably-named library for these parts. I'ved used all of them, and often see them, in AC-DC supplies. But being a generic controller they can perform many functions. They are not specific Texas parts.
2. Good catch. Fixed and thank you.
3 & 4. If you look at datasheets from various vendors you will see pin names vary. I used the shorter and more modern UC3802 pin names that are also used on UC3842 [here](http://www.ti.com/lit/an/slua213/slua213.pdf). There pin names and functions are commonly understood even when they vary from IC to IC and from manufacturer to manufacturer.
5. I'm sorry, I don't understand. Can you help me?
6. `UC3842_DIP8` has all the SOIC8 parts as ALIASes. This was so that each package was listed individually, even as they come from multiple vendors with different MPNs, while the SOIC14 package is a unique symbol. The TSSOP footprint is also listed since that's an option for UCC380x. I thought was was the agreement on how to handle things, but perhaps I misunderstood. Can you check https://github.com/KiCad/kicad-library/issues/1398 and let me know if I got it wrong?

Thanks again!

jkriege2:
ad 5: I looked at the datasheets as provided (usually we try to get the pin-naming as in the provided datasheet, so you know what is meant, when you look it up ... if these are available from different manufacturers, one may make an exception here ;-)

About the naming, I'm a bit confused by now (we had too many diverging discussions and changing rules to track them ...). I think the current idea is:
Use the naming for package, as given by the manufacturer. If a symbol is specificly named for ONE package, then it should link that package in the footprint and in the FPFilters. If several packages with teh same pinout are available and you use the per-package-naming, you'll need to make copies of the symbol, as ALIASes (in the current KiCAD) cannot vary in package ...

Best,
JAN

@SchrodingersGat : Could you comment on that?

evanshultz:
@SchrodingersGat @jkriege2 (if you're back)
Could you take a look? What needs to be done to merge this PR?

evanshultz:
@SchrodingersGat @jkriege2 
Can you please take a look at this? Do I need to change anything, or did I do it correctly?

SchrodingersGat:
As @jkriege2 has been dealing with this one I'll let him finish it :)

evanshultz:
@SchrodingersGat Thank you, Oliver. Jan asked for your input above which is why I added you.

SchrodingersGat:
Ah, ok I will read that then :)

SchrodingersGat:
I think @jkriege2 has it correct - do you follow his advice as listed above?

jkriege2:
The footprint problems still persist:
- SG3525, UC3525, UC3527 link 3 packages in FPFilters, but I only see 2 in the DS
- TL494 looks OK
- UC3842_DIP8 and variants links TSSOP,SOIC,DIP packages, but the name sais only DIP
...

Please rewok this PR, so it matches these rules (which i think we all agreed on):
1. if multiple packages with same pinout match, these can be summarized in one symbol (no footprint, but all matching FPFilters), but then the symbol must not link to footprints that are not in the DS (potentially the symbols here will have to be split into some more variants to get that done)
2. if a single package is available, or the packages have different pinouts or you want to go for the full single-package option: Make each symbol, so it links to exactly ONE footprint (set footprint field and one FPFilter to match it), use manufacturer designation for package-type where it makes sense (maybe the `_DIP8`-option is better, when the part is available from different manufacturers that use different standards).

OK?
JAN

evanshultz:
OK. I think I get why I'm confused and confusing everybody. The explanations and changes are below.

SG3525
ST SG3525 (www.st.com/resource/en/datasheet/CD00000958.pdf) shows narrow SOIC and DIP. On Semi SG3525 (https://www.onsemi.com/pub/Collateral/SG3525A-D.PDF) shows wide SOIC and DIP. Microsemi SG352x (https://www.microsemi.com/document-portal/doc_view/11120-sg1525a-sg1527a-datasheet) is available in wide SOIC and DIP. So that is the same part in all 3 packages from various vendors. In that case, I added all three FPfilters so the user can pick their package.

UC352x
TI UC352x (http://www.ti.com/lit/ds/slus376/slus376.pdf) shows wide SOIC and DIP. This I split into a separate symbol with only those two packages.

UC384x
There are `_DIP8` and `_SOIC8` ALIASes. This is how I captured all the package styles separately. The DIP and SOIC versions share the same pinout and so the same symbol. I'm confused what the problem is with this symbol. It seems to me that this meets all the rules we agreed upon (with the change below). Am I still not getting it?

UCC3800/UCC3813
These are in the same 3 package options, so I split them off.

@jkriege2 Does this work? I still think UC384x was done correctly, but there were other parts which could be split off to best separate the symbols. Sorry that I missed those earlier.

evanshultz:
As soon as I woke up I realized you may not consider this finished. Maybe it was too late to think about it last night.

Anyway, since SOIC14 package is split off then all packages may be separated as well. While it would be OK to put a part that only had DIP8 and SOIC8 together, having the SOIC14 package means we may need more granular division of symbols? Is that what you want? Please confirm.

Also, I noticed that TI makes an SOIC16 package as well.

jkriege2:
1. SG3525 is fine for me ...
2. Yes I think it would be best to split by package here ... since this part is available from different manufacturers and possibly with different package-suffixes, you can keep the `_DIP8`,... postfixes. If the package naming (postfixes) is the same for all different manufacturers, please use the manufacturer postfixes.

OK?
JAN

evanshultz:
DIP8 is "N" and "M". SOIC8 is "D" and "Dx" and "DM". SOIC16 is only "D". SOIC16 is "DW" and only from one vendor.

Due to the overlap I think the package name suffix makes more sense. I've split SOIC8 off and also added the SOIC16 version. Oddly, the datasheet shows two VC pins for SOIC16 instead of VC and VCC; I kept the two separate pins as in SOIC14.
image

Anything else?

jkriege2:
Looks good to me!
Thanks,
JAN

PR #1484, opened 2017-08-03
Add LTC1436/1437 buck regulator

evanshultz:
Datasheet: http://cds.linear.com/docs/en/datasheet/14367afb.pdf

Image:
image


jkriege2:
nice work!

JAN

PR #1486, opened 2017-08-04
removed old wings picoblade connector models

Shackmeister:
have been replaced with the CQ based models from @poeschlr


poeschlr:
I have not created this series. (It is the THT version)
I only created 3d models for smd versions: (`##` stands for the number of pins)
`Molex_PicoBlade_53261-##71_##x1.25mm_Angled`
`Molex_PicoBlade_53398-##71_##x1.25mm_Straight`

But the current 3d models are named wrong.
should be
`Molex_PicoBlade_53047-##10_##x1.25mm_Straight`
is
`Molex_PicoBlade_53047-##10`
The `53048` series is missing completely (angled tht picoplade version)

I hope i will find some time soon to create these 3d models though. (now that i know that this is simply the tht version of my existing smd scripts.)
I might also look into creating the smd version of the ph series.

Shackmeister:
@poeschlr didnt actually notice these were the THT headers, sorry for my fuzz then
should I rename these and change the offset in the footprint to make them usable/used or should they simply be deleted? the colors are also kinda funky
picoblade_tht

poeschlr:
I think we can leave them in there for now. But we should keep them in mind for replacement. (I hope i will find time in the next few weeks.)

Shackmeister:
@poeschlr im just gonna close this now. Since I have no experience with wings I have no Idea how to offset the model,  I could change it in the footprint, but that would need to be reverted when new models would come. Looking forward for your models:)

Issue #1451, opened 2017-07-26
What should we do with dual (or multi) row connectors that do not follow the odd/even pin numbering


poeschlr:
#There are dual row connectors that do not follow the even odd pin ordering used in the current library symbols.

Examples:
There are connectors that have the classic dip style pin ordering (counter clockwise numbering).
https://github.com/KiCad/Connectors_Harwin.pretty/pull/8
Here the pins stay normal numbers but the order is different. I would use the suffix _DIP-numbering to signal this.
(Not sure if this would be clear enough. The description will of course have the info that this symbol has pin ordering similar to the dip packages. The current dual row connectors would get odd/even mentioned in the description.)
**Edit:** I just discovered it does not follow the dip convention but has its own convention where both rows start at the same side and count up in the same direction. (similar to subd) 

There are phoenix multi row connectors where the mating connector is not one part for all rows. (It is one connector per row) I'm sure something similar is produced by other manufacturers as well.
https://github.com/KiCad/Connectors_Phoenix.pretty/pull/46
Maybe a good pin ordering here would be similar to BGAs using a letter for the row and numbers for the pin number within this row. (This also works for 3 an 4 row connectors. Phoenix manufactures such connectors.)
It might even be a good idea to create these connectors as multi unit symbols.
I really don't have a good naming convention for such symbols. Maybe suffix _BGA-numbering

bobc:
>both rows start at the same side and count up in the same direction

We already have connectors that follow this convention, e.g lots of Molex connectors [Mini-Fit Jr](http://www.molex.com/pdm_docs/sd/039288100_sd.pdf)

Not sure there is a problem, surely we should follow the manufacturer? In this case of Molex, they mark pins 1, 2, 3 with stripes and a digit, so it would be *very* confusing to adopt a different convention.

![f6795773-02](https://user-images.githubusercontent.com/972103/28743942-3c114a6c-744e-11e7-9d2e-339addc1bbae.jpg)

Some connectors use a letter for rows, e.g. 1a, 1b, or A1 B1.

I'm not even sure that the even/odd ordering is defined anywhere, although it seems to be a common convention. For dual row 0.1 headers, a nominal pin 1 is marked for orientation purposes, but the pin numbering is up to the user I think.

bobc:
ERNI connectors, where row 1 is a1..an, row 2 is b1..bn. Also marked on connector as such \:

[...]

Issue #1170, opened 2017-04-18
Footprint filters for conn_??x?? symbols are very restricted should they include more footprints?


poeschlr:
We have a lot of footprints for different connectors. But the conn_??x?? symbols have only pin headers and socket strips in their footprint filters.
Any suggestions how the footprint filters should look like? 
Should we add all applicable footprints manually or is there a better option?

SchrodingersGat:
This behaviour is in nightly only but it would pay to make use of this for 5.0.x

poeschlr:
How does the stable version react if there is a ":" character in one of the footprint filters?

SchrodingersGat:
It tries to match it just against the footprint and fails

SchrodingersGat:
It was exactly for the use case that you describe here that I implemented this filter feature :)

poeschlr:
Thanks  for implementing this feature and for the clarification.
If nobody else volunteers i would add this filters to the conn_ symbols. But i probably will not be able to do it before the end of may.

SchrodingersGat:
Should be easy enough to fix with a handy text editor and some regex :)

poeschlr:
@SchrodingersGat What nightly version did add this feature? I would like to test this but i think my nightly might be a bit too old. (I might need to update my operating system to get updates again.)

evanshultz:
@poeschlr Can this be closed now?

PR #1487, opened 2017-08-05
Cleanup of conn.lib

jkriege2:
- moved PAL-symbols to new pld_pal.lib (these will need overhaul in a separate PR)
- renamed PCI&ISA-bus "connectors" in a common style, updated descriptions
- unified connector-naming style CONN->Conn, BUS->Bus, ...
- moved remaining IC (PCI9030-PQFP176) to interface.lib + added FPFilters (symbol also needs overhaul in a separated PR)

This is done as a preparation (in parts) for the proposed changes by @poeschlr there: https://github.com/KiCad/kicad-library/issues/1451

PS: This is not about conforming to KLC, and I will add additional PRs for cleaning up the KLC-issues!


evanshultz:
@jkriege2 
Would you want to add FPfilters to all `Conn_xxXxx` parts here, or wait until later? Currently, it seems there's many missing, or the footprints don't exist, or some other issue. I would think at a minimum, each of these parts could at least have `Conn*`, `Pin*`, and `Socket*` for the FPfilter to at least get rid of most other footprints which aren't close to a connector.

Perhaps it's best to do this later once we have things more settled?

jkriege2:
I'm not sure about FPFilters: There are so many matching connectors and no naming convention on the side of connectors. therefore the best would be (IMHO) to either filter by Library, or simply omitting the FPFilter.

I would like to merge this small PR now and then decide later, if we want FPFilters, or other changes ...
What do you think?

Comments on the other changes?

JAN

evanshultz:
Many of them have an FPfilter now. At least in the fullness of time, I think it's helpful to have some structure here so that connectors are filtered by default.

But reviewing the other changes.
-I really like the camelcase. Much more readable. And the renaming of some of the computer bus connectors is much more salient.
-Is there a typo here?
image
-Should these be camelcase too?
image
image
-Regarding the libraries added and parts moved, I'll let Oliver weigh in on that. Have the new libraries been all settled?

poeschlr:
My script will be capable to generate symbols for single and dual row connectors. Including Conn_Male and Conn_Female if we decide we want to keep them.
(Why do we want male and female variants for single row connectors but not for dual row connectors.)

I will add footprint filters of the style Connector\*:\*_[number of rows]x\* for multi row connectors and Connector\*:\* for single row connectors. (currently most if not all single row connectors do not include the number of rows in their name.)
According to https://github.com/KiCad/kicad-library/issues/1170 this should work for nightly builds and be ignored in stable. (the underline is necessary to filer the number of rows instead of the number of pins.)

In other words: Within the next few days the symbols will be replaced by scrip generated variants. So don't use too much time in this pull request for these symbols.

jkriege2:
OK,

then I will do the smaller Camelcase-Renames and we can go from there. @poeschlr please use the Camelcase-naming?

Best,
JAN

jkriege2:
OK, I fixed the typos/missing renames above.
The part-moving (that strange PCI-controller) should be fine by our current standards, so we can move it on, when we do the restructuring. The PAL-devices ... mmhhh ... I think we didn't decide on a general programmable logic lib, so I took a name that was close-enough and starts with `logic_` ;-) Also these symbols are of VERY low quality. If I find some time, I might draw up proper PAL/GAL symbols ... at least a few, so we have something. I think these are not really used anymore today?

Best,
JAN

poeschlr:
I will use camel case naming for the connectors I create.
What about male and female variants? Keep or delete. (If we keep them should there also be such symbols for multi row connectors?)

poeschlr:
Can't edit my comment on mobile. 
I vote for deletion but would include a note in the readme that there is a script capable to generate such symbols.

jkriege2:
I would keep them ... I find them a neat extension, especially since you can pair the variants in the lib, which can be handy sometimes, BUT I think the limited number of single-row MALE/FEMALEs we have now is sufficient. I think that was also decided, when we discussed that particular PR: https://github.com/KiCad/kicad-library/pull/816
... in the end we also have special connector symbols for other cases (screw terminals, wall-plugs/sockest/MINI-DIN,SUB-D,coaxial...) basically everything, where we want the possibility to say explicitly what we mean in the schematic, but we limited also these to reasonable amounts (I think).

JAN

poeschlr:
@evanshultz anything else that needs to be done here?
I will open my pull request as soon as this has been merged.

evanshultz:
If the libraries are OK, it's fine with me. I only took a quick look and saw the typos. Thanks for checking!

jkriege2:
So can this be merged? @evanshultz @poeschlr

poeschlr:
I had another closer look. I can't find any problems with this PR. Thanks for your work.

PR #1488, opened 2017-08-06
Ads7843 2

Misca1234:
This is push for ADS7843

ADS7843E with aliases ADS7843E/2K5, ADS7843EG4 and ADS7843IDBQRQ1

ADS7843E, ADS7843E/2K5, ADS7843EG4
http://www.ti.com/lit/ds/symlink/ads7843.pdf

ADS7843IDBQRQ1
http://www.ti.com/lit/ds/symlink/ads7843-q1.pdf

bild


jkriege2:
What about stacking the +VCC pins?

jkriege2:
Also, FPFilter should be `SSOP*3.9x4.9mm*Pitch0.635mm*`

Best,
JAN

Misca1234:
Uhm, what do you mean with "stacking", adding to pin number to same pin ?

jkriege2:
stacking: see KLC 4.3 (https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention)

Misca1234:
Like this ?

bild

jkriege2:
no, you simply have one pin +VCC,POWER-INPUT,visible and one pin +VCC,PASSIVE,invisible and move them to the same spot. You'll only see the visible pin in the schematic, but if you connect to it, both pins will be connected to the same signal. See regul.lib for several examples.

Misca1234:
Ok, I see, let me fix it.

But I think this "stacking pins" is a terrible concept. 
At least I would be confused of the symbol and would wonder where the "other VCC" pin is 
and that the designer of the symbol have forgot one pin.

Misca1234:
bild

jkriege2:
yes, the concept is not perfect, but the best we can do with KiCAD's current feature set.

Best,
JAN

PS: Nice work on this symbol :-)

Misca1234:
Thx, live well

PR #1490, opened 2017-08-06
Ak4182 a

Misca1234:
This is push for AK4182A

AK4182A
http://www.akm.com/akm/en/file/datasheet/AK4182AVN.pdf

bild


jkriege2:
FPFIlter should be `QFN*4x4mm*Pitch0.65mm*`
Best,
AJN

jkriege2:
Also it appears that you have doubled a pin10!

Misca1234:
Mmm, that how it goes if checking a copy paste, I forgot to add pin 17 as well

jkriege2:
Nice Work!
JAN

Misca1234:
I suspect that the travis script errors should be ignored, The VBAT is actually an input and not power input and the VREF is redirectional according to data sheet

Misca1234:
Oh, you was fast, thanks

PR #1498, opened 2017-08-08
Add OP179/279

evanshultz:
Datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/OP179_279.pdf

All are ALIAS of another symbol:
- OP179GRT: MCP6L91R (http://ww1.microchip.com/downloads/en/DeviceDoc/22141b.pdf)
- OPA179GS: MCP601 in SOIC (http://ww1.microchip.com/downloads/en/DeviceDoc/21314g.pdf)
- OP279: LM2904 (http://www.ti.com/lit/ds/symlink/lm258a.pdf)


jkriege2:
Hi @evanshultz !

Basically all looks good. But I see a small problem: You made OP179 an alias of MCP601-xP, but OP179 is only available in SOIC, whereas MCP601-xP also lists DIP and TSSOP. Therefore the symbol for OP179 should be singled out, otherwise the FPFilters will match footprints that are not available for this symbol. Also then you can directly set the SOIC in the footprint field. The same problem is even worse for OP279 and LM2904, as the latter lists even more footprints (e.g. also some TO-99, DIFFERENT TSSOPs, ...).

Best,
JAN

jkriege2:
@evanshultz : Looking at linea.lib with my comments above in mind: Maybe another overhaul is required that fixes all these symbol-footprint-relations?

JAN

evanshultz:
This was no mistake or error. I grouped all symbols logically, to keep the smallest number of symbols, and then filled out all applicable FPfilters for the group.

jkriege2:
I suspectedsomething like that ... I think that's not the best way to do this, as we have symbols that (per name) only are available in say 2 packages, but the FPFilter outputs 10 possibilities, of which 8 are plain wrong ... In the end: What would be the pont to have such an ALIAS for specific packages, if the information on the packages is not available?

I think we should really split the symbols (even if this doubles the records in the .lib), to keep the correct symbol-footprint association! (+ I would interpret out FAQ to describe exactly what I'm proposing)

I think the software can easily handle more symbols (they are there anyways as ALIAS, but nevertheless there) ... and file-size is not so severe I think (we're talking Kilobytes, not Megabytes here + it's text, which can easily be compressed!)



@SchrodingersGat @poeschlr @pointhi @CarlPoirier Your thoughts on this?
JAN

evanshultz:
I can only see this happening in linear.lib and transistor.lib right now. Elsewhere, at least from what I can see and recall, symbols would be too unique to do anything like this.

poeschlr:
I agree with @jkriege2 here. If two components don't share all footprints they should be separate symbols.

evanshultz:
OK. Then I can change the library to suit. I'm working on a few others things right now, but I'll plan to get onto this shortly if the consensus is to go this way.

If there aren't any problems with these symbols can they be merged so this opamp family is included?

jkriege2:
Then could you single out the symbols as appropriate here, before we merge? So at least the symbols from this PR are correct.

JAN

evanshultz:
@jkriege2 I make OP179/279 discrete parts, assuming that we will go that direction. I'll do it once I finish off a couple other library fixes. Would you have time to review?

jkriege2:
Great work!
JAN

PR #1499, opened 2017-08-08
PAM2305AAB3XXX added

aliteke79:
pam2305aabxxx


jkriege2:
Hi!

please fix the branch conflicts, so we can continue with the review!

JAN

jkriege2:
I also had a look at the actual symbols (in the current version):
- PAM2305AAB3XXX: Please don't `x`out the voltage ratings, but add each voltage variant (use the first as basis and add the others as ALIASes to that). See other comparable devices in the dc-dc.lib Otherwise: Very nice work!
- LTC4054X: please add the LTC4054-4.2 from teh DS as an ALIAS
- LT3465 looks good

i.e. I can merge, when the ALIAS issues are resolved!

Best and thanks for your contribution,
JAN

aliteke79:
Hi @jkriege2. Some one is also committed the LTC4054X component, so i removed it and done other changes.

jkriege2:
Just had a short look ... please change the FPFilters for TSOT-23-5/6 to `TSOT?23*`. Together with the pin-count filter that will match the correct footprints (i.e. pin-count is not required in FPFilters).

When that's done I'll merge!

Best,
JAN

aliteke79:
Hi @jkriege2. I don't see TSOT-23-5/6, FPFilters looks like TSOT*23*5* and TSOT*23*6*

jkriege2:
You used `TSOT*23*6*` but it should be without the pin-count, i.e. `TSOT?23*` ... does that help?

aliteke79:
Yes, i just updated it.

jkriege2:
great work!
Thanks,
JAN

PR #1505, opened 2017-08-08
Add transistors: BSP129, IRF7403, STS2DNE60

evanshultz:
BSP129: https://www.infineon.com/dgdl/Infineon-BSP129-DS-v01_42-en.pdf?fileId=db3a30433c1a8752013c1fc296d2395f

IRF7403 (ALIAS of IRF8721PBF-1): https://www.infineon.com/dgdl/irf7403pbf.pdf?fileId=5546d462533600a4015355fa23541b9c

STS2DNE60 (ALIAS of FDS6890A): www.st.com/resource/en/datasheet/CD00001537.pdf

Image of BSP129 as the others are an ALIAS:
image


evanshultz:
Note that BSP129 has a stacked pin 4 under pin 2, in accordance with the datasheet and fitting the SOT-223 footprint in the library. I saw other cases where the SOT-223 parts only used 3 pins, which wouldn't connect the tab. Is that an error, possibly from starting with a copy of a 3-pin transistors, and all 4 pins should be connected in our symbols?

evanshultz:
The KLC errors, as you'll see are due to these primitive symbols using 50mil pin grid. Since my symbols are an ALIAS, my PR isn't responsible and it should be OK to merge as is.

evanshultz:
Can this be merged?

jkriege2:
Hi!

symbols look good ... one suggestion: You could simplify the symbol for BSP129 by using the footprint `TO_SOT_Packages_SMD:SOT-223-3Lead_TabPin2` in which the Tab is also designated pin2.

Otherwise I can merge!

JAN

evanshultz:
Done. Thanks Jan!

jkriege2:
nice work!

JAN

PR #783, opened 2016-11-05
Added TPS3839

greenscreenflicker:
Added texas voltage supervisor TPS3839 tp texas.lib

PR #1511, opened 2017-08-11
Update TL770x family, add TPS3831/9

evanshultz:
TPS3839 (from https://github.com/KiCad/kicad-library/pull/783): http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf
- Fixed all issues noted in the PR above
- Added TPS3831
- Added X2SON footprint at https://github.com/KiCad/Housings_SON.pretty/pull/12

TL7702 (Changes from original symbol): http://www.ti.com.cn/cn/lit/ds/symlink/tl7705a.pdf, http://www.ti.com.cn/cn/lit/ds/symlink/tl7705b.pdf
- Renamed pin 1 to REF (it was 2.53V which is only true for TL7702 parts)
- Capitalized all pin names
- Indicated pins 2 and 5 are active low
- Added FPfilter
- Changed symbol names to A and B families (I don't find any without a suffix)
- Made all pins 100 mil in length
- Rearranged pins to match the datasheet and shrunk symbol
- Add DCM entries for all parts
- I can go skinny and tall instead of shorter and fatter if that's the style you all prefer.

Images:
image


evanshultz:
Forgot I already submitted this at https://github.com/KiCad/kicad-library/pull/1497. Closing that case.

Can this be merged?

jkriege2:
Hi!

some minor things:
- TPS3831: The thermal pad should be a separated visible pin "PAD", as the DS says this CAN be connected to GND, but can also be connected to a floating plane: `Connect to ground or to a floating copper plane for mechanical stability.`
- TPS3839: same as above
- TL7705A: This one is available in a SO-8 package (6.20 mm × 5.30 mm), not SOIC (4.90 mm × 3.91 mm), so please single it out, the footprint is not available yet I think (would you have time to draw it?)
- other symbols seem OK to me ... 

Best,
JAN

evanshultz:
Separated the X2SON footprint thermal pad pin.

Sorry about the TL7705ACPSR; I didn't realize the SO was different size. I separated it from the other TL77xx parts and will add that footprint.

Ready for merge!

jkriege2:
Then we only need to decide on the naming of the TL7705ACPSR footprint: SO-8 or SOIC-8?

evanshultz:
Looking at the datasheet and reading into your comment above, I'm thinking you like `SO-8_5.3x6.2mm_Pitch1.27mm`. Done.

jkriege2:
Nice work!

let me know when you do the footprint ...

Best,
JAN

PR #1516, opened 2017-08-12
Changed the RC1602A symbol

Misca1234:
Fixing symbol RC1602A-GHW-ESX

http://www.raystar-optronics.com/down.php?ProID=18

Foot print and 3D model is already merged

bild


poeschlr:
Nice symbol but i would keep K (LED-) as passive. (or has the display an internal led driver. If it has then both A and K should be power input.)

In the footprint you have pads with 'number' A and K but there is no pin in the symbol that connects to these.

Misca1234:
Something like this 

bild

Misca1234:
It seems that I can not name the pins to A and K, I will change the foot print so they are named A1 and K1

Misca1234:
bild

Misca1234:
See 
https://github.com/KiCad/Displays.pretty/pull/34

for new foot print

poeschlr:
You used K1,K2 and K3 in the footprint. Either add K2 and K3 to the symbol as well or name all pads the same.
(Similarly with the Ax pins)

Misca1234:
I updated the footprint
see 
https://github.com/KiCad/Displays.pretty/pull/34

poeschlr:
Thanks

Misca1234:
Thx

PR #1517, opened 2017-08-12
Icl7673

Misca1234:
Intersil ICL7673

http://www.intersil.com/content/dam/Intersil/documents/icl7/icl7673.pdf

bild


Misca1234:
I have no idea why HIP2100IR4 was affected, I recreate push

PR #1518, opened 2017-08-12
Icl 7673 2

Misca1234:
Intersil ICL7673

http://www.intersil.com/content/dam/Intersil/documents/icl7/icl7673.pdf

bild


Misca1234:
Strange, I made the push twice and in both cases kicad have on it's own modified the HIP2100 symbol, 
Is this some auto thing in kicad ?

Gasman2014:
Shouldn't it be GND not GDN? Think it might be wrong on the datasheet.

Misca1234:
I guess that too, but the kicad lads are usually very picky about the names, but I change it it GND

Misca1234:
It looks like this now 

bild

poeschlr:
> Strange, I made the push twice and in both cases kicad have on it's own modified the HIP2100 symbol,
Is this some auto thing in kicad ?

It looks like your version did change the order in which the entries in the lib and dcm file are made.
(Not a problem for me.)

---

Now the symbol review:
The symbol itself looks great. You guys are correct the pin should be named GND (Look at the drawings later in the datasheet. There they use GND.)

Your current footprint filters might not be the best choice. (I get switches in the filter result.) A better choice might be the filters i listed below.

I would create two symbols: (naming from order information. I did not find an explanation for all letters of the MPN. I would simply ignore retired parts.)
- ICL7673CPAZ
  - default footprint: `Housings_DIP:DIP-8_W7.62mm`
  - footprint filter: `?DIP*W7.62mm*`
- ICL7673CBAZA
  - default footprint: `Housings_SOIC:SOIC-8_3.9x4.9mm_Pitch1.27mm`
  - footprint filter: `SOIC*3.9x4.9mm*Pitch1.27mm*`

Misca1234:
Fixed

bild

bild

poeschlr:
Looks good now. Thanks

Misca1234:
Thx

PR #1520, opened 2017-08-12
Ds2756

Misca1234:
Maxim DS2756E

https://datasheets.maximintegrated.com/en/ds/DS2756.pdf

Travis complain about VIN, it is not a power pin but it is an input pin

bild


Misca1234:
Travis complain about VIN, it is not a power pin but a input pin

jkriege2:
nice work!

Thanks,
JAN

Misca1234:
Thanks

PR #1522, opened 2017-08-12
Max1898 eub4 x

Misca1234:
Push for 
MAX1898EUB42

Aliases
MAX1898EUB41

https://datasheets.maximintegrated.com/en/ds/MAX1898.pdf

bild


Misca1234:
PS
I have a number of pushes where travis complain of some pins that it thinks is power input, 
like push https://github.com/KiCad/kicad-library/pull/1520, 
is there a way to tell travis that the pin is actually an input and not a power input ?
DS

poeschlr:
> I have a number of pushes where travis complain of some pins that it thinks is power input,
like push #1520,
is there a way to tell travis that the pin is actually an input and not a power input ?

Just state that information in the pull request description. (If there is a reason a rule is violated we will accept it. Especially if you state what rule is violated and why.)

SchrodingersGat:
Please add a footprint filter to this symbol and then we should be good to merge :)

Misca1234:
Done

bild

Misca1234:
Kicad have changed something in MAX2606, let me fix that one as well

Misca1234:
Replaced MAX2606 with MAX2605
Added aliases 
MAX2606
MAX2607
MAX2608
MAX2609

bild

bild

bild

bild

bild

PR #1524, opened 2017-08-12
Added AD586

chrigro:
screenshot from 2017-08-12 23-11-17


jkriege2:
Hi!

thanks for your contribution. I checked the symbol against our styles and the datasheet and found a few issues:
1. pin length of 100mil is way enough
2. please add all the TP-pins. You can anme them NC, set gtheir type to NOT_CONNECTED and make them invisible, but they need to be there so the pin-count-filter works together with the FPFIlters!
3. You're missing the NOISE REDUCTION pin 8 from the datasheet. Please add it.

Best,
JAN

jkriege2:
PS: You can also move the NC-pins as shown below:
2017-08-18 19_53_29-part library editor -- d__kicad_kicad-library-review_library_interface lib

This way they are out of the way and cannot accidentally be connected.

Best,
JAN

Misca1234:
@chrigro 
Some liberians prefer that you mention the capsules in the description line, that is, change 
High Precision, 5 V Reference
into
"High Precision, 5 V Reference, DIP-8, SOIC-8"

chrigro:
Ok, I added the missing pins and fixed the pin length. 

Not sure what to do about Misca1234's comment. Ssome liberians" does not sound as if there is a general agreement on that?

Christian

jkriege2:
Hi!

adding the package to the description is a consensus requirement by now ... so please add it. Otherwise this symbol looks good and I will merge, when changes are done.

Thanks for your work!

JAN

jkriege2:
Thanks,
JAN

chrigro:
... Sorry the AD8230 should have gone to a new PR. I have to figure out how that works on github.

jkriege2:
Hi!

the AD8230 endet up split over two libs: analog_devices and analog_switches ... Please add this symbol in a second PR and clean the commits for it from this PR! I will merge this one with AD586 only and then review your new PR! Doing it that way is much simpler than adding new symbols late in an already reviewed PR!

Best,
JAN

jkriege2:
ahh ... just saw your commetn (only after writing myself) ... could revert the commit?

I usually generate a branch (from master) for each new PR. You can transfer the commits e.g. with a patch by hand ...

best,
JAN

chrigro:
Done. Thanks for the explanation. I will follow your advice and use specific branches per PR.

best,
Christian

PR #1525, opened 2017-08-13
Max1873

Misca1234:
MAX1873

https://datasheets.maximintegrated.com/en/ds/MAX1873.pdf

bild


jkriege2:
Hi!

nice work!
One small request: Please add the package-name as `, QSOP-16` to the description. Then I can merge.

Best,
JAN

PS: I noted that Maxim has different variants of that package with differen plastic heights ... do you happen to have one of these chips and can confirm that this is the smallest? I suspect it is, as the area 30mm², which Maxim indicates on their homepage) approximately equals the over-pin area estimated for the smalles package ...

Misca1234:
I am sorry, I do not have the chip, do you want me to add a footprint and 3d model of maxims specific package ?

jkriege2:
typo: You added QSOP-12 instead of 16 ... ;-)

The PS: Na, it's just fine ... just wanted to double-check how I (and obviously you) interpreted the DS ...

Please fix the typo I mentioned above and I can merge!

Best,
JAN

jkriege2:
Thanks,
JAN

Misca1234:
Thx

PR #1526, opened 2017-08-13
Add different product from codemercs

lorem-ipsum:
This commit series adds files for product from [Code Mercenaries](http://codemercs.com). They produce specialised ICs for the creation of input devices such as keyboards and joysticks. Currently only some chips are added which I need right now, but I am planning on adding their whole product range (it is rather small compared to some other manufacturers).


poeschlr:
If these are interface devices why not put them into the interface lib?

To be clear: We are moving away from a library that is organized by manufactures towards one where we organize by function. (Most of this reorganization will happen with the v5 release. So currently existing libs will exist for a few more month. But we want to stay away from creating new manufacturer libs if there is a functional lib that is suited for the components.)

Might it be possible to add screenshots of your symbols and a link to the datasheet?
This makes reviewing pull requests a lot easier.

jkriege2:
Hi!

I just took a look ... generally teh symbols look good. Please changes tehse thigns:
1. add package to the description, e.g. `, DIP-24/SOIC-24`
2. JoyWarrior24A10L: X0/B0 (pin1) is BIDI shouldn't that be INPUT? and REG POWER_OUTPUT?
3. SpinWarrior24*: POWER should be input, not POWER_INPUT
4. SpinWarrior24*: EN should be OpenCollector
5. PullToGDN should be INPUT not POWER_INPUT (as per DS)#
6. SpinWarrior24*: move VALUE-label to the same location as for JoyWarrior.
7. As we are currently reorganizing the lib-structure, I would say these symbols should go (for now) to `interface.lib`. So please move them there. If you're interested, you can take part in the lib-restructuring discussion there: https://github.com/KiCad/kicad-library/issues/1402

Otherwise these symbols look fine to me. 

Best & thanks for the contribution,
JAN

lorem-ipsum:
Ok, I will add your suggestions, go through the pin types and move everything to interface.lib. Didn't knew about the new structure, but reading through the thread it appears to make a lot of sense. I will also add screenshots for the next version.

lorem-ipsum:
I've updated the schematics and moved everything into the interface library.

### JoyWarrior24A10L
[datasheet](http://codemercs.com/downloads/joywarrior/JW_Datasheet.pdf)
joywarrior24a10l

### SpinWarriors
[datasheet](http://codemercs.com/downloads/spinwarrior/SW_Datasheet.pdf)
spinwarrior24a3
spinwarrior24r4
spinwarrior24r6

jkriege2:
nice work Thanks,
JAN

PR #1493, opened 2017-08-07
Analog devices library overhaul and new parts added

evanshultz:
Image:
image

Notes:
1. I have not moved power pins to the top and bottom of the isolator symbols. Since they have distinct power domains, I want to give everybody a chance to give feedback. Should they be moved to the top and bottom or not?
2. It would be helpful to get an answer for my question at https://github.com/KiCad/Housings_QFP.pretty/issues/18 since it affects this PR.

All:
- Added package type to description
- Standardized description and keywords
- Made pins 100mil in length

AD637: http://www.analog.com/media/en/technical-documentation/data-sheets/AD637.pdf
- New part
- VIN is not a power input so KLC error can be ignored
- Used "NC" instead of the datasheet's "NIC" so KiCad automatically recognizes this as a no connect pin

AD654: http://www.analog.com/media/en/technical-documentation/data-sheets/AD654.pdf
- New part

AD736: http://www.analog.com/media/en/technical-documentation/data-sheets/AD736.pdf
- New part
- VIN is not a power input so KLC error can be ignored

AD9831: http://www.analog.com/media/en/technical-documentation/data-sheets/AD9831.pdf
- New part

AD9852: http://www.analog.com/media/en/technical-documentation/data-sheets/AD9852.pdf
- New part
- Added EP as a discrete pad since it's not tied internally but is suggested to use a ground pad; see page 38 of datasheet
- VOUT is not a power output so KLC error can be ignored

AD9854: http://www.analog.com/media/en/technical-documentation/data-sheets/AD9854.pdf
- New part, ALIAS of above

AD9910: http://www.analog.com/media/en/technical-documentation/data-sheets/AD9910.pdf
- Stacked power and ground pins; shrunk symbol body
- Buried NC pins inside symbol body

ADE7758: http://www.analog.com/media/en/technical-documentation/data-sheets/ADE7758.pdf
- Moved VN pin to the right side since it's the VA/VB/VC reference and not ground

ADM2484E: http://www.analog.com/media/en/technical-documentation/data-sheets/ADM2484E.pdf
- Dramatically simplified the isolation barrier drawing to look like datasheet
- Stacked ground pins

AD2582E/2587E: http://www.analog.com/media/en/technical-documentation/data-sheets/ADM2582E_2587E.pdf
- Changed two isolation lines to one line datasheet
- Stacked power and ground pins; shrunk symbol body

ADuM6000: http://www.analog.com/media/en/technical-documentation/data-sheets/ADuM6000.PDF
[...]

PR #1529, opened 2017-08-13
add ADuM5000

lorem-ipsum:
it's basically the same chip as the ADuM6000, just with a lower protection rating. The only change is to add the missing alias.
I've used the description style used by #1493, but did not change anything else at the chip to minimize conflicts with that PR.


jkriege2:
Thanks,
JAN

Issue #1447, opened 2017-07-26
RFC Footprint naming convention (originally naming convention for EP size)


poeschlr:
It seems the exposed pad is not always the same size for the same package.

Something similar was discussed in an issue over at the dfn qfn repo: https://github.com/KiCad/Housings_DFN_QFN.pretty/issues/24
In the issue we discovered that different manufacturers use different pin and exposed pad sizes.
(I'm not sure we reached a conclusion over there of what the naming convention should be.)

Now there is also a pull request for qfp packages with different exposed pad sizes but same pin sizes. (All by the same manufacturer)
https://github.com/KiCad/Housings_QFP.pretty/pull/17

We should find a naming convention that lets us communicate these differences.

For the qfp pull request i would suggest to include the x and y dimension of the exposed pad.
Example: LQFP-144-1EP4x4mm_20x20mm_Pitch0.5mm

I think this should be fixed and added to the KLC. (This is why i opened the issue here.)

SchrodingersGat:
@poeschlr definitely a good idea, although I think the package dimensions should have a higher priority in the name. 

[Here](https://www.cskl.de/fileadmin/downloads/PCBLIBRARIES/Documentation/IPC-7351C-Land-Pattern-Naming-Convention_.pdf) is a document that I have been referencing a lot for direction on naming.

Shackmeister:
+1 for both 
Maybe we should also think about a naming scheme to seperate two similar footprints with different pad shapes (oval and ret)

poeschlr:
As far as i understand it oval pads are only used because rounded rect is not yet supported. (IPC specified oval because rounded rect was not supported by all major cad tools back then.)
The current recommendation is to use rounded rect (not sure if it is already in IPC.) It seems that rounded rectangles behave a bit better when reflow soldering. [Source](https://www.pcblibraries.com/forum/rounded-rectangle-pad-shape_topic1462.html) (I seem to remember that i found a better source when i looked for it last time. Can't find it today.)

How should the rounded rects look like?
- Corner Radius = 25% of Pad Width (pad width is the narrowest part of the SMD pad)
- Maximum Radius = 0.25 mm, so regardless of how wide the pad is the corner radius will never exceed 0.25 mm (10 mils)

So it might be a good idea to choose rounded rect for all our footprints as soon as it is available in a stable kicad release. (I don't think it is a good idea to mix oval/rectangular/rounded rect.)

poeschlr:
[...]

PR #1530, opened 2017-08-13
Added support for PIC18F2xK50 family

carlo-bramini:


carlo-bramini:
Hello, excuse me, this is my first contribution to kicad-library. I think I corrected all pin placement to 100 mil multiple and I added the missing descriptions, but I do not know how to fix this error:

    Symbol name 'PIC18F24K50-I/ML' contains invalid characters as per KLC 1.7

If I understood correctly, it complains that a '/' character is inside the symbol name, but actually this is the name of the component, like the other MCUs inside microchip_pic18mcu.lib or .dcm
How can be this defect fixed?

carlo-bramini:
immagine

carlo-bramini:
I followed the instructions from this commit:

https://github.com/KiCad/kicad-library/pull/1436/commits/76168473213886a0b1e98d547d69f7c6106206cc

"1. Prohibited characters - please do not use ( ) / \ characters in part names (as per the KLC).
Note: Yes, this goes against a lot of the Microchip MCU parts, but we are working to improve them
So, I would rename as thus:
PIC16F18325-ISL"

So, I did the same thing. Thank you.

poeschlr:
I think the naming convention is in preparation for the new file format where there will be one file per symbol. (A '/' in a filename is a bad idea.)

The symbol would look even better if the pin names are shorter.
Choose a name that identifies the pin and use that. (Example: Rxy; x = A, B, ..; y=0, 1, 2, ...)

Yes the current symbols in this lib show the alternative functions in the pin name. 
In other libs we already removed them. (Results in a symbol that is a bit easier to read.) 

We hope the new lib format will allow for alternative pin names. Until we have this new format we can not really do better than to use shorter names.

For inspiration have a look at the nxp_armcpu lib.

carlo-bramini:
I followed your suggestion and I stripped functions from pins and I made more compact components.
Thank you for your time.

immagine

poeschlr:
What is the difference between -EML and -IML? Either highlight this difference in the description or have only one symbol (replace E/I with a wildcard. Example instead of having PIC18F24K50-IML and PIC18F24K50-EML only use PIC18F24K50-xML)

The same is true for other packages (-I[package code] and -E[package code])

What is the difference between PIC18F24K50 and PIC18**L**F24K50? (Highlight it in the description.)

Package versions MV (uqfn), P (pdip) and PT (TQFP) are missing. (Did you forget them or are they not available. Or is it too much work to include them.)

I will review the symbols them selves after lunch. (pin types and pin numbering, footprint field and filter)

carlo-bramini:
Hello, according to the datasheet:
immagine

I did the fixes that you requested into the descriptions. Thank you.

carlo-bramini:
For the difference between 'I' and 'E':
immagine

poeschlr:
Normally we do not include different symbols for different temperature ranges.
So i would go with using a wild card instead and reduce the number of symbols by a factor of two.

Including the voltage range is ok. (As long as it is mentioned in the description.)

carlo-bramini:
I see, I removed the temperature range and I used a wildcard instead as you suggested. Thank you.

poeschlr:
Looks good. Maybe the qfn footprint is not quite right. (Exposed pad too large.)
If you are motivated you could add a separate qfn package with the correct exposed pad size.

Sadly the naming for such a package is not yet fixed.
https://github.com/KiCad/kicad-library/issues/1447

carlo-bramini:
Actually these mcu identical footprint to the PIC18F2450 series, so I just did a copy paste from these components since the already existed in library. Are they wrong too?

poeschlr:
I had another look. The footprint might be ok. In it the pad for the exposed pad is 4.25x4.25mm.
The package pad is nominal 3.7mm. But the maximum is 4.2. (Would mean 4.25 is ok for the maximum size.)

Sorry for the confusion. (I have not noticed that the thermal pad has such large tolerances when i first looked at it.)

Now only one thing to say: Thanks for your contribution.

PR #1531, opened 2017-08-13
Ltc4060

Misca1234:
LTC4060EDHC and LTC4060EFE

https://www.elfa.se/Web/Downloads/_t/ds/ltc4060_eng_tds.pdf

bild

bild


jkriege2:
Hi!

looks mostly good, the exposed paf should be tied to GND according to DS (it's internally connected to GND). therefore, please rename pin17 to GNDPAD, make it visible, a POWER_INPUT and move it to the bottom of the symbol (next to GND).

Otherwise: very nice work!

Best,
JAN

evanshultz:
@jkriege2 One question: If the thermal pad is internally connected, could it be stacked on the existing GND pin? I can understand if the thermal pad is floating but only recommended to be connected to ground that it should be a separate pin, but if the connection is already present internally isn't that a perfect condition for pin stacking?

jkriege2:
yes, I think that should be OK .... then call it simply `GND`.
JAN

Misca1234:
Now so

bild

bild

Misca1234:
Like this 

bild

bild

jkriege2:
Please fix branch conflicts so I can merge!

Thanks,
JAN

jkriege2:
Hi!

please have a second look ... your last merge deleted `$ENDCMP` from the DCM and `ENDDRAW` and `ENDDEF` from the lib!

Best,
JAN

jkriege2:
Great!

Thanks,
JAN

PR #1533, opened 2017-08-15
make all files non-executable

lorem-ipsum:
the only file permission git trackes is the execute bit. Some files in
this library are marked as executable, but none should be.


lorem-ipsum:
Mhm, github seems to die while comparing the changes. I've uploaded the output of a git diff as a [gist](https://gist.github.com/lorem-ipsum/0e87864065b36dc835a0dcd450650693). Most of the changed files are the 3d_packages

SchrodingersGat:
Thanks :)

PR #1534, opened 2017-08-15
Traco Power TMR 1 series

lorem-ipsum:
Isolating DC-DC converters with 1.5kV isolation. This patch adds the complete series. [datasheet](http://assets.tracopower.com/20170811130314/TMR1SM/documents/tmr1sm-datasheet.pdf)

<s>Currently the footprints are missing and the current naming I made up makes no sense. I am wondering where they should go and how to name them. The SIP packages use a SIP6 where pins 3 (and 5 on some) are missing. Similar for the SMD housing where some pins are missing.</s>
I have found the Converter_DCDC_ACDC library and apparently someone already produces the SIP packages for the TMR-1 a year ago. Now only the SMDs are missing, I will create a PR for them soon.

tmr-1


jkriege2:
Hi!

nice work ... I have some requests though:
1. please add the output-volatge to the description, so the description makes clear the difference between the different version/ALIASes
2. please remove the vertical bar --- we don't have that in any other symbol (or is there a rationale behind it? do these DC/DCs do galvanic isolation? In the latter case, we use dashed lines for that in some symbols.
3. The pin "Remote" should be named `Remote on/off` as in the datasheet.

Otherwise the symbols look good ... I'll also have a look at the footprint https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/19

Best,
JAN

lorem-ipsum:
I have pushed a new version. I did not change the descriptions as the earlier version already had the output voltages in their description. I've replaced the vertical bar with a dashed line, as they do have galvanic isolation. I also adjusted the Remote label, now the devices are sadly very asymmetric as seen below. But if I would have made them symmetric again they would be very wide.
tmr-1s-assymetric

jkriege2:
Hi!

1. instead of simply adding the volatge in front of the description, please orient the description on other symbols, especially in regul.lib, the TSR-series is not a good example ;-), i.e. add a text like `, 5V fixed output voltage` ...
2. The REFDES should be Ù`, not `T`! Please fix that

Otherwise the symbols look good!

Thanks,
JAN

lorem-ipsum:
Next round :-)
If changed the REFDEFs and adjusted the descriptions style to
`Regulated 1W DC/DC converter with 1.5kV isolation, 4.5V-9V input, 5V fixed output, SIP-6`

jkriege2:
Great work!

Thanks,
JAN

PR #1538, opened 2017-08-15
Add schematic symbols for ACPL-7900-300E

rasilon:
ACPL-9700-300E is an isolating op-amp


jkriege2:
Hi!

.... and thanks for your contribution. I had a quick look at it and found some issues. Please correct them, before we can merge this PR:
1. please don't change microchip.lib in your PR (there is a delete `#` for no obvious reason). the PR should only touch those files that it is concerned with.
2. You got the housings wrong: This device is (according to the datasheet) available in a standard 8-pin dip `Housings_DIP:DIP-8_W7.62mm` and a gull-winged SMD DIP (`Housings_DIP:DIP-8_W9.53mm_SMD` seems a close fit, although the DS proposes another landpattern). Therefore: Please empty the footprint field and add two FPFIlters `DIP*W7.62mm*` and `DIP*W9.53mm*SMD*` if the footprints I picked are correct.
3. If you have time, you could make the footprint proposed in the datasheet and we can link it here with an appropriate FPFIlter.
4. Please correct the package noted in the description field.

Best&Thanks,
JAN

jkriege2:
PS: Would you care to add the variants ACPL-790A/790B from the datasheet as ALIASes?

rasilon:
Hi @jkriege2 
My apologies for the noise; I was still getting my workflow organised.  I believe I've implemented all your review points, if you have the time to take another look.

Thanks,
Derry

jkriege2:
Hi!

I had a nother look (thanks for your tweaks so far): I found a small things that still remain to be fixed:
1. rename `ACPL-7900-300E` to `ACPL-7900`, the `-300E` specifies a specific package, but as you create a multi-package sybol that is not necessary and should really not be there (the user selects the package himself). 
2. remove the FPFIlter `DIP*W10.9mm*SMD*`, I did not find any indication for such a package in the DS (at least there are only drawings for the THT and the single SMD-version) and the package matched by `DIP*W9.53mm*SMD*` matches the drawing in teh DS.

When these are fixed, I think I can merge!

Best,
JAN

jkriege2:
aahhh ... just saw you package PR https://github.com/KiCad/Housings_DIP.pretty/pull/14 ... if you make a pakcgae PR, specific to a symbol, please link it here, so we see that it is there!

Thanks,
JAN

rasilon:
Hi Jan,
I believe I've implemented all your comments.  There are now three footprint filters, the one you suggested in [KiCad/Housings_DIP#14](https://github.com/KiCad/Housings_DIP.pretty/pull/14), the one from that PR (which looks a little different to me), and the one I used on the actual board and found to hand solder reasonably.
I've also added a 3D model with the suggested name, that appears a reasonable facsimile of the actual  part.

Thanks,
Derry

jkriege2:
Hi!

OK, mostly fine now:
1. please change `DIP*8*W7.62mm*SMD*` to `DIP*W7.62mm*` (we don't need the pin-number in the FPFilters, there is a pin-count filter in KiCAD + you were missing the non-SMD/THT-variant shown in the DS)
2. Where did you get the 3D model from? If you made it yourself, please add the sources to https://github.com/KiCad/kicad-packages3D-source and the WRL + a STEP-version to https://github.com/KiCad/kicad-packages3D. Also there should also be the STEP-version here. If you downloaded it from the Internet, please ensure that yiu are allowed to upload it here under the licensing of KiCAD (see https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions#submission-of-3d-models)

Thanks
JAN

library/opto.lib
```diff
@@ -133,8 +133,8 @@ F3 "" 45 5 50 H I L CNN
 ALIAS ACPL-790A ACPL-790B
 $FPLIST
  Avago*DIP*W10.9mm*SMD*
- DIP*8*W7.62mm*SMD*
- DIP*W9.53mm*SMD*
```
jkriege2: This can stay SMD, as there is no wide-pin THT-variant ...

rasilon:
Hi,
I've changed the wide footprint back. I created the model by  modifying DIP-8_W7.62mm_LongPads.wrl, so there is no other source for it and it didn't seem appropriate to add it to the other repos, since there is no Step, FC, or Wings master.  Is that OK, or should I pull it from the PR, or try to remodel it (which may take a while)?

Thanks,
Derry

poeschlr:
3d models for dip packages are script generated:
https://github.com/easyw/kicad-3d-models-in-freecad/tree/master/cadquery/FCAD_script_generator/DIP_packages

Simply input the measurements of your package in the parameters file and generate your model this way.

jkriege2:
@rasilon : Will you do a 3D package using the script @poeschlr pointed you to? 

In your footprint I've see that with your 3D model you need an offset to correctly position it. Nowadays we try to keep the offset at 0,0,0 and scaling at 1,1,1 in footprints and move the 3D models accordingly before exporting them to WRL/STEP.

As the scripts exist, I'm quite sure that someone will sonn provide a proper 3D model, so I would propose, if you don't want to make the model as suggested above, that we remove the model from this PR and set the offset in the footprint to the default values and I will merge then (if there are no more resons not to ;-)

Best,
JAN

jkriege2:
PS: If you make the model, just notify me and I will check back, when it's finished. OK?

SchrodingersGat:

PR #1539, opened 2017-08-15
Added LM25119 and LM25119-Q1

SG-O:
Also the WQFN-32 Package has been added.


poeschlr:
The symbol it self looks good.

A few minor points:
I don't think the alias for the automotive version is necessary. It clutters the lib and the user can just add `-Q1` to the value field if they need to communicate that they want to use the automotive version. (If we would include this detail, we would double the number of entries in the library.)

If VCCx is on the right side, maybe PGNDx should also be on the right side. (If i understand there function correctly.)

This symbol would be well suited for making it a multi unit symbol. (3 units, 1 per channel plus one for the support stuff) We can merge as is if you don't want to invest time to do it that way.
If you do it that way, VCCx and PGNDx should be on top/bottom for the sub units.

SG-O:
Thanks for the Ideas! I will change those aspects.

poeschlr:
I think the SS pins should be part of their respective units. I would definitely put it on the left side (towards bottom) of the unit.

I would also add EN2 to the unit for channel 2. (Again on the left side. But this time towards the top.)

Move the SW pin such that it ends up between HO and LO

Could you move the Reference and Value field outside the symbol?
---

There might also be the possibility to move the  Ramp, FB and Comp pins to the left side of their respective unit. Might make the symbol easier to read. (But would force the user to use labels to connect them even with the simplest external circuitry. Two labels might be required.)
I leave the decision about that in your hand. You know how this part might be used. (And how the circuit really looks like.)

SG-O:
EN2 is only pulled down or unconnected so it makes more sense to put it at the bottom of unit 2

poeschlr:
I fear something went wrong with your latest changes. Suddenly there are 3 pins named UVLO.
Did you disable the "Edit pins per unit" mode? (If this is at the wrong setting, you edit the pins on all units if they are on the same position as the one you are currently editing in the active unit.)
screenshot from 2017-08-19 12-07-34

SG-O:
I am absolutely sure I enabled "Edit pins per unit". The strange thing is that all the other pins are fine, only two pins were named UVLO and their type was changed to input. Still the pin  numbering was fine. I fixed the problem and double checked all pins.

poeschlr:
Thanks for this great contribution.

PR #1541, opened 2017-08-16
NXP PCF8883 and PCA8886

lorem-ipsum:
PCA8886 is two PCF8883 in one housing. Schematic is a two part, each
based on the PCF8883. I hope I did that correct.

### [PCF8883](http://www.nxp.com/docs/en/data-sheet/PCF8883.pdf)
The footprint for the WLCSP housing is missing. Seems very special to that chip, so I've given it a vendor / chip specific name.
pcf8883
### [PCA8886](http://www.nxp.com/docs/en/data-sheet/PCA8886.pdf)
pca8886a
pca8886b


jkriege2:
Hi!
thanks for the contribution!
I had a quick look ... mostly this looks fine ... but I have some comments:
1. I wouldmake the symbols overall a bit smaller and sort inputs to the left and control/outputs to the right (see e.g. the ADCs in KiCAD):
2017-08-18 20_49_37-part library editor -- d__kicad_kicad-library-review_library_sensors lib

```
#
# PCF8883
#
DEF PCF8883 U 0 40 Y Y 1 F N
F0 "U" -200 350 50 H V R CNN
F1 "PCF8883" 50 350 50 H V L CNN
F2 "Housings_SOIC:SOIC-8_3.9x4.9mm_Pitch1.27mm" 950 -350 50 H I C CNN
F3 "" -100 -250 50 H I C CNN
$FPLIST
 SOIC*3.9x4.9mm*Pitch1.27mm*
 NXP*WLCSP*PCF8883*
$ENDFPLIST
DRAW
S 300 -300 -300 300 0 1 10 f
X IN 1 -400 0 100 R 50 50 1 1 I
X TYPE 2 400 100 100 L 50 50 1 1 I
X CPC 3 400 -100 100 L 50 50 1 1 B
X VSS 4 0 -400 100 U 50 50 1 1 W
X VDD 5 0 400 100 D 50 50 1 1 W
X OUT 6 400 200 100 L 50 50 1 1 O
X CLIN 7 400 -200 100 L 50 50 1 1 B
X VDD(int) 8 -100 400 100 D 50 50 1 1 w
ENDDRAW
ENDDEF
```

2. I think a pin-length of 100mil is enough (see above)
3. PCF8883 As this is available in two packages that have the same pinout, please empty the footprint field ... the FPFilters seem fine ... Did you also post a PR for the WLCSP-package?
4. PCA8886 mostly correct ... styling I would do as in item 1 + you should change the naming of the pins in unit A ... they should have a 1 instead of a 2 ;-)

best,
JAN

lorem-ipsum:
Finally came around to integrate your recommendations.
I've reduced the pin length and shrunk the overall size following your suggestion. Thinking of it as an ADC this makes sense. The dual package is a bit wider as there is not enough distance between the labels with numbers. And of course the labels in the A unit should have been 1... :-)
I will try to create a footprint for the WLCSP. I've already started it but ran into some difficulties as it is really tiny.

jkriege2:
looks good,
Thanks,
JAN

PS: Let me know if oyu need assistance with the footprint ...

PR #1543, opened 2017-08-17
Add MCP2221

slowcoder:
My first PR, so I might need some extra help.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/20005292C.pdf
The DIP/SOIC/TSOP versions are pin-compatible with each other, but the QFN is not, which is why there's two components in the PR.
mcp2221
mcp2221_qfn

The verification scripts complain about the footprint filters and pin-types, and I can't figure out why, so if someone could enlighten me, that would be appreciated.


slowcoder:
Ok, figured out the pin-type violations (~RST + Inverted symbol). I was trying to be extra clear, but apparently I shouldn't.

Still can't figure out the footprint filters violation.  As for the non-QFN parts DIP footprint, the filter (I think) shouldn't end with *, as it would match incompatible footprints such as _SMD

t3hcatpaw:
Hi @slowcoder,
in `interface.lib`, you have set the footprint filters to
```
$FPLIST
 Housings_DIP:DIP-*_W7.62mm
 Housings_SOIC:SOIC-*_3.9x8.7mm_Pitch1.27mm*
 Housings_SSOP:TSSOP-*_4.4x5mm_Pitch0.65mm*
$ENDFPLIST
```

However, according to an example given in KLC, §4.12.iv.b, they should look like this:
```
$FPLIST
 DIP*W7.62mm
 SOIC*3.9x8.7mm*Pitch1.27mm*
 TSSOP*4.4x5mm*Pitch0.65mm*
$ENDFPLIST
```

Best regards. Dennis

poeschlr:
I would suggest to add one symbol per footprint even if the pinout is the same. This allows you to set the correct footprint in the footprint field.

The footprint filters suggested by @t3hcatpaw are correct. (Do not use the lib name in the footprint filter. Kicad stable ignores such filters.)

poeschlr:
Regarding naming convention: Some time ago we decided to use the Manufacturer part number to indicate the package variants https://github.com/KiCad/kicad-library/issues/1398
So name the symbols according to the following scheme:

- QFN version: `MCP2221-ML`
- DIP (PDIP): `MCP2221-P`
- SOIC: `MCP2221-SL`
- TSSOP: `MCP2221-ST`

For the QFN symbol you need to add the pin for the exposed pad (Number 17) and the two NC pins. (Electrical type not connected, invisible)

james-above:
Ok, cool. Thanks for the help. I'll make the changes and amend the PR.

However, what's the verdict on the filter for the DIP variant?
`DIP*W7.62mm*`
This would match `DIP-14_W7.62mm_SMD`, `DIP-14_W7.62mm_SMDSocket_SmallPads`, etc, which isn't compatible at all.

poeschlr:
Sadly the naming convention for DIP smd packages is not well though out. I think the SMD part should move towards the front. I will open an issue about that. (https://github.com/KiCad/Housings_DIP.pretty/issues/12)
For now we have to live with the fact that your filter will also result in SMD packages to be suggested.

jkriege2:
ABout the pin-count: THat's not a problem, as there is a pin-count filter in place ... sou the only remaining problem would be the SMD-version ... even the differen sockets (also SMD-sockeckts) would be fine.

JAN

poeschlr:
@jkriege2 The problem is not the pin count. All his example footprints have the same pin count as is required. The problem is that by ending the filter with an asterix, the SMD versions of the package appear in the result list.

slowcoder:
Alright, Travis is happy with me at least.

Keep in mind, this is my first PR, so human verification is appreciated. :)

t3hcatpaw:
@poeschlr Maybe a solution - or rather, a temporary dirty workaround - would be to put multiple filters in place, e.g. `DIP*W7.62mm` (note the missing asterisk at the end), which excludes all modified or SMD fp's, as well as `DIP*W7.62mm_LongPads` and `DIP*W7.62mm_Socket`?

poeschlr:
I would rather fix the underlying problem than to introduce a dirty workaround that creates a maintenance nightmare.

t3hcatpaw:
Well, I guess you've got a point there.

jkriege2:
I commented on the issue ... I'm not so sure whether there is a simple&good solution to our problem ... also maybe the SMD-version CAN actually be seen as a variant ... let's see where the discussion leads to ...
JAN

t3hcatpaw:
> also maybe the SMD-version CAN actually be seen as a variant ...

@jkriege2 What you describe in #12,


> On the other hand the SMD-footprints are not totally wrong (also not totally correct), as you can still bend the pins and solder it onto that footprint (In that sense the footprint can be seen as a variant) ...

seems to be the (only) reason why this footprint exists at all, at least in case of DIP:

> [...] To be able to cover the demand [for SMD packages] and sell DIP ICs that had already been produced, manufacturers startet reworking DIPs to SMD packages by bending their pins below the bottom edges outwards at a right angle. An SMD variant with DIL dimensions was created [...]

Source: https://de.wikipedia.org/wiki/Dual_in-line_package#DIL_versus_SMD_bzw._SO (translated from German as accurate as I could)

That's the only thing I could find on the topic of DIP SMD packages so far and there's no citation given either, but it adds to your point.

slowcoder:
It has been 10 days since the last comment. What should I do with this PR ?

Misca1234:
If you make distinct symbol for each package type then you can assign a specific foot print to each of the symbols.
If so, the FP filter is superseded by the dedicated foot print and therefore the format of the FP filter is of less concern.

jkriege2:
Hi!

sorry this took a bit longer ... I just had another look at your symbol. Mostly it's done (I think), but a few things remain to be fixed:
1. please add the package to the description field (e.g. add `, DIP-16`)
2. please shorten the pins to 150mil, or even 100mil if you set the graphic style of RST to LINE and change the name to `~RST`, which will add an overline indicating the negation!
3. please move the end of the NC-pins onto the outline as shown below (reduces the risk of unwanted connection to these invisible pins.
4. Setting Pin Name Positio Offset to 20 instead of 40 makes for a nicer symbol (I think this requirement will be part of KLC in future ...)

here's an example of how the symbol could look:
2017-09-24 13_53_21-part library editor -- d__kicad_kicad-library-review_library_interface lib

jkriege2:
An a propsoed look with 100mil pin-length:

jkriege2:
@slowcoder are you still interested in this?

Thanks,
JAN

jkriege2:
@slowcoder please fix the conflicts!

Thanks,
JAN

PR #1158, opened 2017-04-07
following of PR 1038

jeromebeguin:

PR #1552, opened 2017-08-19
TPS6213x family

evanshultz:
This is the updated fix for https://github.com/KiCad/kicad-library/pull/1158#issuecomment-323351516. I was surprised to see that TPS62130 already existed in texas.lib, but this symbol is better and adds all the PNs in the datasheet.

I made the changes @jkriege2 suggested in that PR:
- Move to dc-dc.lib
- Stack pins
- Change pin length to 100mil

Datasheet: http://www.ti.com/lit/ds/slvsag7e/slvsag7e.pdf

Image:
image


evanshultz:
I got trigger happy and forgot to move the parts. While I was in there I moved some adjacent parts from texas.lib to dc-dc.lib too.

jkriege2:
nice work!
Thanks,
JAN

PR #1556, opened 2017-08-20
Lpt80 a

Misca1234:
ixing LPT80A in opto.lib

http://www.osram-os.com/Graphics/XPic4/00209607_0.pdf/LPT%2080A.pdf

The foot print have been modified 
A 3D model have been created 

kicad-library push branch is
https://github.com/KiCad/kicad-library/pull/1556

pretty push branch is
https://github.com/KiCad/Opto-Devices.pretty/pull/17

packages3D push branch is
https://github.com/KiCad/packages3D/pull/125

https://github.com/KiCad/packages3D-source/pull/50

bild

bild

bild


jkriege2:
Hi!

thanks for the contribution ... could you elaborate a bit why you changed this? The 3D model is really cool (thumbs up!), but the new footprint is lacking a drawing of the actual lens (and thus the direction of detection). Could you bring the latter back?

Also the new symbol deviates significantly from the standard symbols we use for phototransistors (which should be IEC-compliant). Is there a reason for that change? If not, please revert to the old symbol.

For reference, here are the old symbols/footprints:
2017-08-21 00_30_45-part library editor -- d__kicad_kicad-library-review_library_opto lib

2017-08-21 00_28_36-library browser opto-devices c__development_kicad_share_kicad_modules_opto-d

Best,
JAN

Misca1234:
Naaa, it seems that the latest fashion trend within KiCad for symbols seems to be cubics filled with 
foregrounds so I guessed that if I did not update the symbol, it would have been requested :).

But apparently I was a little to fast, 
I change back the symbol and add the "window" on the foot print

Misca1234:
New symbol

bild

jkriege2:
Could you please hide the pin-names (C/E)?

Thanks,
JAN

jkriege2:
Hhmmm ... looking at the 3D model: somehow it looks more complicated than the actual package:
2017-08-21 20_51_26-lpt 80a pdf - adobe acrobat reader dc

Also I have one of these here ... and they are really that simple ;-)

Could you also make the package transparent?

Thanks,
JAN

Misca1234:
Like this 

bild

Let me check if there is a transparent colour in the FreeCad toolbox

Misca1234:
Something like this ?

(There is no transparent colour in the toolbox)

bild

Shackmeister:
Stepup can export transparent colors

jkriege2:
You have two rectangular areas on the front-side sides that stand out a bit (~1mm???). These are actually not there in the real package.

Misca1234:
oh, ok, but I used the image in the top of the datasheet

bild

On this image they are visible, but I guess it is an old image

jkriege2:
Also the package itself is too thick (~2.3mm) vs. 1.45-1.7 in the technical drawing.

Please note that there is already a 3D model for LEDs that are in a similar package: https://github.com/KiCad/kicad-library/blob/master/modules/packages3d/LEDs.3dshapes/LED_SideEmitter_Rectangular_W4.5mm_H1.6mm.FCStd

Best,
JAN

Misca1234:
Perhaps we should use that one instead because I can not see how I get the StepUP macro to export transparent objects

Misca1234:
What do yo uwant to do, use your 3D model or mine ?

jkriege2:
Hhmmm ... do you own one of them? I have a quite old one (~10 years?) ... so maybe they changed the package, although it's not in the DS? If you don't have one with the bars, I woluld propose to go with the existing one ... i.e. simply make a copy of it with the correct name in the correct .pretty ...
OK?
JAN

PS: Sorry for the lonng delay I've been away on holidays ...

Misca1234:
Hi

No, sorry, I do not have an example of it 

Also
...\kicad-library\modules\packages3d\LEDs.3dshapes\LED_SideEmitter_Rectangular_W4.5mm_H1.6mm.wrl
does not become transparent in kicad 3d viewer, nor does the pin lenght match the data sheet

What do you say about this 3d model, I have made a chamfer on top and at bottom and made it a little more narrow.

bild

jkriege2:
looks good to me ... do the measures of the 3D model match up with the sketch in the datasheet? If so, could you update all PRs that contain the 3D model and I can merge all these!

Best & thanks for the work you put in,
JAN

Misca1234:
It is done

jkriege2:
OK!

Thanks for the work (and patience) yo put into this ,-)

Best,
JAN

Shackmeister:
@Misca1234 have you tried selecting a LED material when assigning material properties in stepup? That should give a transparent look :)

Misca1234:
I will check it out next time

PR #1558, opened 2017-08-21
Fix AD637 naming issue

evanshultz:
Fixes an error reported at https://github.com/KiCad/Opto-Devices.pretty/pull/15#event-1213185459. Also move the text for ADuM6000 to be consistent with the rest of this library.


jkriege2:
Thanks,
JAN

PR #1528, opened 2017-08-13
Ltc4054 x 4.2

Misca1234:
THC4052ES5
alias LTC4054XES5

https://www.elfa.se/Web/Downloads/42/xf/405442xf.pdf

bild

PR #1559, opened 2017-08-21
fix LTC4054

jkriege2:
In response to @aliteke79 's commetn there: https://github.com/KiCad/kicad-library/pull/1528

- fix LTC4054 package (SOT-23 --> TSOT-23)
- minor styling
- tags cleanup

@poeschlr could you have a quick look ... should be a no-brainer ;-)


evanshultz:
Thanks Jan! Should the package name in the description be `TSOT-23-6` to indicate the pin count? Similar to how we would write `SOIC-8`?

jkriege2:
fixed ...
JAN

PS: I will be off for holidays for 1-2weeks from tomorrow!

evanshultz:
Thanks Jan!

PR #1217, opened 2017-05-04
KLC cleaning the audio library

maqifrnswa:
Changes:

- added missing datasheet (one symbol)
- added missing footprints and footprint filters (several symbols)
- fixed font sizes on all symbols
- set outlines to 10 mil thickness and background filled (on many symbols)
- moved pins to be on 100 mil grid (on LL1587, LM1875, LM1876, LM1877*, LM3886*) and made pins 50 mil increments

still more work to be done, but it's a starting point

PR #621, opened 2016-07-19
Added TDA7052

jbtronics:
I added the 1W mono audio amplifier TDA7052 to audio.lib . Here is the datasheet for this IC: [http://www.nxp.com/documents/data_sheet/TDA7052.pdf](http://www.nxp.com/documents/data_sheet/TDA7052.pdf)

PR #1560, opened 2017-08-21
Overhaul audio.lib

evanshultz:
I'm overhauling audio.lib, and found https://github.com/KiCad/kicad-library/pull/1217. While most of those comments are about changing parts that stay in the library, there was one comment that was to move some symbols.

I did work to eliminate KLC errors on the parts I moved, but I didn't make an attempt to improve the symbols graphically. If this is required to merge, please advise.

LL1587: http://www.lundahl.se/wp-content/uploads/datasheets/1587.pdf
- Moved to transf.lib
- I also added a datasheet field
- I'll make a footprint (it's on my long list so no FPfilters yet)

NSL-32: http://lunainc.com/wp-content/uploads/2016/06/NSL-32.pdf
- Moved to opto.lib
- I added a default footprint and FPfilters (footprint already exists)
- Also made the rectangle filled with 10mil outline

SSM*
- Suggested to move transistors to transistors.lib
- These are all audio-specific parts and I believe they should stay where they are
- See the part description and applications in the datasheets
- Every audio engineer will recognize SSM* parts as being audio-specific

TR-AUDIO-2P
- Suggested to move it to device.lib
- This appears to be an attempt at a generic LL1587, but I don't see it has value
- It's a generic part meant for a specific mic input application
- I vote to simply removing it

If you believe something different than the above should happen, please state your opinion.


evanshultz:
I ended up having enough time to get through the entire audio.lib before this PR was merged (and probably reviewed). So I changed the title and piled on the updates.

General:
- Changed pin text size to 50mil
- Made all opamp (triangle) symbols the "standard" 200mil size except TDA7293-7296
- Moved pin text offset to 20mil
- Added/corrected default footprints and FPfilters
- Add/fixed all DCM entries
- Mostly split into multiple units for dual, quad, etc. parts
- Combined symbols where lead prep was the only difference and added FPfilters for all packages

LL587:
- Moved to transf.lib

LM1877:
- Combined DIP and SOIC symbols

LM4990:
- Added symbols for all package variants

NSL-32:
- Moved to opto.lib

PAM8302A:
- Added symbols for all package variants

PGA4311:
- Completely reworked the symbol
- Fixed pin names
- Stacked pins

SSM2017:
- Since this is an obsolete part, I didn't add the SOIC symbol

SSM212x:
- I removed the "X" on the left side of the symbol; I can add on all VCAs if you wish
- The level detector symbol for SSM2120 is very basic; I can make it more detailed if needed

SSM22x0:
- Copied symbol style from device.lib
- Changed ref des prefix from U to Q
- Left out the base-emitter diodes

STK*:
- I find a Sanyo datasheet (http://datasheet.octopart.com/STK430-Sanyo-datasheet-107060.pdf) and On Semi one (http://www.onsemi.com/pub/Collateral/ENA2102-D.PDF). The existing symbol is for the Sanyo part so I named it as such. I haven't seen symbols named by the manufacturer before, so please check this.

THAT2180/2181:
- Split into two symbols since their package options aren't compatible

TR-AUDIO-2P:
- Seems to be a useless copy of LL1587, which I moved from audio.lib, and I suggest to remove it
- Otherwise, if it's kept it should be moved to device.lib

Additions:
- LM2876: http://www.ti.com/lit/ds/symlink/lm2876.pdf
- LM380: http://www.ti.com/lit/ds/symlink/lm380.pdf
- LM384: http://www.ti.com/lit/ds/symlink/lm384.pdf
- LM386: http://www.ti.com/lit/ds/symlink/lm386.pdf
- LM4752: http://www.ti.com/lit/ds/symlink/lm4752.pdf
- LM4755: http://www.ti.com/lit/ds/symlink/lm4755.pdf
- LM4766: http://www.ti.com/lit/ds/symlink/lm4766.pdf
- TDA7052 (https://github.com/KiCad/kicad-library/pull/621): http://www.nxp.com/docs/en/data-sheet/TDA7052A_AT.pdf
- TDA7264: http://www.st.com/resource/en/datasheet/tda7264.pdf
- TDA7266: http://www.st.com/resource/en/datasheet/tda7266.pdf
- TDA7266D: http://www.st.com/resource/en/datasheet/tda7266d.pdf
- TDA7266M: http://www.st.com/resource/en/datasheet/tda7266m.pdf
- TDA7266P: http://www.st.com/resource/en/datasheet/tda7266p.pdf
- TDA7293: http://www.st.com/resource/en/datasheet/tda7293.pdf
- TDA7295: http://www.st.com/resource/en/datasheet/tda7295.pdf
- TDA7296: http://www.st.com/resource/en/datasheet/tda7296.pdf
- TDA7297: http://www.st.com/resource/en/datasheet/tda7297.pdf
- TDA7496: http://www.st.com/resource/en/datasheet/tda7297.pdf

KLC violations:
- MSGEQ7 'GND' pin is not actually a ground, but a reference that should by bypassed to ground with a cap
- SSM2165: VOUT is not a power pin
- TR-AUDIO-2P: Missing all doc and footprints, but as I said above I vote to just remove this symbol

Image of at least all unique symbols:
image
image
image

evanshultz:
As there are a number of outstanding footprint PRs, I can understand finishing the footprints first. In case of any footprint name changes, I can update this PR and then merge it cleanly.

evanshultz:
Having now added the Lundahl LL1587 transformer footprint, I updated the symbol. One less KLC violation!

lorem-ipsum:
Hi @evanshultz,

I looked through the library and from a first glance the changes appear to be fine and very much needed. Thank you very much for your work.
But I have a general remark for the usage of git. The way you did the updates they are very difficult to check. This is mostly because your git commits do not follow logical changes, but appear arbitrary. There are commits only deleting footprints, others adding them, and all have the same name. It would be a lot easier for reviewing if you could separate commits into logical changes and name them appropriately. For this (and similar) PRs it would be great if you could have commits either only moving components without changes, adjusting the components to adhere KLC, or adding new components. This would allow to check the changes commit by commit and make reviewing a lot easier.

evanshultz:
Note that there are a number of footprints which may have changed names since I created this PR. Please wait to merge until the footprints are sorted out; I'll make a note here when that's done and I've updated this PR to match.

jkriege2:
Hi!

what's the status of this? What about moving the transistors to transistor.lib?

Best,
JAN

evanshultz:
@jkriege2 Thanks for checking. The status is that I'm waiting for feedback on https://github.com/KiCad/TO_SOT_Packages_THT.pretty/issues/26 to settle the footprint names used in this library. I would like Rene's feedback as well since he's thought about this quite a bit. I'll ping him.

As mentioned above, the transistors are audio-specific and I think they belong here. More info is in my first 2 comments.

evanshultz:
Submitted https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/29 to update TO-220 footprint names. Once that's done I can update this PR to match and then it's ready for review.

jkriege2:
please note that we start a reorganisation of the libs, which entails renaming some libs to make the overall structure more logical. Since we renamed transf.lib to Transformer.lib, please move your parts to that new lib in order to fix the conflicts named above!

evanshultz:
Update footprints to match https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/29#event-1280849317 and https://github.com/KiCad/Housings_SOIC.pretty/pull/31.

evanshultz:
I had to remove LL1587 from the transformer library due to the rename, but now this branch has no conflicts. I will add this symbol back into the transformer library when this is merged.

Please see the comments above, as they cover all 3 KLC violations. This PR can be reviewed and merged!

SchrodingersGat:
Evan once again thanks for your work here, this is a great contribution :)

PR #1565, opened 2017-08-23
Usb3300

slowcoder:
Just a quick update to a component I was about to use. The footprint filter matched a few incompatible packages, so I fixed that.
Also updated the outline width to get it to pass KLC

usb3300-ezk


poeschlr:
Could you update the datasheet link as well? http://ww1.microchip.com/downloads/en/DeviceDoc/00001783C.pdf

The symbol should also be renamed to `USB3300-EZK` and the footprint field should be set to `Housings_DFN_QFN:QFN-32-1EP_5x5mm_Pitch0.5mm`

But this footprint is not exactly correct for this component. The exposed pad is 3.45x3.45mm instead of the recommended maximum of 3.3x3.3mm.
If you are willing to commit a corrected footprint use the new naming convention that we are currently "designing" over at this issue: https://github.com/KiCad/kicad-library/issues/1447

james-above:
Ok, I can make the changes to the symbol (datasheet, name and fp field).

I haven't read up/unstood the KLC for footprints yet, so I think that's better left to a different PR.

However, while I've got your attention, and I have no idea of where to reach a library maintainer in a better forum, I've got a question for you.
There's a few issues where you are discussing changes, like #1447  and #1402 

How do I know which ones are just discussions/ideas, or things I (as a non-maintainer) should take into account?

This symbol (USB3300) is currently in microchip.lib, but issue #1447 indicates that interface_usb.lib (which doesn't exist) might be a better place.

poeschlr:
If you read 1447 you will see that this is a discussion about kicad v5.0.
It will come into affect when kicad v5 is released.

The other discussion is effectively already part of the current interpretation of the KLC. But we have not yet finished it for all possible footprints. (I think it will take a few weeks until we can officially add it to the KLC/FAQ. In the mean time we will inform contributors about this sort of thing if their contribution is affected.)

The renaming of the symbol stems from another recent discussion (which has a few open points) https://github.com/KiCad/kicad-library/issues/1398 This will again enter the KLC/FAQ as soon as all open questions are answered.

Yes this is a tedious process but in a community library the rules should be decided by the community not by a single individual.

slowcoder:
Well, it's a bit of a tedious process, but I understand why it is so.. It's not easy to get a whole bunch of distributed people thinking along the same lines. :)

I have two other questions then.
First, I have a couple of symbols that should go into nordicsemi.lib (as it is right now). Will I create more work for you guys when you'll rearrange everything for KiCad 5.0?  As in, should I hold off on the contributions?
Second, How often do you merge PRs?  When they're "done", or do you have a schedule?

evanshultz:
@slowcoder 
Thanks for contributing!

1. Go ahead and merge them now. KiCad 5 is over the horizon.
2. PRs are merged when they're ready. For schematic symbols, users will require manual work to add them to KiCad but at least they're in the library here.

poeschlr:
Thanks for this fix.
And as @evanshultz already said: your contributions are welcome at all times. (The migration of symbols will be done using a script. So it should not be that bad.)

PR #1566, opened 2017-08-23
Add Nordic Semi multiprotocol SoCs

slowcoder:
nrf8001
nrf51x22-qfxx
nrf52832-qfxx


slowcoder:
Travis is unhappy with a component I didn't modify. The two I added passed, but travis still marks it as a failure..

poeschlr:
No problem.

Your contribution did change the line order for the NRF8001 symbol. (Moved rectangle definition before pin definition.)
This is why travis thinks you changed that particular symbol.
You can ignore this error. (Nobody would be mad at you if you would fix that symbol. But it is not required.)

slowcoder:
Interesting. I really didn't touch that symbol in the previous commits. It must have been the library editor that did that when I saved the library..

Anyhow, I updated the NRF8001 part to follow the KLC. I didn't do a full review of the datasheet (pin-names/numbers and electrical type) though, but at least it's better than it was before I started. :)

Misca1234:
@slowcoder 
That happens to me as well, sometimes it depends on how old the lib is and it seems that newer version of kicad rearrange symbols itself, like sorting and stuff.

t3hcatpaw:
Hi @slowcoder, hi @poeschlr,

hope this PR is still "alive". I would really love to see this merged. :smile:

I've created several other nRF symbols including the nRF52840 (and its footprint) which I would like to add to this lib.

Regards, Dennis

poeschlr:
Thanks @t3hcatpaw for reminding us that this pull request is still open. (I forgot about it. Sorry @slowcoder for the long wait.)

For the nRF52832 the datasheet link does not work. Replace it with: http://infocenter.nordicsemi.com/pdf/nRF52832_PS_v1.4.pdf

The symbols them selves look good. (pin numbers, pin names and electrical types are ok. Pin grouping is done well.)

Maybe the GNDPAD could be named EP (as we use this name in other symbols)
And the NC pin could be placed inside the symbol to reduce the danger of accidentally connecting it. (such that it does touch the outline with its connection point.)

Instead of nRST you could use `~RST` (This would place a line above RST)

The only thing that should really be fixed is the datasheet link. Everything else is a bonus

slowcoder:
Seems like the library editor wanted to move around all the lines in the .lib file, so the diff looks a lot bigger than it actually is.

The Travis warning is for the alignment of the N/C pin on the 8001. Should be fine to ignore.

poeschlr:
That was fast.
Thanks @slowcoder for this great addition.

t3hcatpaw:
Thanks @slowcoder and @poeschlr!

PR #1569, opened 2017-08-24
Add Cortex Debug Connector

t3hcatpaw:
Added ARM's standard Cortex-M Debug Header based on the 20-pin IDC header `Conn_ARM_JTAG_SWD_2` from the same lib.

See http://infocenter.arm.com/help/topic/com.arm.doc.faqs/attached/13634/cortex_debug_connectors.pdf

image


james-above:
Since the component is the header itself, shouldn't pin 4,2,6,8 be considered inputs?  As in, be on the left side of the symbol.

Also, pins 4,6,8 are uni-directional, and pin 2 is bidirectional.

poeschlr:
The data flow is from the programmer to the chip via this connector. From the point of view of the pcb the connector is the source of data. (right side for pins is ok.) 

One could consider to change the electrical types of the programmer pins to be:
- clk: output
- swdio/tms: bidirectional
- TDO and TDI: input or output? (I would need to look into the jdag specification to find out the correct type.)

poeschlr:
Regarding footprint filter:
Why limit the pin header filter to 1.27mm grid?

Why limit the IDC filter to Tag-Connect? (What is Tag-Connect is this a manufacturer?)

t3hcatpaw:
Thanks for your feedback. I was not not sure about making them "active" (because debugging interface) or "passive" (because connector). Not all pins should be considered inputs, though.

What I see now is that my settings are inconsistent anyway because `nRESET` is still an output in this "passive" scenario.

I'm going to change pin types as follows:
- SWDCLK: output
- SWDIO: bidi
- SWO: input
- NC/TDI: input

james-above:
Ha, JTAG is one of my pet interests. :)

TDO is data out from the chip being debugged, TDI is data in to the chip.

I.e, TDI has the same direction as CLK. TDO has the opposite direction.

t3hcatpaw:
@james-above is correct about TDO and TDI, I got a little confused.
So that means
- SWDCLK: output
SWDIO: bidi
SWO: input
NC/TDI: output

james-above:
@t3hcatpaw nRESET is an open-collector, correct. For most cases, "output" would work as well. I wouldn't react to either of those being used for nRESET.

I don't really agree with the TC2050 filter either. TC2050 doesn't define a pin-out, just like a 2x5 header doesn't. Using a TC2050 cable and a TC2050-ARM2010 adapter will give you the pinout you're describing however.

@poeschlr If you wanna make a "ruling" here, here's a site where you can see some pictures to see what we're talking about:
http://www.tag-connect.com/STLINK
On that page, the white ST-Link has a standard 0.1" 2x10 JTAG connector. The little PCB is the TC2050-ARM2010, which connect some of the pins from the 2x10 JTAG to other pins on the 2x05 header on the same pcb. The cable is the actual TC2050 connector. Using these two together gets you the pinout described above.

t3hcatpaw:
> I don't really agree with the TC2050 filter either. TC2050 doesn't define a pin-out, just like a 2x5 header doesn't.

That's true, it doesn't, but I don't see how this opposes a FP filter? It's just an "allowed" footprint... Oh wait, just as I'm writing, I see that
```
IDC?Header*
Tag?Connect*IDC*
```
can be reduced to just `*IDC*`. If that's what you mean, I totally agree, just didn't notice.

slowcoder:
@t3hcatpaw I'm new to the conventions in the official kicad library, so I'm no authority. I do however work with TagConnects and JTAGs (not only as a user, but I build JTAG DAPs as well) daily, which is why I chimed in, so don't put too much weight behind my opinions when it comes to the kicad stuff.

But, I think you're right. I guess I got it the wrong way around in my head, as in "the TC2050 footprint should have this pinout", instead of "This pinout can be used with a TC2050".

t3hcatpaw:
@slowcoder 
> "This pinout can be used with a TC2050"

Thanks, I understand the issue now :) What you said is exactly what I wanted the FP filter to reflect by adding `Tag?Connect*IDC*` earlier or just putting an additional asterisk before `IDC*`, respectively.

t3hcatpaw:
@poeschlr What do you think about changing the filters? I'm ready to push.

t3hcatpaw:
Done :smile:

poeschlr:
The symbol looks ok.
I would rename it though. We have another symbol for the 20pin variant: `Conn_ARM_JTAG_SWD_20`
So i would name this `Conn_ARM_JTAG_SWD_10`.

Regarding footprint filters. I would add `Conn*:*2x*` This would catch all dual row connectors. (In the development version and therefore in kicad v5)

t3hcatpaw:
No problem :wink:

I've followed your suggestions and changed the symbol accordingly. `checklib.py` (at least the version I've been using recently) doesn't like the underscore in `Conn*:*_2x*`. Should I remove it or ignore the error?

poeschlr:
checklib will complain about the underscore in the footprint filter that is correct. But because we don't have a prefix for the row count for such connector footprints we need to include this information. (2x can mean a lot more. Take for example a part that has body dimensions 5.02x10mm. A filter `*2x*` would give a false positive here.)

poeschlr:
I nearly forgot: Thanks for your contribution.

t3hcatpaw:
> checklib will complain about the underscore in the footprint filter that is correct. But because we don't have a prefix for the row count for such connector footprints we need to include this information. (2x can mean a lot more. Take for example a part that has body dimensions 5.02x10mm. A filter *2x* would give a false positive here.)

Thanks, I see. And you're welcome 😄

PR #1570, opened 2017-08-24
LT6233 aliases to LM2904 (as all other-this-kind amplifiers)

dejfson:


evanshultz:
@dejfson 
Thanks for this contribution! Can you please update the datasheet link to something that ends in `.pdf` like other datasheet links? The correct URL is http://cds.linear.com/docs/en/datasheet/623345fc.pdf.

dejfson:
aah. you've managed to find the right PDF. No matter what, I always got only the generic link. Thanks.

evanshultz:
@dejfson 
Thanks! That's the right link. I had to open the datasheet and copy the URL from my browser since the Google search result showed the link you originally attached.

Sorry that I missed it before, but we usually use the full datasheet description for the part. Then add the quantity at the front if it's not in the datasheet description and "Operational Amplifier" at the end. That would mean changing it to `Dual 60MHz, Rail-to-Rail Output, 1.9nV/√Hz, 1.2mA Op Amp`. However, I'll merge without this unless you don't mind changing it.

Lastly, would you be willing to add the other members of this family? The dual in DFN, and possibly even the single and quad? If not, I'll merge this PR. But I won't miss the chance to further grow the KiCad library if you're willing.

dejfson:
Ok, I have changed the description. What concerns the others, I cannot do now, maybe at some later stage. But for sure I will contribute with other components which I need for my work :)

evanshultz:
@dejfson 
Sure. No problem. I'll merge without those extra parts.

I'm ready to merge, but I see that LT6233 has two entries in the DCM file: lines 723 and 837. Can you please remove the first one so the DCM file is alphabetically ordered?

dejfson:
This has to be some 'new' or by me misunderstood feature of kicad. I use git version and recent update somehow changes the formatting of the files. So even if I just open the linear.lib, do some very minor modification and save it, it results in plenty of modifs in that library. I'm not sure what's going on there. I'll remove the entry (by manually editting) and push...

evanshultz:
@dejfson Yes, sometimes KiCad likes to reorganize when a library is saved. Thanks again for your contribution to KiCad!

PR #1572, opened 2017-08-25
Add IDC connectors

wykys:
Add IDC headers connectors: IDC_02X03, IDC_02X05, IDC_02X07, IDC_02X08, IDC_02X10, IDC_02X13, IDC_02X17, IDC_02X20, IDC_02X25.
image


Misca1234:
Are these, in the end, dedicated to a specific type of connectors ?

The standard way of adding connectors to your construction is to use the default connector symbols 
and during the "Run CvPcb to associate components and footprints" you connect the default connector symbols with the desired footprint that represent the connector you want.

If the MLW, connector is a new type of connector that does not previously exist you should create the foot print (and perhaps, but not necessary the 3D model) for the connector and push that instead.

wykys:
This symbol is dedicated to the general connector with the lock. For greater schematics.

MLW is only a business demo of some vendors for IDC connectors.
For example: http://www.ezk.cz/konektory_mlw.htm

jkriege2:
Hi!

first of all: Thanks for your contribution.

But: We try to keep the lib as lean as possible and for such 1/2/3-row connectors we are trying to have one set of connector symbols (with few variants only) that match all footprints as good as possible. See https://github.com/KiCad/kicad-library/pull/1561 for the pull-request for the current set.

Do your connectors have any advantage (besides the look) above the ones I linked? Also the pin-numbering is already there in the existing lib of connectors symbols, so I don't think we will accept your contribution ... 

@evanshultz @poeschlr @pointhi @SchrodingersGat What do you say on this?

Best,
JAN

wykys:
OK, my connectors haven't any advantage then #1561 except lock.

evanshultz:
I would not want to add any more connector symbols since these are already covered. Latching vs non-latching will be handled by picking the desired footprint and should not be part of the symbol.

jkriege2:
OK. Then I will close this PR.
JAN

PR #1576, opened 2017-08-27
Removed the '-' in the FP

Misca1234:
According to 
https://github.com/KiCad/kicad-library/wiki/Status-of-the-libraries

The zetex.lib library was not 100% KLC compliant

This push makes the library 100% compliant, 
the problem was a '-' in the foot print filter


evanshultz:
Good catch and thank you!

PR #1577, opened 2017-08-27
Modified the FP filter

Misca1234:
According to 
https://github.com/KiCad/kicad-library/wiki/Status-of-the-libraries

The silabs.lib library was not 100% KLC compliant

This push makes the library 100% compliant, 
the problem was some '_' in the foot print filter


evanshultz:
Thank you!

PR #1578, opened 2017-08-28
74xx rework (remove hidden power pins)

### LINKED ISSUE #605: 74xx and 74xgxx forced power pins

bobc:
## Purpose

The primary goal is to resolve #605. Because the fix for hidden power pins require many incompatible changes, an opportunity is taken to rework the library to be much closer to current KLC, as well as create a more compact set of symbols.

## Summary of changes
- remove all hidden power pins
  - power pins on unit for single units
  - power pins in separate block if multiple units
- uniform width where possible (600 mil)
- smaller logic symbols (300x300)
- shorter pin length (150 mil)
- fixed several errors in pinouts
- compliance with KLC
  - background color and line width
  - no double inversions
  - footprint spec added
  - datasheet added
- removed single gate ICs
- clean up descriptions, keywords

[74xx test print (pdf)](https://github.com/KiCad/kicad-library/files/1271444/test.pdf)

## KLC waivers

- Some gate symbols (OR gates, NOT gates), require non-standard pin lengths to accommodate shape of the symbol
- some pins are on 50 mil grid due to shape of symbol and desire to keep compactness


poeschlr:
Does your script extract the pin number assignment from the old symbols or is it necessary to check this?

bobc:
Yes, the pin numbers are extracted from existing 74xx library. There were a few symbols which were obviously incorrect, e.g duplicate pin numbers, and some others where I spot checked against data sheet and found errors.

evanshultz:
First, this is awesome. This library was in horrid shape and really needed some attention. Thanks Bob!

Some thoughts after a quick look:
- I still prefer no box around the power pins
- Will the smaller symbols (7400, 74HC04, etc.) not get background fill? Not even ones without arcs?
- Can pin length be shorter? 74HC595 has 200mil pins.
- Do we want to use logic pin types (invert, clock, etc.) or always use a normal Line type? This forces a bigger pin text offset to accommodate pin graphics.
- Some symbols look excessively large. For example, 74LS123. Thoughts on wide vs tall symbols?
- I personally dislike having the pins come out of a corner, like 74LS40, but I can live with it.
- There is a lot of text crashing, but perhaps that's unavoidable with the current symbol file format.
- What will happen to the single gate ICs? Is there another PR to add them to an appropriate lib?

bobc:
> I still prefer no box around the power pins

It was accidental, but the logic gates have an alternate power unit with no box. That feature could be added to all power units.

> Will the smaller symbols (7400, 74HC04, etc.) not get background fill? Not even ones without arcs?

We could fill AND gates but not any type of OR, XOR. The concave arc fills on the wrong side. I think it would look odd if some logic gates were filled but not others. I can experiment with polylines to see if that works OK.

> Can pin length be shorter? 74HC595 has 200mil pins.

Not much I think, 150 is on the limit with a inversion circle and 2 digit pin number. 150 mil then causes a problem with 100 mil grid, so the body has to be bigger/smaller.

> Do we want to use logic pin types (invert, clock, etc.) or always use a normal Line type? This forces a bigger pin text offset to accommodate pin graphics.

I think that's a good question. Personally, I don't find a clock symbol on a pin named CLK gives me anything useful, but generally I kept the decoration already in 74xx. I fixed some cases of inconsistent usage.

> Some symbols look excessively large. For example, 74LS123. Thoughts on wide vs tall symbols?

Ah, 74LS123 is unintentional big, I will fix that. Generally there is a default width of 600 mil, with exceptions for some bigger/smaller symbols. Otherwise I don't have much preference. 

> There is a lot of text crashing, but perhaps that's unavoidable with the current symbol file format.

I wasn't too happy about that, but a consequence of making symbols smaller is there is less space for text - laws of physics. It would be nice to have per-unit fields like Eagle. At least the user can move text to where they want in the schematic.

> What will happen to the single gate ICs? Is there another PR to add them to an appropriate lib?

They are duplicates of symbols already in 74xgxx

bobc:
I found a reasonably efficient way to fill a concave arc using a polyline, so I have updated all the simple logic gates with body fill color.  The polyline has lower resolution than an arc, because the lib file only has integers at 1 mil resolution, whereas KiCad can render arcs at full device resolution, but it seems to give a reasonable result.

I improved the shape of OR/XOR gates, which also stops the pins getting shortened.

I also reduced width of 74LS123.

Updated preview :  [test.pdf](https://github.com/KiCad/kicad-library/files/1261611/test.pdf)

poeschlr:
In your pdf, 74LS51 and 74LS55 do not have background fill. Is this an accident or is this on purpose.

bobc:
Yes, 51, 54, and 55 have some intricate artwork which I copied from the current library. I think I can convert the OR gates to poly lines and fill them, but it is a lot of effort for a few symbols which probably no-one ever uses.

bobc:
I guess I have too much time on my hands :) I wrote a parametric OR gate generator and used it to update the remaining symbols without a filled body.

jkriege2:
Wow! tremendous peace of work :-)))))))))))))

BTW: I coudl you make the AND-symbol for 4 input slightly larger (half a pin-width) to make the look better:
2017-09-01 13_46_26-choose component 178 items loaded

Best,
JAN

PS: I will have a closer look and maybe giove more hints in the process ...

jkriege2:
About the 50mil-grid for gates: IMHO a KLC-exception is fine for this case.

Best,
JAN

jkriege2:
About the OR-gates: can the shape be done with fewer control points? One proposal would (also) be to use this scheme: Make the dark lines with arks and make the filling by having a polygon with fewer control points, filled with background ad using a line-width of -1, which produces polygons with filling, but without lines in my version (is that specified???). Then thad filling polygon could have significantly fewer vertices (3-10 for a quarter arc?), as it only needs to ensure that lines are always covered by the rather thick outlines ...

Best,
JAN

jkriege2:
Two more comments:
1. I think symbols like `74LS109` should have shorter pins (100mil should be sufficient, as long as the pin-number is 2-digit)?
2. Will you also provide the symbol generator (somewhere)? If so I would be really interested to take a look at it!

JAN

bobc:
> Make the dark lines with arks and make the filling by having a polygon with fewer control points, filled with background ad using a line-width of -1

I like that idea! I didn't realize that was possible. I see some strange rendering issues, a 1 mil outline appears and disappears as zoom changes (v4.0.6), but either way it gets hidden by the thicker outline.

> I coudl you make the AND-symbol for 4 input slightly large

I think that is no problem, I can add a parametric generator for AND gates.

> I think symbols like 74LS109 should have shorter pins

No problem. 

> Will you also provide the symbol generator (somewhere)?

There is an early version at https://github.com/bobc/kicad-symgen/tree/master/symgen but its very out of date, I will update it.

bobc:
@jkriege2  I think I addressed all the issues you raised, and also a couple of other things which needed tidying up. Here is latest pdf : 
[test.pdf](https://github.com/KiCad/kicad-library/files/1271438/test.pdf)

bobc:
Is there any more I need to do on this PR?

After this is cmos4000 #1601 , and then I will tackle ttl_ieee and cmos_ieee.

jkriege2:
@SchrodingersGat You merges the 40xx-PR ... if this here is of the same quailty ... could you also merge it?

Thanks,
JAN

SchrodingersGat:
Fantastic work again :) This is a huge improvement, thanks!

PR #1581, opened 2017-08-31
Schematic symbol for RECOM R-78E-0.5 buck converters.

ppelleti:
[Datasheet](https://www.recom-power.com/pdf/Innoline/R-78Exx-0.5.pdf).

Meant to be used with [this footprint](https://github.com/KiCad/Converters_DCDC_ACDC.pretty/pull/21).

checklib.py complains:
```
Checking symbol 'R-78E5.0-0.5':
```

But, this seems to be the same pin arrangement used by the other DC-DC converters.

r78e-symbol


jkriege2:
Hi!

thanks for this contribution! We reworked regul.lib a few weeks ago (introducing a slightly changed, but therein standardized style). Please follow that style, which would mean making your symbols slightly smaller and moving the REFDES and NAME. Here's a screenshot of the two different styles:
2017-09-01 12_58_17-eeschema kicad_review-project2 _ d__kicad_kicad_review-project2

Also the description should be reformulated to match the ones on regul.li, e.g. 
`500mA step-down regulator, 7-28V input, 5V fixed output, replacement for LMxx linear regulators, -40°C to +85°C, TO-220 compatible package`

Otherwise: Very nice work! I already merged the footprint for this, so as soon as you do the changes above, I can merge!

Thanks,
JAN

ppelleti:
Thanks; I've incorporated your suggestions.

r-78e5-revised

jkriege2:
Thanks!

I think you missed `R-78E3.3-0.5` when reworking the descriptions. Please fix and I'll merge.

JAN

ppelleti:
Thanks for catching that!  Should be fixed now.

jkriege2:
nice work! Thanks,
JAN

PR #1583, opened 2017-08-31
Add schematic symbol for OMRON G5LE-1 relays.

ppelleti:
[Datasheet](http://www.omron.com/ecb/products/pdf/en-g5le.pdf).

Meant to be used with [this footprint](https://github.com/KiCad/Relays_THT.pretty/pull/19).

Running `checklib.py -v -c G5LE-1 library/relays.lib` results in:
```
Checking symbol 'G5LE-1':
    Component G5LE-1 has metadata errors:
```

Unfortunately, it doesn't give any clue what the metadata error is.  Any idea what I should do?

g5le-1-symbol


SchrodingersGat:
Use the `-vv` option to display the extra info :)

ppelleti:
Tricky... but thanks!  I've now changed the description to avoid the error.

hackscribble:
Hi @ppelleti just one comment - KLC prefers `*` as the wildcard in footprint filter, so it would become ...

`Relay*SPDT*OMRON*G5LE*`

ppelleti:
Thanks; I've changed the wildcard.

hackscribble:
Many thanks.

PR #1584, opened 2017-08-31
Add library symbol for through-hole WS2812 LEDs.

ppelleti:
There isn't a real datasheet for this device, although the product page is [here](https://www.adafruit.com/product/1938).

This symbol is meant to be used with the existing footprint `LEDs:LED_D5.0mm-4`, and the pinout is chosen accordingly.  I don't think this symbol can be used for SMT WS2812 LEDs, because I believe the pinout is different.

neopixel


jkriege2:
Hi!

looks mostly good to me ... but could you tweak the symbol a bit more please:
1. add a strike-through-line through the LED symbol (so it looks like the usual LED symbols). I think the line may end INSIDE the package (see below).
2. move the arrows so the symbol looks more like existing LED symbols.
3. rename to `LED_Neopixel_THT5mm` so other options (in other packages) can be added and it is clear what is meant
4. change description to `Neopixel 4-color LED, 5mm LED package` or something like that, including the package type
5. FPFilter should be `LED*D5.0mm*`
6. Maybe you cam make the symbol more in line with existing symbols (e.g. text inside ...). Here's a proposal:
2017-09-01 12_42_53-part library editor -- d__kicad_kicad-library-review_library_leds lib
```
#
# LED_Neopixel_THT5mm
#
DEF LED_Neopixel_THT5mm D 0 5 Y Y 1 F N
F0 "D" 100 225 50 H V L BNN
F1 "LED_Neopixel_THT5mm" 100 -225 50 H V L TNN
F2 "LEDs:LED_D5.0mm-4" 100 -300 50 H I L TNN
F3 "" 100 -375 50 H I L TNN
$FPLIST
 LED*D5.0mm*
$ENDFPLIST
DRAW
S 300 200 -300 -200 0 1 10 f
P 2 0 1 0 -70 90 -70 70 N
P 2 0 1 0 -30 90 -30 70 N
P 2 0 1 0 30 40 30 -40 N
P 3 0 1 0 -150 0 30 0 100 0 N
P 3 0 1 0 -30 50 -70 90 -50 90 N
P 3 0 1 0 10 50 -30 90 -10 90 N
P 4 0 1 0 -50 40 -50 -40 30 0 -50 40 N
X DOUT 1 400 0 100 L 50 50 1 1 O
X GND 2 0 -300 100 U 50 50 1 1 W
X 5V 3 0 300 100 D 50 50 1 1 W
X DIN 4 -400 0 100 R 50 50 1 1 I
ENDDRAW
ENDDEF
```

Best,
JAN

PS: Would you be willing to also add the other variants in SMD-package? (optional!!!)

ppelleti:
Thanks!  I've adopted your version of the visual style.

For the name, I have a counter-proposal: `LED_NeoPixel_THT`, because it is also available in an [8mm version](https://www.adafruit.com/product/1734) with the same pinout.  (There isn't currently a footprint for that version, but there could be.)

For the SMD variant, I did a little digging and discovered that it already exists: symbol `WS2812B` in the `WorldSemi` library.

jkriege2:
Hmmm ... for the naming, then we should shorten to `NeoPixel_THT` and add a second FPFilter `LED*D8.0mm*` ... This way a possible future footprint is also matched. OK? The LED I would leave out as we usually don't have such a prefix for symbols.

JAN

ppelleti:
Thanks; I've renamed `LED_Neopixel_THT` to `Neopixel_THT`.

The footprint filters should already have both the 5mm and 8mm version:

neopixel-footprints

jkriege2:
great!

but if you have two distinct FPFilters, the footprint field should be empty (see KLCs). The rest is fine with me and I will merge, when this is fixed!
JAN

ppelleti:
> but if you have two distinct FPFilters, the footprint field should be empty (see KLCs).

Sorry about that.  Should be fixed now.

jkriege2:
Thanks for the great work!
JAN

PR #1585, opened 2017-09-02
Schematic symbol for Adafruit HUZZAH ESP8266 breakout.

ppelleti:
[Product page](https://www.adafruit.com/product/2471).

Meant to be used with [these footprints](https://github.com/KiCad/Modules.pretty/pull/12).

huzzah-symbol


Misca1234:
Just checking, the PCB have 26 connections, but your symbol have 20, 
have you skipped the 6 connections the short end ?

Now, this can be a target for discussion but if you add the character '~' before the R in RST (pin 1) you will get a line above the text RST. If you do then you can use pin graphical style type "input" instead of "inverted"

I guess that you can use a shorter pin length, to 3.810 mm

Also, try to name the pins to what they are named in the datasheet.


When I read the page 
https://learn.adafruit.com/adafruit-huzzah-esp8266-breakout/downloads
I can see the pinout of the PCB and those pins does not correspond with your symbol ?
For example GND pins are not the same ?

ppelleti:
> Just checking, the PCB have 26 connections, but your symbol have 20,
> have you skipped the 6 connections the short end ?

Yes.  The 6 pins on the short end are meant to have male header which faces up, for connecting an FTDI cable for programming.  The other 20 are meant to have male header which faces down, so they are the ones which would go into the PCB.  See [this picture](https://cdn-shop.adafruit.com/970x728/2471-11.jpg).

> Also, try to name the pins to what they are named in the datasheet.

Can you give an example?  I'm going by what the pins are named in [this photograph](https://learn.adafruit.com/assets/24794) from the product page.  I believe I'm using all the same names.  The only thing I changed is that they have a number sign in front of single-digit pin names like #0, and I dropped the number sign to match two two-digit pin names like 15.

> I can see the pinout of the PCB and those pins does not correspond with your symbol ?
For example GND pins are not the same ?

When oriented [like this](https://learn.adafruit.com/assets/24680) and ignoring the 6 pins at the bottom, the GND pins are the bottommost pins on each side.  I numbered the pins like a DIP package, so the left side starts at 1 at the top and goes to 10 at the bottom, and the right side starts at 11 at the bottom and goes to 20 at the top.  This would make the GND pins 10 and 11.

ppelleti:
> Now, this can be a target for discussion but if you add the character '~' before the R in RST (pin 1) you will get a line above the text RST. If you do then you can use pin graphical style type "input" instead of "inverted"

Thanks; I've implemented that suggestion.

> I guess that you can use a shorter pin length, to 3.810 mm

OK, I've decreased the pin length to 150 mils.

jkriege2:
Nice work!

Thanks,
JAN

PR #1587, opened 2017-09-02
Add TSIC 206 and alias TSIC 306.

Schlumpf:
Adds the TSIC digital temperature sensor.

image

Datasheet: https://shop.bb-sensors.com/out/media/Datasheet_Digital_Semiconductor_temperatur_sensor_TSIC.pdf

I'm not sure about the footprint filters, please check.


jkriege2:
Hi!
thanks for the contribution! As the two package variants have different pinout, you'll have to split the symbol into a SO-8-variant and a TO-92-variant (ALIASes will not work for this perpose). Then set the footprint fields and FPFilters as listed below:
- TO-92-variant: name `TSIC206-TO92`/`TSIC306-TO92` (as per first page of DS), footprint: `TO_SOT_Packages_THT:TO-92_Inline_Narrow_Oval`, FPFilter: `TO?92*Inline*Narrow*Oval*`
- SO-8-variant: name `TSIC206-S08`/`TSIC306-S08`, footprint: `Housings_SOIC:SOIC-8_3.9x4.9mm_Pitch1.27mm`, FPFilter: `SOIC*3.9x4.9mm*Pitch1.27mm*`

Best,
JAN

jkriege2:
PS: Add the package to the descriptions, as `, SO-8` and `, TO-92` as postfix to the existing descriptions!

Thanks,
JAN

jkriege2:
PPS: One last thing: the SO-8-symbol will need all the NC-pins present (as hidden NC-pins!), so the pin-cpunt filter can match up the correct package!

Best,
JAN

Schlumpf:
Thanks for the feedback. Done 🙂
image
image

jkriege2:
Thanks for your great work!

JAN

PR #1588, opened 2017-09-03
Microduino core

nerdyscout:
This pull request contains all parts of Microduino (a variation of an Arduino) including footprint, library and a template to use it as new project to start from.

for further information about microduino see [here](http://wiki.microduinoinc.com/Microduino-Module_Core).

Part A
image
Part B
image
Part C
image

Footprint
image

### DIFF
--- a/library/modules.lib
+++ b/library/modules.lib
@@ -191,8 +191,8 @@ ENDDEF
 DEF Microduino-Core U 0 40 Y Y 1 F N
 F0 "U" -50 1050 50 H V R CNN
 F1 "Microduino-Core" -50 950 50 H V R CNN
-F2 "Modules:Microduino-27upin" 100 -1100 60 H I L CNN
-F3 "http://wiki.microduinoinc.com/Microduino-Module_Core" 100 -1200 60 H I L CNN
+F2 "Modules:Microduino-27upin" 100 -1100 50 H I L CNN
+F3 "" 100 -1200 50 H I L CNN
 $FPLIST
  Microduino*
 $ENDFPLIST


Misca1234:
Nice, I have made a 3D model of 

bild

When this is merged I can make a push for the 3D model

Misca1234:
Do you have a link to you foot print push as well ?

nerdyscout:
ohh.. I am sorry, the footprint should be already part of this push. As I am not very experienced with git I have no clue how I can split this to a separate one now.

Misca1234:
Sorry, my mistake. I was single minded on "normal" pushes, these "module pushes" is different from those

jkriege2:
Hi @nerdyscout and thanks for your contribution.

Please be aware that we have come up with a set of rules for symbols and footprints in this lib. They are called KLC (KiCAD Library Conventions) and you can have a look at them here:
- https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention
- https://github.com/KiCad/kicad-library/wiki/Frequently-asked-questions

We only accept contributions that comply with these rules, but we are always happy to help getting contributions there.

I had a look at your design in this PR and I have some suggestions:
I think it would be better to split this one up a bit:
1. Have a PR that adds a symbol for this Arduino-variant to modules.lib in the standard lib, so the Arduino can also be used outside this template. This should then represent an arduino mounted into female connectors soldered to the PCB from the top. Make sure that your PR follows the style for symbols in this lib (have a look at the existing Arduinos in the lib for a sample, also I think that this small Arduino is simple enough to not split it up into three subcomponents, but that may be discussed if there are good arguments for having the submodules).
2. You then need a second PR for the actual footprint, which should go to https://github.com/KiCad/Modules.pretty. Make sure this footprint also conforms to the KLC, again have a look at the exsiting Arduino footprints in the lib I referenced. As the Microduino contains male and female connectors in the same board, it might make sense to have two footprints, one for each variant!
3. Now you can use these general symbols/footprints in your template. When doing the template, please limit the linked libs to what is absolutely necessary (otherwise the template is easily outdated, when some libs are renamed ... which is planned for the near future!). Also make sure to note in the template, which variant of connection (female or male connector in Microduino) should be used.

Let me know, if you need any help in the process!

Best,
JAN

nerdyscout:
Hello @jkriege2,

thank you very much for your feedback.
I made now several changes to comply with the KLC and some of your suggestions.

I agree the symbol does look much more tidy now.
image

As all Microduinoparts are delivered with already soldered connectors, which are female on top and male on the bottom at the same time, I would not state this separately in the library or the footprint.
The footprint is named "Microduino 27upin" which does does comply to their
[specification](http://wiki.microduinoinc.com/Description_of_Microduino_Upin27_Pins). Therefore the socket is more generic and could be used for other Microduino modules as well.

Currently I don't really understand how I can split this PR into two... seems like I have to learn much more about git. Is it possible to do a PR about a single commit instead of a whole branch? Otherwise I don't really have a clue how this should work.

thx!

Rantanen:
> As all Microduinoparts are delivered with already soldered connectors, which are female on top and male on the bottom at the same time, I would not state this separately in the library or the footprint.

As you said, Microduino has two ways to connect by default - using either the female sockets or male pins.

I believe the idea was to provide two footprints for this - one for the female connectors and one for the male connectors. This would allow the user to choose between mounting the microduino on a board into a female socket versus creating a 'shield' for the microduino using male pins. You can still have one symbol.

In any case! The reason I butted in wasn't that.

> Currently I don't really understand how I can split this PR into two... seems like I have to learn much more about git. Is it possible to do a PR about a single commit instead of a whole branch?

No. Git pull requests always concern a full branch.

There are few ways to split branches up. The different ways have different benefits. Here I would opt for redoing the whole branch. This is the simplest way to choose which files to commit, but you will lose the individual commit history.

```
# Create a new branch based on the current master.
git checkout -b microduino-modules origin/master

# Bring the latest 'Microduino-Core' changes on top of the new branch.
git merge Microduino-Core

# Reset the current branch history back to the origin/master.
# This drops the merge commit from the history, but leaves
# the working copy in a merged state.
git reset origin/master

# Check which files have changed, add necessary files and make a new commit.
git status
git add ....
git commit
```

Once this is done, the `microduino-module` branch should contain only the changes to the specified files. You can push this branch and send a PR of it.

Once that is done, you can check out this branch again with `git checkout Microduino-Core`. You can then clone the [kicad/Modules.pretty](https://github.com/kicad/Modules.pretty) repository to another working directory and copy the footprint file from under the templates to the `Modules.pretty` repository. At this point you could also create the second footprint, which has the Microduino mounted through the female pins onto male pins (as if, into a shield) and include that in the Modules.pretty repository as well. Once you have these files in the Modules.pretty, you'd git add, commit and push these and finally make a pull request into [that repository](https://github.com/kicad/Modules.pretty).

Then as a final step, you'd go back to the `kicad-library` repository and using the steps above, you'd split the remaining `Microduino-Core` changes into `Microduino-template` branch - this time selecting the template/ files, instead of the library/ files. Then make necessary changes to these templates so that they refer to the footprint in the Modules.pretty, push and make a pull request.

nerdyscout:
somehow I managed to fuckup completely my repository. sorry but i can't resolve it. Maybe I will come back later with a net pull request.

PR #1590, opened 2017-09-03
Lm2776

Misca1234:
The author of push 
https://github.com/KiCad/kicad-library/pull/1336
have not responded in 2 month so this push is a replacement

http://www.ti.com/lit/ds/symlink/lm2776.pdf

bild


evanshultz:
@Misca1234 
Thanks for finishing off this seemingly abandoned PR. It can be closed one this is merged.

Some quick thoughts:
- Are the other changes in the files intentional?
- This symbol might look better with 20mil pin text offset spacing instead of 40mil.
- How does this look with a cap between C+ and C-? It might be better to push those pins a bit farther apart so a cap symbol fits nicely there.
- Make the package `SOT-23-6`, retaining the number of pins.
- Keywords shouldn't be a copy of the description

Misca1234:
>     Are the other changes in the files intentional?
No, It seems that my kicad did that on its own, shall I redo the push from the master ?

>     This symbol might look better with 20mil pin text offset spacing instead of 40mil.
Fixed
>    How does this look with a cap between C+ and C-? It might be better to push those pins a bit farther apart so a cap symbol fits nicely there.
See screen shoot
However, in the original push, one of the librarians wanted the cap connections on the right side so I putted them there.


    Make the package SOT-23-6, retaining the number of pins.
See screen shoot
  >   Keywords shouldn't be a copy of the description
See screen shoot

bild

bild

evanshultz:
Pushing again from Master may be best, if that avoids any other changes. Otherwise, you could only add this symbol into your PR's branch. Either option is OK to me if you can avoid other changes in the library.

The `EN` pin should stay on the left side since it's an input. I only mentioned spreading C+ and C- apart if it helps with wiring in a capacitor.

Lastly, let's use the datasheet description and clean that up a bit. Please change to `Switched capacitor Inverter, +2.7 to +5.5V Input Voltage, 200 mA, SOT-23-6` or something similar if you have another suggestion.

Misca1234:
https://github.com/KiCad/kicad-library/pull/1636

PR #1591, opened 2017-09-04
MAX1112 & MAX1113

lorem-ipsum:
This adds the MAX1112 (8-channel) and MAX1113 (4-channel) ADCs to the adc-dac library.
[Datasheet](http://datasheets.maximintegrated.com/en/ds/MAX1112-MAX1113.pdf)

I am not really sure about the type of the COM and REFIN ports. Currently both are set to power input. I am not sure whether that is correct for voltage references.

max1112
max1113


jkriege2:
Hi!

nice work!

Two small requests though:
1. please add the keyword `ADC` to the tags
2. please make pins shorter ... 100mil (as in most ADCs/DACs in the lib) should be enough.

Thanks,
JAN

lorem-ipsum:
Hi JAN,

thanks for the review, I've adjusted the schematics as requested.

Cheers,
Hendrik

evanshultz:
I think it would look better if the pin text offset was moved from 40mil down to 20 or 10.

lorem-ipsum:
I agree (at least with 20), but I am not sure about changing it for this specific component. A general standard as requested in your issue seems to be the better approach.

jkriege2:
Since there is currently no real standard, change to a setting that looks visually pleasing to you ;-)

Let me know, ehn done so I can merge! (I often also use 10, especially in small symbols like VRegs ;-)

Best,
JAN

lorem-ipsum:
I've changed it to 20. If you're happy with them you can merge

jkriege2:
Thanks, Great work!
JAN

PR #1594, opened 2017-09-05
Fix MPU-9250 AUX_SDA

smallzzy:
According to page 19 of the [datasheet](https://www.invensense.com/wp-content/uploads/2015/02/PS-MPU-9250A-01-v1.1.pdf), the AUX_SDA pin should be pin 21 instead of 6.

While I am at it, I also fix the pin number in the footprint filter to accommodate the convention.


hackscribble:
Thanks for the fix.

PR #1595, opened 2017-09-05
Move, improve MAX3182x temperature sensors

hackscribble:
This is a replacement PR for closed PR https://github.com/KiCad/kicad-library/pull/1316

Scope:
- move MAX31820 and aliases, MAX31820PAR and aliases and MAX31826 from `maxim.lib` to `sensors.lib`
- improve the symbols and make KLC compliant

DS2401 was an incorrect alias for MAX31820PAR (it is an ID chip, not a temperature sensor). It has been left as a standalone unchanged symbol in `maxim.lib`.  **EDIT** KLC errors now fixed in it.

@jkriege2 I have made the last changes you suggested in the old PR ...

> 1. set footprint of MAX31820 to TO_SOT_Packages_THT:TO-92_Inline_Narrow_Oval
> 2. MAX31820PAR: footprint as above and please add the NC pin 3 as invisible NC-pin
> 3. add the package-name to all description fields (e.g. add , TO-92 to MAX31820 ...)

I will raise a separate PR for the other moves that you requested ...

> 4. DS1822Z, DS18B20U, DS18S20Z, DS18S20Z are still in maxim.lib although other DS1822/...-variants have been moved ... is there a reason for that?
> ...
> 6. would you also move DS28EA00 to sensors.lib, as it is the last remaining thermometer in maxim.lib as far as I can see?

image
image
image


jkriege2:
nice work!

Thanks,
JAN

PR #1460, opened 2017-07-28
Overhaul digital_audio library

evanshultz:
Overhaul of this library with the following changes:
All:
- Standardize description and keywords
- Add package name to description
- Update datasheet links if needed
- Correct/add default footprint if needed
- Correct/add FPfilter if needed
- Set all line thicknesses to 10mil, if not
- Shorten all pins to 100mil
- Move all pins on 100mil grid
- Move ref des and value above symbol
- Stack power and ground pins

AKM*:
- AK5392: Add datasheet link (from DigiKey)
- Update footprint to new 18.7mm tall package

CS4245:
- Move pins so in on left and out on right

CS43L21:
- Move FILT+ up 50mil to be on 100mil grid
- Change ANGD pin to type POWER INPUT
- Add missing pins (NIC and TSTO)
- Add pin 33 for QFN package

CS5361:
- Move audio pins to top for better layout
- Show RST pin as active low

CS8406:
- Add missing pins
- Change pin names for both HW and SW modes

CS8414:
- Adjust pin positions to fit symbol
- Add datasheet

CS8416:
- Consolidate into just SOIC/TSSOP and QFN symbols
- Update datasheet link
- Add pin 29 for QFN package

TLV320AIC23BRHD:
- Add pin 29 for QFN package

TLV320AIC23BxQE:
- Create symbol for datasheet package not in lib based on existing symbols ([PR for new footprint](https://github.com/KiCad/Housings_BGA.pretty/pull/17))

TPA5050:
- Fix datasheet URL
- Fix broken FPfilter
- Add pin 17 for QFN package

PR #1597, opened 2017-09-06
Digital audio footprint fix

evanshultz:
While reviewing the merge comment at https://github.com/KiCad/kicad-library/pull/1460, I checked to make sure all footprints were in the library. I didn't find any missing, so that's good.

I did spot a typo, though. This should be a simple thing to review and merge.


SchrodingersGat:
It was a simple thing to review and merge.

PR #1601, opened 2017-09-07
cmos400 rework, remove hidden power pins, better KLC compliance

bobc:
## Purpose

The primary goal is to remove hidden power pins (#605). Because the fix for hidden power pins require many incompatible changes, an opportunity is taken to rework the library to be much closer to current KLC, as well as create a more compact set of symbols.

[demo_cmos4000.pdf](https://github.com/KiCad/kicad-library/files/1284333/demo_cmos4000.pdf)

## Summary of changes

  - power pins on unit for single units
  - power pins in separate block if multiple units
- uniform width where possible (600 mil)
- smaller logic symbols (300x300)
- shorter pin length (150 mil)
- fixed several errors in pinouts
- compliance with KLC
  - background color and line width
  - no double inversions
  - footprint spec added
  - datasheet added
- clean up descriptions, keywords

## KLC waivers

- Some gate symbols (OR gates, NOT gates), require non-standard pin lengths to accommodate shape of the symbol
- some pins are on 50 mil grid due to shape of symbol and desire to keep compactness


SchrodingersGat:
@bobc these look fantastic. Thanks for the effort you'd made here

PR #1602, opened 2017-09-08
Add Microchip MCP2200 USB-to-UART serial converter

Heiko-L:
checklib.py throws no errors, I have given my best ;)
mcp2200
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/200022228D.pdf


jkriege2:
Hi!

nice work!

Thanks,
JAN

PR #1603, opened 2017-09-08
Fixed Errors for ENC28J60 (and added few aliases)

Heiko-L:
checklib.py throws no errors, I have given my best ;)
grafik
Datasheet: http://ww1.microchip.com/downloads/en/devicedoc/39662e.pdf


jkriege2:
Which errors did you fix exactly?

Overall this looks good and I will merge soon. One comment: We usually don't add the different temperature variants (but use an `x` as wildcard to hide them in the symbol name), could you do that for these symbols too? 
Thanks,
JAN

Heiko-L:
> Which errors did you fix exactly?

- There were not allowed characters in the names (changed "/" to "-")

- Updated link to datasheet

- added ENC28J60-SS, ENC28J60T-I-ML(alias), ENC28J60T-I-SO(alias), ENC28J60T-I-SS(alias), ENC28J60T-SS(alias)

- added default footprints and footprint filters

- Power pins were off 100 mil grid -> corrected by shortening pins to 100mil and slightly change symbol outlines

Sorry, I'm new to this, this information had better been placed in the commit...

> One comment: We usually don't add the different temperature variants (but use an x as wildcard to hide them in the symbol name), could you do that for these symbols too?

Sure I'll push that to this branch in my forked repo.

Heiko-L:
Hmm, now I have to ask- there exist the following types of this chip:

- ENC28J60-SS
- ENC28J60-I-ML
- ENC28J60-I-SO
- ENC28J60-I-SP
- ENC28J60-I-SS
- ENC28J60T-SS
- ENC28J60T-I-ML
- ENC28J60T-I-SO
- ENC28J60T-I-SS

The "T" stands for tape and reel packaging, "I" stands for industrial temperature range. So I think if you don't add different temperature variants that applies also to different packaging variants?

So, the library should look like this:
- ENC28J60x-x-SS
- ENC28J60x-x-ML
- ENC28J60x-x-SO
- ENC28J60x-x-SP

Or did I get something wrong?

SchrodingersGat:
Yup, the package should be a wildcard (x)

Heiko-L:
So I reduced it to four symbols using only one wildcard, looks prettier that way.

jkriege2:
nice work!
Thanks,
JAN

PR #1605, opened 2017-09-09
motor_drivers: DRV8833, L297 and TB6612FNG

Lichtso:
drv8833
l297
tb6612fng


SchrodingersGat:
Hi @Lichtso I have spotted a few issues with these symbols:

## DRV8833

The three footprints for this symbol are incompatible:

image

1. The HTSSOP and WQFN have exposed central pad which would need to be added to the symbol
1. The TSSOP does *not* have exposed pad
1. The pinouts are different between packages

You will need to make a symbol corresponding to each footprint. Use the package names as given in the datasheet:

1. HTTSOP - DRV8833PWP
1. TSSOP - DRV8833PW
1. QFN - DRV8833RTY

## L297

Looks good

## TB6612FNG

As there is only one possible footprint for this symbol, please fill out the footprint field with the correct footprint. Check if the correct footprint exists in the `Housings_SSOP` library.

Lichtso:
What pin number should the ground pads have? 17 for a 16 pin housing?

Because I could not find any footprints for:
HTSSOP-16*
WQFN-16*

SchrodingersGat:
Yes you should use pin 17.

If you cannot find a suitable footprint then are you willing to submit this too?

e.g. to [here](https://github.com/KiCad/Housings_DFN_QFN.pretty)?

Lichtso:
I added PRs for both footprints and updated the FPLIST in this PR.

SchrodingersGat:
Thanks for your contributions :)

PR #1608, opened 2017-09-10
Add Maple Mini Microncontroller module by LeafLabs

Rantanen:
Schematic and the footprint (in a separate pull request to the modules.pretty repository) for the [LeafLabs Maple Mini](https://cdn.sparkfun.com//assets/parts/6/9/3/3/11280-01c.jpg) microcontroller module. The module is somewhat discontinued by LeafLabs, but thanks to it being open source, the clones are still being manufactured and given the STM32 on board, make for rather powerful alternatives to the usual Arduino-solutions.

Here's an embedded picture of the symbol for a quick glance:

maple_mini_sch

As this is my first pull request, there's some possible issues that I wasn't sure of. These concern both the schematic and the footprint so I'll open the discussion for all of it here just to keep it all in one place. I'll refer this PR from the modules.pretty repository PR later. Also I feel like I should apologize for the length of the text. I love consistency, so I'm trying to learn how to interpret KLC in the best way to make the parts fit together with the existing KiCad parts.

Also I'm primarily a software engineer, so apologies beforehand for probably screwing up most of the electronics bits here. :)

## 1. Reference designator

Probably the easiest issue to solve! What would be the preferred reference designator. I saw the Arduino parts were using 'A?' so I went with 'M?' for now. No idea if M has an alternate meaning. Was also playing around with the generic 'U' designation.

## 2. Bidirectional power pins

Maple Mini, just like Arduino, has several different power pins. The Vcc pin is tied to the STM32 Vcc pin and should be at 3.3 V while the Vin goes through a built in regulator. The module can be powered up with voltage to Vin at which point Vcc can be used as +3V3 output - or the module can be powered up directly through the Vcc pin with regulated +3V3 voltage.

Currently the part defines Vin as Power Input and Vcc as Power Output - following the example set by the various Arduino symbols in the modules library. However I'd much rather define both as Power Input as the +3V3 is very limited in its output capabilities for larger voltages on Vin and at least personally I prefer using the +3V3 as a power input pin to drive the module, which allows me to use a better regulator.

## 3. Pin 1 indicator

I was going through the existing parts and saw a lot of different designs for the pin 1 indicators. Is there a preferred way to indicate pin 1? The four most common designs are displayed in the picture below. Currently the footprint uses the leftmost one, as I'm expecting the Maple Mini to be soldered using pin headers and the leftmost design is what the pin headers use currently.

pin1s

## 4. Silkscreen offset

What is the preferred offset between the actual part dimensions and the silkscreen? Based on KLC, the silkscreen should be drawn outside the part - but there seems to be differences in how much outside the silkscreen is drawn. Pin headers (which I again used as an example) draw the silkscreen so that it touches the part. A close up of Maple Mini footprint is displayed above. The yellow F.Fab layer should follow the physical dimensions of the module and the F.SilkS layer is offset by half the line width outwards - this should make the inner edge of the F.SilkS lines to follow the physical dimensions. This is how the pin headers seem to be drawn. However some other parts have a visible gap between the F.Fab and F.SilkS graphics.

maple_mini_kimod_silkscreen

## 5. Footprint orientation (also Pin 1 orientation)

Maple Mini doesn't have a manufacturer defined "Pin 1" so I had to make a choice on my own. I ended up going with the Vcc (+3V3) pin next to the USB connector as Pin 1. The other option would have been to use the Vin pin at the opposite corner as Pin 1.

### Vcc as Pin 1 (USB on top)

- GND becomes Pin 2
- USB connector on top seems to be the more common way to picture this module in various pinout graphics, etc.
- The silkscreen texts are oriented USB connector on left/top. (Ironically the Chinese clones have one text USB on top and the other vertical text USB on bottom...)

### Vin as Pin 1 (USB on bottom)

- Vin is the more generic "power input" - Vcc can be used as power input, but requires regulated +3V3 voltage.
- Existing Arduino Nano layout uses USB connector on bottom orientation.

I didn't even think of making the layout horizontal given all the DIP/SSOP/etc. packages have the pins in vertical lines.


Rantanen:
Also a pre-emptive question concerning pin lengths. I've seen several comments in different PRs advocating shorter 150 mil pins. How would one lay out the component with 150 mil pins while still maintaining the 100 mil grid for pin origins?

image

I've seen variations of 0, 50 and 100 mil margins. The 0 mil margin has an issue with the inverted Reset pin where the line above the text collides with the symbol edges. Personally I'd like the 50 mil margin best but unfortunately it's incompatible with the 150 mil pins on the 100 mil grid.

And sorry for probably making up issues as I go. I'm just a sucker for consistency. :)

(Also I'm growing to dislike the very single sided "all IO pins on the left side" layout of the current symbol here. I'm kind of wishing someone complains about it here so I can move aV+, aV- to the left side and D15-D32 to the right side. This would result in somewhat balanced pin counts and would end up mirroring the physical sides of the pins on the module - even if somewhat reversed. The big question with that layout is on how to name the pins. I would have done it already but thanks to the pin functions,  the Dxx numbers won't line up - unless the pin names on the right side are changed into FUNC/Dxx instead of Dxx/FUNC... Are the pin names cosmetic for the symbol or are they used for something else as well? >_>)

Rantanen:
Actually given how much I dislike the current single sided design, I went ahead and tried a double sided design:

comparison

Also capitalized the pin names (although can do this on the current symbol just as well). Moving the RESET pin at the bottom solved the symbol edge overlapping the inversion overlining.

The one thing that I feel might be bad is the naming of the right side pins that I referred to in the previous comment. The left side pins have the Dnumber first and the alternate function second. The right side pins have the function first and Dnumber second - this was done so that the Dnumbers would line up.

As long as the pin names are used only for visual purposes, I feel this would be okay - if the pin names appear in lists somewhere, then this might be bad for consistency.

Haven't pushed these changes into this branch. Still waiting for some kind of okay from whoever is reviewing this at some point.

Rantanen:
And again responding to myself.

Concerning the alternate of driving the module with either 3.3 V to Vcc or up to

SchrodingersGat:
@Rantanen wow this is probably the most comprehensive first-time submission I have seen! Please continue to contribute 👍 

# 1. Reference Designator

Please use `A?` as the RefDes for this. `M?` is reserved for motors (etc). You can refer to the general [RefDes guidelines here](https://en.wikipedia.org/wiki/Reference_designator)

# 2. Bidirectional Power Pins

You can define these as power inputs as the ERC (electrical rules check) allow for multiple power inputs to be connected together, while connecting multiple power outputs could produce an error. In this case where there is some ambiguity, I think "input" is easier.

# 3. Pin 1 indicator

From your drawing, 1) or 3) are the most acceptable. We need to formulate some better examples for this rule.

# 4. Silkscreen offset

As long as the silkscreen is visible after loading the part, that's fine. This is more important for SMD footprints where the presence of silkscreen could impede loading and registration of the part.

If you want to have a small gap between the part and the silkscreen, that's fine. We also require at least 0.2mm clearance from pad+mask.

# 5. Pin-1 Orientation

Good choice. Vcc pin next to USB connector is the obvious candidate to me.

# Double sided symbols

Your double sided symbol is much better :) The "all inputs on the left" rule (like many of the KLC) is a guideline and there are use-cases where it is best ignored.

# And again responding to myself

I think a single symbol with Vcc as `power input` would be fine.

---------------------------

If you want to push the double-sided symbol changes, I'll review :)

Rantanen:
Here is the updated symbol. Changed to double sided, changed RefDes to A and made VCC a power input.

The remaining question for me is whether the naming of the pins on the right side is okay. It looks cleaner to me with the Dnumbers lining up but differs from the left side.

### Symbol

image

image

Footprint filter is `Maple*Mini*`.

However... now that I looked at the filter, I came to the realization that this is essentially a 15.24 mm wide DIP-40 package. I think I'd still prefer a separate footprint for this so we can have the USB connector visible in the F.Fab layer and the F.CrtYd changes that implies. However should the `DIP-40*W15.24mm*` be given as an alternate filter? Not sure if this makes a difference in case someone wants to use a DIP40 socket or a ZIF socket for example.

### Footprint

image

image

3D Settings refer to `${KISYS3DMOD}/Modules.3dshapes/Maple_Mini.wrl`, which is going to be a missing model - I actually have a 3D model for this, which can be seen in http://jubjubnest.net/~wace/n2k/n2k-dashboard.front.png, but as it's made with KiCad's wrl export there's few issues with it:

- The height offset is incorrect, as KiCad want's to export it with the PCB on zero level while it should be used with the pin header bottom as zero. This is the biggest reason I'm not including the 3D model in the PR. I might end up fixing that with a wrl editor at some point.
- The solder mask is wrong. Maple Mini has a red solder mask. The wrl export insists on green.
- The model uses a Micro USB connector - the real one has Mini USB, but KiCad didn't have a 3D model of that so I substituted it with the Micro USB just so there would be some connector on it.
- The empty pads at the end are push buttons, but no 3D model for those.

SchrodingersGat:
Footprint filter is fine, if users want to explicitly force a different footprint then they can. I agree there is value on adding the USB connector to F.Fab for example.

Your pin naming is also good (lining up the Dxx numbers for e.g.)

The 3D model can remain "missing" for now but it's good to put the link to it now so that the change does not need to be made later.

Thanks for your contribution, I'll merge the symbol :)

PR #1610, opened 2017-09-11
Potentiometers

feichtinger:
Hi there

I found out that the current symbol POT does not really tell where the start (left) and end (right) positions of the potentiometer are, so I think it would be helpful to show the pin numbers. Also I think we should stick to the convention:
1: start
2: slider
3: end
as far as possible. I think in most symbols and footprints this is already the case.

Also i fixed the pin numbers at POT_Dual_Separate to fit to the other symbols, the were completely messed up...

bildschirmfoto vom 2017-09-11 10-11-11


jkriege2:
Hi!

just a quick note: Please check the pins in the dual-pot again ... they are 123 now in both units (I think). Easiest to edit this is using a text-editor on the .lib-file directly.

Best,
JAN

PS: I will have a closer look this evening!

feichtinger:
> Please check the pins in the dual-pot again ... they are 123 now in both units (I think). 
Only the pin names (which are irrelevant for this symbol), not the numbers. Fixed it anyway.

jkriege2:
Thanks,
JAN

feichtinger:
I also started to fix some footprints: [](https://github.com/KiCad/Potentiometers.pretty/pull/23)

PR #1611, opened 2017-09-11
ADA4870ARRZ - high power high speed amplifier

dejfson:


evanshultz:
Hi @dejfson 

Thanks for another contribution.

As you've been told at https://github.com/KiCad/Housings_SSOP.pretty/pull/41 and https://github.com/KiCad/Relays_SMD.pretty/pull/1, always include a screenshot and link to the datasheet (or other documentation) when submitting a PR.

Review comments:
- Remove "57MHz, " and "Operational " from the description to align with the datasheet.
- Use the standard triangle opamp symbol. It can fit the 3 control pins.
- Pin 8 (output) should be Output type.
- The thermal pad is connected to ground at http://www.analog.com/media/en/technical-documentation/user-guides/ADA4870ARR-EBZ-UG-685.pdf and I don't see any clear documentation on this in the datasheet. Are you able to find out the possible connection(s) of the thermal pad?
- Unless this reveals that the thermal pad can totally float, it should be of type Power Input.

dejfson:
Hi, ok i'll do the changes tomorrow (hopefully). The only one not clear to me is how to fit the pin names to the ordinary triangular opamp symbol. I've been searching for one before in kicad libs, but all the symbols with 3 additional pins were with pin names invisible, and i think it is not what we should want in this case right?

evanshultz:
Pin names visibility is done on a case-by-case basis. It will be OK to what cannot fit nicely in the symbol body to be invisible. Pin placement (top, bottom, side, etc.) is also case-by-case. Update the symbol what looks best to you and then let us know why for the review. Thanks!

dejfson:
Hi, before I do go deeper into component definition, what about this one:
ada4870-1

dejfson:
to be said, the picture above is not finished, i just want to first get green light to a form+text around the pads before I finish the symbol

evanshultz:
It's looking much more nice now! Thank you.

I'd rather have straight pins than angled ones. What about moving PAD to the y-axis and putting TFL on bottom, which leaves room for /ON and /SD on top? That should work nicely.

Also, squeezing the pin text offset to 10mil should clean things up a bit too.

Lastly, did you contact ADI about the possible/recommended PAD connections?

evanshultz:
And so that it is included in the PR, the datasheet is at http://www.analog.com/media/en/technical-documentation/data-sheets/ADA4870.pdf.

dejfson:
You mean something like this?
hf3-51-2

dejfson:
ada4870-2

this version represents the last patch

evanshultz:
@dejfson This is looking good. Thanks! Please lower the pin text offset and bury the NC pins inside the triangle so they not likely to be accidentally connected. Then I'll review.

dejfson:
screenshot_20170913_211717

I have put NC pins inside body, but i'm not sure whether i understand what you mean by 'lowering pin text offset' would you elaborate on this? thaaanks

evanshultz:
Change "20" to "10":
image

dejfson:
hmm. fancy feature :) done. thanks

evanshultz:
Thanks! Final few things:
- "PAD" may fit inside the body, but otherwise remove the text by the pins.
- Remove the "Z" from the end of the part name. "ARR" describes the package and that's all we do.
- /ON and /SD pins should be of Input type.
- TFL should be of Output type.

dejfson:
Hi @evanshultz,

thanks for comments, repaired the three bottom comments, but i do not understand your first comment about removing the text by the pins. Could you elaborate what you mean? If i remove the 'PAD', 'TFL', '\SD' and '\ON' texts, nobody will know what these pins are assigned to. Or is it something else?

screenshot_20170914_082941

evanshultz:
The user can refer to the datasheet for pins. The generic triangle symbol doesn't leave room for pin text for all pins in all cases, like this one.

Also, instead of using a line between the pin and the triangle, just lengthen the pin and remove the line.

dejfson:
@evanshultz ,

I'm sorry, but I cannot agree at all with your statement, that user can look in the datasheet. Actually, I'm working in electronics industry for more than 25 years and I can tell you, that there is nothing worse, than having a printed version of the schematic in front of you, and 'trying to guess' what this and that 'undocumented' pin does. When you have the schematic opened in the computer of course this might be an argument, but for printed versions this is from my experience _very far from reality_ that one has a datasheet to a hand. I do not think that leaving the pin names in the form I did makes so disastrous effects on the quality of the library component I have made, and thus I'd like to ask you to accept my argument above. Hence here's the deal: i have repaired the pin lengths, but left the pin names there. If you can accept the component as is, please merge it (we've both lost already 3 days for this), if not, just close the pr because i'm not intending to spend any minute more on this.

thanks
david

Misca1234:
Is it possible to swap place of the number and the texts, perhaps it is more readable  ?
Or add the text as separate text boxes just "behind the number" ?

evanshultz:
@jkriege2 @SchrodingersGat 
I saw nameless pins all over the place in linear.lib, and I don't find referring to the datasheet to be objectionable. This PR is from an opposing perspective. Can you weigh in so we can decide how to handle this case?

SchrodingersGat:
I do not think that we should be allowing nameless pins. It's a hard problem:

1. Triangular symbols don't leave much room for labels
1. You cannot position the labels per-pin
1. You cannot show/hide labels per pin

I'm not sure how to provide a solution that isn't ugly - but I think that having blank symbols isn't the right way to do it...

jkriege2:
For rectangular symbols I would require names ... that's clear I think anyways. 
For OPAMPs it's often clear what is what, so maybe we can anme the pins and hide the names. Then we could add text-labels where necessary (+/-, poweris clear if they are the only un-named pins, then add user text labels at good position (as shown above) ...). This would also leave some more freedom in symbol design.

If we do that, I would add a small lib of examples to the KLC and require people to stick to them as far as possible (i.e. have a few OPAMPs in different configs, an instrumentation Amp ...).

Just a random example:
2017-09-26 07_32_31-part library editor -- d__kicad_kicad-library-review_library_linear lib
instead of:
2017-09-26 07_29_11-part library editor -- d__kicad_kicad-library-review_library_linear lib

PS: What will the new .sweet-fileformat do for us here?

evanshultz:
@SchrodingersGat 
There is one other option: Change pin text size on a per-pin basis which can be done today. Not that I think it's a great solution, but it's another knob to twiddle.

Having random text instead of pins names seems harder to maintain and, IMO, is ugly. In most cases, the pin name is either blank or contains the same text that would also be free-floating on the symbol. I don't like any result that is possible currently, but leaving off the few pins where names don't fit seem the least offensive to me.

Yes, I realize pin names are nice but is having a browser tab open when using KiCad such a burden (until there is a better way)?

jkriege2:
Hmmm I played around with the text-size per pin and text besides pin:
2017-09-28 08_18_35-greenshot

What do you all think? This might be an option ... also small text for some pins and text inside might:
2017-09-28 08_23_58-greenshot

jkriege2:
@SchrodingersGat What do you think?

SchrodingersGat:
I think the first option looks neater... Both are adequate though I think. Both will necessitate a change to the KLC to allow smaller pin names but that's fine.

dejfson:
Gentlemen,
once it is decided on what form the component should have, I will update it - providing that pin names stay. To support my argument, here is an example of pin-names-less hell which I'd like to very much avoid.
img_20170926_092935

(scanned from one un-named electronics design journal)

Please let me know your decisions.

jkriege2:
@SchrodingersGat Do you know what will be possible for pins in the new SWEET-format?

SchrodingersGat:
Download this - https://lists.launchpad.net/kicad-developers/binDp0KdUNMWc.bin and rename to a .odt

That's the "latest" spec doc for the new format

antoniovazquezblanco:
This is out of topic but I always wandered why isn't JSON used as the format for sweet files. That would reduce the code lines to maintain by using existing serializers/deserializers that can instantiate objects from readable text files.

SchrodingersGat:
Yeah JSON would be a very good idea!

evanshultz:
@SchrodingersGat @jkriege2 So is either one a solution here? I personally like the 2nd option because it retains a different color for pin text which seems more readable. And the pin text seems more closely related to each pin instead of being squeezed between pins.

JSON has a lot of overhead compared with the current file format, especially when the amount of information carried is small. Not saying it's good or bad, just that it is what it is.

SchrodingersGat:
Second option is good for me

SchrodingersGat:

PR #1613, opened 2017-09-11
Sfh617 a

Misca1234:
Push of optocoupler SFH617
http://www.vishay.com/docs/83740/sfh617a.pdf

SFH617A-1   DIP-4
Alias: SFH617A-2 SFH617A-3 SFH617A-4 SFH617A-1X001 SFH617A-2X001 SFH617A-3X001 SFH617A-4X001

SFH617A-1X006   10.16mm Wide DIP-4
Alias: SFH617A-2X006 SFH617A-3X006 SFH617A-4X006 SFH617A-1X016 SFH617A-2X016 SFH617A-3X016 SFH617A-4X016

SFH617A-1X007T 8mm clearance SMD DIP-4
Alias: SFH617A-3X007T SFH617A-2X017T SFH617A-3X017T

SFH617A-2X009T  SMD DIP-4
Alias: SFH617A-2X019T

bild

bild

bild

bild


Misca1234:
Not ready, seems that a different push was included

Misca1234:
Cypres push have been removed

jkriege2:
Hi!

thanks for your contrib. I only have one request: Could you make the style of your symbols look like the ones we already have in the lib? e.g. TLP291 or TLP785F? Also I think you can simply add your symbols as ALIASes:
- `SFH617A-1X006` and variants as ALIAS to `TLP785F`
- `SFH617A-1X007T` I think there are no matching symbols in the lib yet (with that package)
- `SFH617A-2X009T` and variants as ALIAS e.g. to `TLP291` ... also note: There is a mismatch between FPFilter and footprint here!
- `SFH617A-1`and variants as ALIAS e.g. to `TLP785`

Could you fix these?
Thanks,
JAN

Misca1234:
> Thanks for your contrib. I only have one request: Could you make the style of your 
> symbols look like the ones we already have in the lib? e.g. TLP291 or TLP785F? 

But, is it not a common remark that the diodes should not be filled but instead be transparent with a line through ?

The alias approach is a little strange, because SFH617 and TLP785 do not match their data (CTR for example) nor does their approvals. 

> SFH617A-2X009T and variants as ALIAS e.g. to TLP291 ... also note: There is a mismatch between FPFilter and footprint here!

I fix that, but it have an appointed FP so the filter wont be used

> SFH617A-1X007T I think there are no matching symbols in the lib yet (with that package)

It should be this one 
https://github.com/KiCad/Housings_DIP.pretty/pull/11
Approved by a guy with the name Jan ;)

evanshultz:
@Misca1234 I'll let Jan continue with the review, but I'll chime in with a few answers so you can continue working if you like.

This is the IEC diode standard to which KiCad adheres (where possible).

The ALIAS method affects only the symbol's visual representation. There are separate entries in the DCM for each ALIAS, which is where you can capture CTR and other differences.

If the footprints don't match, then they cannot be ALIASes. Good catch, if that's the case. You can still copy the symbol and then change only the footprints section.

jkriege2:
about filled diodes: Yes you're right ... simply edit the existing symbol and your PR in addition fixes this issue ;-)

About the ALIASes: Having an ALIAS only means that the same symbol (including footprint and FPFilter) is used. All the data is (in the scope of this lib) summarized in the DCM files. they contain the data entered in this dialog:
klc_props
This means that you can basically use one symbol and add everything as ALIAS, which has the same symbol + footprint + FPFilters + pinout. Using ALIASes has the advantage, that the library is way easier to manage (e.g. fixing at the LED-symbol): You need to fix fewer symbols, when a bug is found.

The FPFilter should always be set, because sometimes there are variants of a package (e.g. `_HandSoldering`. Then the footprint is set by default, but if a variant shall be used, you have a working filter that only leaves the matching symbols for selection.

About the package: ... Oh that Jan guy, I think he should write down what he means in a more understandable way ;-) ... I meant that this cannot be implemented as an ALIAS, as I didn't find any pre-existing symbol that matches all required criteria ;-9

Best,
JAN

Misca1234:
I know, i just try to protect my creation, the following have changed

SFH617A-2 SFH617A-3 SFH617A-4 SFH617A-1X001 SFH617A-2X001 SFH617A-3X001 SFH617A-4X001
alias to TLP785

SFH617A-1X006 SFH617A-2X006 SFH617A-3X006 SFH617A-4X006 SFH617A-1X016 SFH617A-2X016 SFH617A-3X016 SFH617A-4X016
alias to TLP785F

SFH617A-2X009T SFH617A-2X019T
alias to TLP291

The following is keept
SFH617A-1X006 10.16mm Wide DIP-4
Alias: SFH617A-2X006 SFH617A-3X006 SFH617A-4X006 SFH617A-1X016 SFH617A-2X016 SFH617A-3X016 SFH617A-4X016

jkriege2:
OK.
thanks for your contrib,
JAN

PR #1615, opened 2017-09-12
Fix SN65HVD230/231 Rs pin type

Rantanen:
Was running some ERC checks through a design and it kept flagging the Rs pin on the SN65HVD230 which I had pulled down to ground as per the datasheet. The pin was marked as Power Input in the library. As far as I could tell, it should be a normal input.

I was planning on just opening an issue for discussing the change, but as the change is such minor, I figured it'd be easier to do a PR even if I've gotten something wrong and it ends up rejected.

| Name | NO. | Type | Description|
| ------- | ---- | ------ | ------------------ |
| Rs | 8 | I | SN65HVD230 and SN65HVD231: Mode select pin: strong pull down to GND = high speed mode, strong pull up to VCC = low power mode, 10kΩ to 100kΩ pull down to GND = slope control mode |

> http://www.ti.com/lit/ds/slos346n/slos346n.pdf (Page 5)

(On another note... While I was trying to make the change, my version of nightly library editor wanted to do a lot of changes in the lib file. Mainly re-ordering the various T/C/P commands in different order (C/T/P maybe?). Is this a known change at some point and is there a plan to do a sweeping fix for it? It makes making small isolated changes difficult - especially if that happens in larger libs.)


SchrodingersGat:
Thanks for this fix :)

Regarding library ordering, this is an issue caused by people editing the libraries using the stable and nightly versions of KiCad. The nightly version has a different ordering for the libraries. Agreed that it does make things difficult!

PR #1617, opened 2017-09-13
Schematic symbol for Texas Instruments BQ25504 Energy Harvester

MatthewStephenson:
My first attempt at a contribution, hoping I've got everything right, and suggestions very welcome.

This is a symbol for the TI BQ25504 energy harvester, which takes a low-voltage very low-current input from a small solar cell or thermoelectric generator, and boosts it up to a usable voltage to charge a li-ion cell or capacitor. Useful for things like battery powered sensor nodes.

Datasheet is here: [http://www.ti.com/lit/ds/symlink/bq25504.pdf](http://www.ti.com/lit/ds/symlink/bq25504.pdf)

I've tried to follow the KLC guidance as closely as I can but I'm sure there will be something I've missed.

image


SchrodingersGat:

PR #1619, opened 2017-09-14
Add OPA330(AI) in library/texas.lib

Heiko-L:
- OPA330AID SOIC-8
- OPA330AIDBV SOT-23-5
- OPA330AIDCK R-PDSO-G5 (package nonexistent)
- OPA330AIYFF DSBGA-5 (package nonexistent)

grafik

Datasheet: http://www.ti.com/lit/ds/symlink/opa330.pdf

- Deleted OPA330AIDBV in texas.lib and added as alias for MCP6L91T-EOT in linear.lib

- Deleted OPA330AID in texas.lib and added as alias for OPA333xxD in linear.lib

- Deleted OPA330AIDCK in texas.lib and added as alias for LMV321 in linear.lib

- Moved OPA330AIYFF from texas.lib to linear.lib

## Some corrections for oddities found when adding OPA330AI:

- Deleted LMV331 and added as alias for LMV321 (same pinout, same package)

- Deleted OP179GS and added as alias for OPA333xxD (same pinout, same package)

- Deleted alias MCP601-xSN from MCP601-xP (wrong package) and added as alias for OPA333xxD (same pinout, same package)

- Renamed alias LM6171 to LM6171xxN (package!)


jkriege2:
Hi!

thanks for you contribution. I'll have a closer look this evening. In the meantime: PLease move this symbol to linear.lib, where we have most of the OPAMPS. Maybe there is even a symbol there already that can be used as a base for this one, i.e. if possible implement this OPAMP as ALIAS to an existing symbol (if it's not possible due to pinout, package, ... just leave it as a separate symbol.

Best,
JAN

jkriege2:
Hmmm ... I had a look at the current state of linear.lib ... and with respect to packages it is a mess ... The symbols you chose for base for the SC-70 and SOT-23 do not only have the required package, but link both in their FPFilters ... so I would suggest to use the separate symbols you initially had in the texas.lib .... sorry for all the mess ...

THanks,
JAN

Heiko-L:
That catched my eye also... Hope you will like my next one 😀

evanshultz:
@Heiko-L @jkriege2 
Since I updated linear.lib recently, let me pop in here. Using the symbols you made is fine. I'll be sorting out symbols with mutually exclusive packages in the near future.

If you make other changes please always mention them in the PR. This will make it easier to understand what you intend to change and review your contribution. After you moved the parts to linear.lib, you made a number of other changes without any description of it. I see:
- Removed LMV331
- Removed OP179GS
- Moved MCP601-xSN
- Renamed LM6171

Unless I've missed something, these changes are unwanted removals of parts or do not fix anything because of the logical groupings instead of physical groupings of symbols. Why did you do this?

Also, always include a datasheet (or other related documentation) in addition to an image of the symbol/footprint. In this case, the datasheet is at http://www.ti.com/lit/ds/symlink/opa330.pdf.

Lastly, would you be willing to create the missing footprints?

Heiko-L:
> If you make other changes please always mention them in the PR. This will make it easier to understand what you intend to change and review your contribution.

I tried so hard to write good commit messages and then failed updating the PR description ;) I changed that now.

> Why did you do this?

I hoped to reduce redundancy.

> Lastly, would you be willing to create the missing footprints?

I can try, but AFAIU the size of DSBGA-5 is not really defined, so this will be a problem.

jkriege2:
About the DSBGA-5 I didn't find a simple sketch of the proposed footprint, but there is an application note http://www.ti.com/lit/an/snva009ag/snva009ag.pdf which describes this series of packages in general. That AN says that with pitch 0.4mm (I think that's the correct one) and solder ball diameter 0.240-0.265mm, the pad should follow these measures:
Copper Pad: 0.225 ± 0.02 mm
Solder Mask Opening: 0.325 ± 0.02 mm

Does that help? I think @evanshultz and @SchrodingersGat might be able to help a bit more with these packages ... I will try and review your new changes this evening or during the weekend.

best,
JAN

Heiko-L:
Land patterns are there but no "outlines", IMHO because of different wafer sawing(?).

By the way:
How to deal with this kind of sorting KiCad does with libraries? 
Is it enough to open it before committing or do I have to save a part to let KiCad "sort" the library to avoid this nasty line fiddling?

evanshultz:
For the DSBGA parts, the general drawing doesn't include the dimensions but the part datasheet does. See page 47 of the datasheet:
image

My guess is that TI creates the general drawing but they use different packaging houses for different parts which can vary a bit in size (but all stick to the generic TI drawing). Maybe that's not correct, but it will work. If you use the average of the X and Y dimensions in the datasheet, we'll have an accurate footprint for OPA330's DSBGA-5 package and if there are other variations on DSBGA-5 they can be created when needed.

Thanks!

jkriege2:
Or use the max, then any package of THAT type should fit ... the copper pattern is the same anyways ...

jkriege2:
PS: Maybe we should limit this PR to adding OPA330 here only and leave out all other reworks of the lib (resorting as ALIAS ...) ... I think the new sorting from the nightly could be kept and merged here ... in the end we will have to do that at one point anyways ... the other changes I would leave to a more generic cleanup-PR (e.g. the one @evanshultz proposed ...). OK with you?

Heiko-L:
Yes, I will revert the other changes.

Heiko-L:
I have revert the commits. Would it be better to close this one and file a complete new PR for the OPA330 that don't touches texas.lib?

jkriege2:
I think that would be cleanest ...

Thanks,
JAN

PR #1620, opened 2017-09-14
Use generic parts in linear library

Heiko-L:
How about using some ;-) generic symbols and alias against them?

grafik

A generic symbol is needed for every package and pin configuration but with this approach a new real existing part can be an alias for a generic part. So -for example- an amp from TI is no alias for a complete different amp from Microchip anymore.

Creating a new part would involve to look through the generic parts and set an alias if one fits or create a new generic part and set an alias for it.

### Naming scheme for generic parts
Amp.[S/D/Q].[package].[pin assignment from low to high numbers]

### pin assignment abbreviations:
Nc: not connected
Ii: inverting input
Ni: non-inverting input
Pn: negative power supply
Pp: positive power supply
O: output


evanshultz:
https://github.com/KiCad/kicad-library/pull/1114. You can close this PR and contribute your comments to the other one if you like.

Heiko-L:
Should I cherry pick the commits to the other branch?
Or better cherry pick to this branch?
Or none of that- then tell me, I'm new to this contributing stuff :)

Heiko-L:
I'm closing this one because one discussion is enough -> #1114

Heiko-L:
Oh, wait, I reopen because #1114 is for generic operational amplifiers in _device.lib_, this is not what I intended.

library/linear.dcm
```diff
@@ -732,8 +768,14 @@ K quad opamp
 F http://www.ti.com/lit/ds/symlink/lmc6484.pdf
 $ENDCMP
 #
-$CMP LMV321
```
Heiko-L: Split to LMV321M5 and LMV321M7 to distinguish between packages

library/linear.dcm
```diff
@@ -222,6 +222,42 @@ K single comparator with reference
 F http://www.analog.com/media/en/technical-documentation/data-sheets/ADCMP350_354_356.pdf
 $ENDCMP
 #
```
Heiko-L: here come the generic parts

library/linear.dcm
```diff
@@ -744,8 +786,14 @@ K quad opamp
 F http://www.ti.com/lit/ds/symlink/lmv324.pdf
 $ENDCMP
 #
-$CMP LMV331
```
Heiko-L: Split to LMV331M5 and LMV331M7 to distinguish between packages

library/linear.lib
```diff
@@ -461,6 +461,141 @@ X V+ 4 0 300 150 D 50 50 1 1 W
 ENDDRAW
 ENDDEF
 #
```
Heiko-L: Here come the new generic parts with their aliases

library/linear.lib
```diff
@@ -1564,59 +1699,16 @@ $FPLIST
...
 X TR B3 -500 200 150 R 50 50 1 1 I
 X Q C3 500 200 150 L 50 50 1 1 O
-X R C2 -500 -200 150 R 50 50 1 1 I I
-X CV C1 -500 0 150 R 50 50 1 1 I
-X THR B1 500 -200 150 L 50 50 1 1 I
-X DIS A1 500 0 150 L 50 50 1 1 I
-ENDDRAW
-ENDDEF
-#
-# LMV321
```
Heiko-L: LMV321 deleted.
LMV321M7 is now an alias of Amp.S.SC-70-5.Ni-Pn-Ii-O-Pp
LMV321M5 is now an alias of Amp.S.SOT-23-5.Ni-Pn-Ii-O-Pp

library/linear.lib
```diff
@@ -1564,59 +1699,16 @@ $FPLIST
...
-#
-DEF LMV321 U 0 20 Y Y 1 F N
-F0 "U" 0 200 50 H V L CNN
-F1 "LMV321" 0 -200 50 H V L CNN
-F2 "" 0 0 50 H I L CNN
-F3 "" 0 0 50 H I C CNN
-ALIAS MCP6561U MCP6001U OPA333xxDCK OPA376xxDCK
```
Heiko-L: OPA333xxDCK -> Amp.S.SC-70-5.Ni-Pn-Ii-O-Pp
OPA376xxDCK -> Amp.S.SC-70-5.Ni-Pn-Ii-O-Pp
MCP6001U -> Amp.S.SOT-23-5.Ni-Pn-Ii-O-Pp
MCP6561U -> Amp.S.SOT-23-5.Ni-Pn-Ii-O-Pp

library/linear.lib
```diff
@@ -1564,59 +1699,16 @@ $FPLIST
...
-P 4 0 1 10 -200 200 200 0 -200 -200 -200 200 f
-X V- 2 -100 -300 150 U 50 50 0 1 W
-X V+ 5 -100 300 150 D 50 50 0 1 W
-X + 1 -300 100 100 R 50 50 1 1 I
-X - 3 -300 -100 100 R 50 50 1 1 I
-X ~ 4 300 0 100 L 50 50 1 1 O
-ENDDRAW
-ENDDEF
-#
-# LMV331
```
Heiko-L: LMV331 deleted.
LMV331M7 is now an alias of Amp.S.SC-70-5.Ni-Pn-Ii-O-Pp
LMV331M5 is now an alias of Amp.S.SOT-23-5.Ni-Pn-Ii-O-Pp

library/linear.lib
```diff
@@ -2514,49 +2606,6 @@ X ~ 8 100 -300 250 U 50 50 1 1 I
 ENDDRAW
 ENDDEF
 #
-# OP179GRT
```
Heiko-L: OP179GRT deleted.
OP179GRT now is an alias of Amp.S.SOT-23-5.O-Pp-Ni-Ii-Pn

library/linear.lib
```diff
@@ -2514,49 +2606,6 @@ X ~ 8 100 -300 250 U 50 50 1 1 I
...
-P 4 0 1 10 -200 200 200 0 -200 -200 -200 200 f
-X V+ 2 -100 300 150 D 50 50 0 1 W
-X V- 5 -100 -300 150 U 50 50 0 1 W
-X ~ 1 300 0 100 L 50 50 1 1 O
-X + 3 -300 100 100 R 50 50 1 1 I
-X - 4 -300 -100 100 R 50 50 1 1 I
-ENDDRAW
-ENDDEF
-#
-# OP179GS
```
Heiko-L: OP179GS deleted.
OP179GS now is an alias of Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc

library/linear.lib
```diff
@@ -2735,32 +2784,6 @@ X NC 16 0 -100 100 U 50 50 2 1 N N
 ENDDRAW
 ENDDEF
 #
-# OPA333xxD
```
Heiko-L: OPA333xxD deleted.
OPA333xxD -> Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc

library/linear.lib
```diff
@@ -2735,32 +2784,6 @@ X NC 16 0 -100 100 U 50 50 2 1 N N
...
-#
-DEF OPA333xxD U 0 10 Y Y 1 F N
-F0 "U" 0 250 50 H V L CNN
-F1 "OPA333xxD" 0 150 50 H V L CNN
-F2 "" 0 0 50 H I L CNN
-F3 "" 150 150 50 H I C CNN
-ALIAS OPA340P OPA340UA OPA376xxD MCP6L91RT-EMS MCP6L91RT-ESN TSV911IDT THS4631D
```
Heiko-L: OPA340P -> Amp.S.DIP-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc
OPA340UA -> Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc
OPA376xxD -> Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc
MCP6L91RT-EMS -> Amp.S.MSOP-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc
MCP6L91RT-ESN -> Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc
TSV911IDT -> Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc
THS4631D -> Amp.S.SOIC-8.Nc-Ii-Ni-Pn-Nc-O-Pp-Nc

evanshultz:
See from the linked PR that generic opamps have wide support, but my PR wasn't merged. Yet?

Furthermore, I do not think this is the approach that is planned for the KiCad library. Using periods to separate the symbol name violates KLC, for one. @SchrodingersGat @jkriege2 

In addition, I'd prefer to split up the opamps since I have records on what I've done when I modified linear.lib. If that's OK with you once a direction has been agreed upon.

Heiko-L:
> See from the linked PR that generic opamps have wide support, but my PR wasn't merged. Yet?

You can add a "datasheet" for the generic parts, so the automatic check will succeed.
Look at mine: http://www.thefreedictionary.com/generic ;)

> Furthermore, I do not think this is the approach that is planned for the KiCad library.

Having a real part like the OPA330 and other parts as alias for this one is very ugly especially when the real part is a "multi footprint" setup. In #1619 @jkriege2 called it a mess :)

> Using periods to separate the symbol name violates KLC, for one

It is explicitly allowed- rule 1.7.IV

What do you mean by saying "split up the opamps"?

jkriege2:
Hi!

I ran into the same problem when I reworked regul.lib ... There I went for an approach to tak a real part as basis with either a single footprint, or a set of FPFilters. Then when adding a new part/looking at the next I tried to find an existing part that exactly matches (either the available range of footprints or the specific one) and put the new part as ALIAS to the one I found. If there was none that exactly matches, I created a new, which could be used for others as basis. 

The problem with the generic OPAMPs is that we would have a large set of these in different variants (different pin-names, available pins, pinouts, packages, ...) and they cannot be hidden. For the dual and quad OPAMPS it's a bit different: There the pinouts are more or less standardized (with few exceptions that can be handled separately) ... but I'm not sure whether these generics can be sued as basis for actual OPAMPs due to the footprints: Either the generics simply list all possible footprints as FPFILTERS and the user has to select the correct one, although he already selected a special type (ALIAS) which is in fact only available in a small number of packages (hopefully a subset of the FPFILTERs), or we have variants for each specific footprint (giving a large number of generics).

I think the generics should serve as basis for non-axistent parts and not really list any footprints/FPFILTERs (i.e. should not be the basis of an ALIAS), much like the diode/transistor generics in device.lib (yes there we have all possible variants, but they are way fewer!) ...

My two cents on that!

JAN

@SchrodingersGat @evanshultz @pointhi @poeschlr What do you think?

poeschlr:
Disclaimer: Because of a lack of time lately i am a bit out of the loop. I'm also not really sure what @Heiko-L wants to achieve with this PR.

---

Now my 10c for generic parts in the linear lib:
I don't like generic symbols in random libs. If we add generic parts they should be in a lib where the user knows that these are generic (currently device lib). All other libs should hold atomic or semi atomic parts. (Semi atomic parts are parts that are produced by many manufactures sharing a base MPN but have differing option codes. There i think we could use Base_MPN + footprint_name as the identifier.)

What would i expect all symbols to look like in the linear lib?
 - one symbol per footprint, footprint field assigned to all symbols. 
 - Named via the full manufacturer part number or via the base mpn for semi atomic parts.

If different symbols share the same pin names/ pin numbers and the footprint they can of course be aliases of each other. For that there does not need to be a generic symbol in the lib. (Similarly to what is done in the regul lib.)

Heiko-L:
@jkriege2 

> Then when adding a new part/looking at the next I tried to find an existing part that exactly matches (either the available range of footprints or the specific one) and put the new part as ALIAS to the one I found. If there was none that exactly matches, I created a new, which could be used for others as basis.

That would be much faster if you only had to look at the generics without cycling through the whole library.
I like the idea of reducing redundancy with aliases but for me it feels a bit strange when OPA4134 is an alias for LM2902 ;)
I could live with both of them being an alias for a generic though.

> The problem with the generic OPAMPs is that we would have a large set of these in different variants (different pin-names, available pins, pinouts, packages, ...) and they cannot be hidden. 

Yes, that is really a contra :(
But maybe with a "Z" in front they would be sorted to the end of the list.

> Either the generics simply list all possible footprints as FPFILTERS and the user has to select the correct one, although he already selected a special type (ALIAS) which is in fact only available in a small number of packages (hopefully a subset of the FPFILTERs), or we have variants for each specific footprint (giving a large number of generics).

For this it would be very helpful if KiCad would shift the footprint and footprint filter fields to the .dcm files.

@poeschlr 

> I'm also not really sure what @Heiko-L wants to achieve with this PR

When I made OPA330 I intuitively made a **copy** of an existing symbol and named it.
Then @jkriege2 told me to check if a symbol with the same pinout exists and then add OPA330 as alias. So I had to cycle through all symbols (ok, filtered by package) which was not very pleasant and I found further **copies**. So there is a lot of redundancy in the libraries.

BTW: Atomic parts add the most redundancy. That must not be if KiCad gives us the chance to use aliases.

If there was a bunch of generic symbols this search would be much easier, because if there is no fitting symbol in the generics one knows at once that a new one must be drawn- without cycling through the library, just with "one look".

IMHO maintenance would be easier too.

Heiko-L:
Something about the redundancy:
I found OP179GS as an discrete symbol, but it could be an alias of OPA333xxD.
The same for LMV331 which could be an alias of LMV321.

Now imagine that all these discrete symbols would have different drawings and you place many different of them on a schematic, that would look really inconsistent.
To fix this you would have to crawl through the whole library and change a lot of drawings.

Sure, with generic "parent" symbols you also have a bunch of drawings to check, but they gather at one place in the library no need to crawl through the whole library

jkriege2:
Hi @Heiko-L 

two comments:
1. Having the footprint/FPFilter sin the DCM would be really great, but unfortunately we don't have this option right now. Also there is a new library fileformat under development, which will (in my understanding) solve many of these problems ... so I think we just need to be patient here ...
2. The parts you proposed above should be ALIASes and reworking the style of symbols  is both possible in either case (generic symbols or non-generic symbols as ALIAS-base.)

Best,
JAN

Heiko-L:
> Also there is a new library fileformat under development, which will (in my understanding) solve many of these problems ... so I think we just need to be patient here ...

That would be great. Are you, the library maintainers, getting involved?

> The parts you proposed above should be ALIASes and reworking the style of symbols is both possible in either case (generic symbols or non-generic symbols as ALIAS-base.)

For sure. I'll open another "cleaning" PR for that if that is okay.

jkriege2:
@SchrodingersGat What do you think on this?

Heiko-L:
And take a look at #1623 posts to see how maintenance can be positively affected.

jkriege2:
After the lengthy discussion there: https://github.com/KiCad/kicad-library/pull/1114 I think we can close this. 

Best,
JAN

PS: Please reopen, if I'm mistaken!

PR #1621, opened 2017-09-15
Add schematic symbol for AD-121F2 RGB 7-segment display.

ppelleti:
[Datasheet](http://usasyck.com/products/AD-121F2_cat_e.pdf).

One thing that the datasheet doesn't make obvious is that the no-connect pins (7, 8, and 12) are not physically present.  For this reason, I am leaving those pins out of the schematic symbol and footprint, rather than modeling them as no-connect pins.

This symbol is meant to be used with [this footprint](https://github.com/KiCad/Displays_7-Segment.pretty/pull/13).

ad-121f2

ad-121f2-symbol


SchrodingersGat:
Thanks!

PR #1623, opened 2017-09-16
Add opa330 second try

Heiko-L:
Added OPA330xxDBV as alias of MCP6L91T-EOT
Added OPA330xxD as alias of OPA333xxD
Added OPA330xxDCK as alias of LMV321

Datasheet: http://www.ti.com/lit/ds/symlink/opa330.pdf

Added new symbol for OPA330xxYFF
bildschirmfoto von 2017-09-16 11-06-50


jkriege2:
Hi!

OK, thanks for the cleaned PR. Now let's clean up the package/ALIAS mess:
1. OPA333xxD should be a separate symbol (not an ALIAS), because the ALIAS-base links also to DIP and MSOP-packages, in which it is not available. + the ...D-option is SOIC-only. So this should be a separate symbol with FPFilter for SOIC-8 and footprint field set to the appropriate footprint 
2. OPA330xxDBV and  should be a separate symbol, for the same reasons as above ... here it is even more important, as the SC-70 adn SOT-23-packages are different AND have different pinouts.
3. OPA330xxYFF looks fine ... 

Best,
JAN

Heiko-L:
> OPA333xxD should be a separate symbol (not an ALIAS), because the ALIAS-base links also to DIP and MSOP-packages, in which it is not available. + the ...D-option is SOIC-only. So this should be a separate symbol with FPFilter for SOIC-8 and footprint field set to the appropriate footprint

OPA330xxD is SOIC only just like his alias-base OPA333xxD. OPA333xxD should only have SOIC footprint, so I recommend tidying up instead of adding redundancy.

> OPA330xxDBV and should be a separate symbol, for the same reasons as above ... here it is even more important, as the SC-70 adn SOT-23-packages are different AND have different pinouts.

The same here- the alias-base MCP6L91T-EOT is SOT only, so the alias-base is incorrect and should be corrected.

That is exactly what I meant in 1620... If the alias-base would be named Amp.S.SOT-23-5.Ni-Pn-Ii-O-Pp everyone could see with just one look that this is SOT-23-5 with given pin assignment.
If there is just MCP6L91T-EOT one has to open the datasheet and search the table where EOT is explained... That sucks.

jkriege2:
I would rather not merge this PR, if it leaves the lib in an unclean (and incorrect) state. Currently I see three options (the generic parts do not seem to find too many advocats): 
1. Either we wait until the lib is cleaned up and then merge this symbol into that
2. You clean up "around" this symbol here, so we have a partial makeover that already helps a great deal (but that's a lot of work)
3. We merge these symbols as separate symbols, so the OPA330 is correct and clean ... then they can be merged into other symbols in a separate cleanup-PR later.

I would prefer 3.

best,
JAN

Heiko-L:
I try option 2...

jkriege2:
Hi!

OK, I'm looking into this right now ... now we lost several devices (with their packages):
- MCP6001 is also available in SC-70, but only the SOT-23-5-version is in the lib now
- LM6171 is now only available in DIP, but there is also a SOIC-variant LM6171D
- OPA340 is available in SOIC as OPA340PD
- TS881 is also available in SC-70
- MCP6561 is also available in SC-70
- OPA376
- ...
Note: This list is for sure not complete! Please have a very careful look at all the devices you touch ...

Best,
JAN

Heiko-L:
I hope you don't say that there are too much changes for one PR in the end...

evanshultz:
@Heiko-L 
I updated linear.lib by grouping logically. Since then, the librarians have decided to group both logically and physically, so differing footprints should not share a symbol.

I simply needed a break of a couple weeks after finishing several library updates, and you were caught in that valley before I went back to linear.lib. Apologies for that.

However, I think Jan is saying above that this PR should cover adding OPA330 and nothing else. I believe this is the best way to go. Let's get your nice and shiny part in the library, and any other outstanding PRs, and then linear.lib can be quarantined during another big overhaul to get it to meet the current library design. Adding changes beyond OPA330 is only likely to drag this out and make the library less uniform than it is now. Unless Jan objects, please only add OPA330 to the master branch and then this PR can be merged.

Again, I apologize for the bad timing. Nobody intended for this to be frustrating and we all value your contributions. I'm trying to suggest the smoothest way to move forward from where we are today.

Heiko-L:
> Adding changes beyond OPA330 is only likely to drag this out and make the library less uniform than it is now

It is more uniform than before, that's the case. It is not my fault if there are devices which are not defined in all variants **before** I made my PR (OPA376), the "cleaner" has to add them anyway if that is so important.
I don't understand why there should be even more bloat someone has to tidy up at the end.

jkriege2:
@Heiko-L to explain my comments above a bit:
I tried to go through the changes made you your last two commits. In these you renamed symbols, such as the general LM6171, which had SOIC and DIP listed to the SOIC-only variant. This way we "lost" a valid symbol and that's what I listed above. As you decided to go for the partial workover above, I try to review this PR to make the workover + your symbols as good and clean as possible(since you touch both, both are important for this PR)!

Initially I proposed the "simply add your devices as separate symbols" strategy, just to prevent this mess (which I would have left for @evanshultz ... as he already stated a few weeks ago that he would do the workover).

So: Either way, I will try and review this PR as good as possible. I'll have a further look this evening (after work ;-)

Best,
JAN

Heiko-L:
> This way we "lost" a valid symbol and that's what I listed above.

Cough... Valid?... Cough ;)
I can, at a pinch, understand that, but OPA376? Come on...

jkriege2:
... ahhh yeah, that one was one too much (my fault, sorry) ... basically I looked at all the symbols you touched and checked all ALIASes of them, that I could find.

jkriege2:
OK, I'll merge this now ;-)

Thanks for your work an patience.

Best,
JAN

Heiko-L:
Thank you, sorry for being a bit short-tempered.

PR #1628, opened 2017-09-17
Added AD8230

chrigro:
ad8230


jkriege2:
Hi!

nice work!

please move this symbol to linear.lib and make a symbol that orients itself on the existing symbols for instrumentation amplifiers (triangle-shape) in that lib. Otherwise: looks good to me!

THanks,
JAN

chrigro:
Done. Thanks for the patience, I am learning ... ;-)

Christian

jkriege2:
I think setting "Pin Name Position Offset" to 10 might improve this symbol a bit:
2017-09-20 08_36_20-part library editor -- d__kicad_kicad-library-review_library_linear lib
... the pin-names move apart a bit that way ...

What do you think?
JAN

chrigro:
For sure. But also, this symbol is too large. If I make it smaller, I have to remove the names of pins 3, 6, 7. I personally would have voted for larger symbols + pin names, but the lib convention seems different.

Christian

jkriege2:
Hi @chrigro!

Have a look at other instrumentation amplifiers in the lib ... they are as large as your original symbol ... so the size is fine. I just wanted to move the Vr1 and Rg label a bit apart, so they don't collide ... otherwise your symbol was fine!

Best,
JAN

jkriege2:
@chrigro please fix the conflicts in this PR!

chrigro:
I changed the symbol back to the large size and fixed the label offset.

To fix the merge conflicts: I merged the up to date master into this branch taking all changes from remote + I kept the AD8230 related stuff. I hope that was ok.

Christian

jkriege2:
Thanks, all fine now!
Best,
JAN

PR #1632, opened 2017-09-19
Fixed pin numbering on the INA138/INA168 schematic symbol

danwood76:
Fixed pin numbering on the INA138/INA168 schematic symbol as per the TI datasheet:
http://www.ti.com/lit/ds/symlink/ina138.pdf


SchrodingersGat:
Thanks @danwood76 for your contribution :)

PR #1634, opened 2017-09-19
AP2127 LDO regulators

ppelleti:
[Datasheet](https://www.diodes.com/assets/Datasheets/AP2127.pdf).

This is failing the KLC script, but I don't think any of the other regulators have their power input pin on the top, either.

```
Checking symbol 'AP2127K-3.3':
     - Pin VIN (1) @ (-300,100)
```

ap2127k


jkriege2:
Hi!

looks good to me (i.e. I will merge in this state) ... would you be willing to also add the other variants from DS, i.e. the SOT-23-3 and the SOT-89 variants? the SOT-23-3 could simply be aliases to `APE8865N-12-HF-3` and the SOT89 could be ALIAS of `MCP1703A-3302_SOT89`

That would be great ... if you don't have time, just let me know ...

best,
JAN

ppelleti:
Sure, I can do that.  You also make a good point about the aliases.  I think the SOT-23-5 variants can just be aliases of `AP2204K-1.5`, so I didn't need to create a new symbol.

ppelleti:
There are actually four variants:

* SOT-23
* SOT-23-3
* SOT-23-5
* SOT-89

The SOT-23 and SOT-23-3 footprints shown in the datasheet are very similar, but subtly different.  (Pages 20 and 21.)  The KiCad footprint library seems to only have SOT-23 but not SOT-23-3.  Should I just skip the SOT-23-3 variants?

ppelleti:
OK, I think I have all the variants aliased correctly now.  (Except for the SOT-23-3 variants.)

jkriege2:
Hi!

I just checked the footprints and it seems a bit of a mess with all the variants:
- KiCAD SOT-23 matches SOT-23 from DS
- KiCAD TSOT-23 is a very good match for SOT-23-3 from DS
- KiCAD TSOT-23-5 seems to be the best match for SOT-23-5 from the DS (the pads of SOT-23-5 are a few 1/10mm too short, while the ones for TSO-23-5 are a bit too wide) ... on the other hand I poked around a bit and it seems that the SOT-23-5 is often a bit wider and often they propose a bit wider pads, so one could also change that and leave the TSOT for thinSOT ... 

But basically the pads for all these symbols are wide enough to fit all variants (I think)
2017-09-20 08_22_10-pcbnew d__kicad_kicad_review-project2_kicad_review-project2 kicad_pcb
Note: The width "Pads" is the distance from the lhs pad left outer edge to the rhs pad right outer edge.

Does that help?
JAN

PS: @evanshultz @SchrodingersGat Do you know something/can comment on the different types of "SOT-23" packages ...

SchrodingersGat:
If you are not confident that the footprints we have in the library do not match *precisely* (or close enough) then I would suggest adding a variant of the `SOT-23-3` with `DIODES_` prefix.

However I would think that they are going to be good enough for purpose.

ppelleti:
Sounds like the consensus is that we can use the same footprint for SOT-23 and SOT-23-3.  I've gone ahead and added aliases for the SOT-23-3 variants.

jkriege2:
I think the SOT-23 footprint we have is a bit too narrow for SOT-23-3 from the DS .. I would propose using TSOT-23, especially since the datasheet requires the distance of the outer pad edges to be 3.6mm ...

ppelleti:
Sure.  I created a new part for `AP2127N3-1.2` and `AP2127N3-1.5`, and gave it the `TSOT-23` footprint.

jkriege2:
Hi!

nice work! 

Best,
JAN

PR #1639, opened 2017-09-20
Several adjustable regulators had "fixed" as a keyword, or vice-versa.

ppelleti:
I just noticed that in `regul.dcm`, several adjustable regulators had "fixed" as a keyword, or vice-versa.  This PR corrects that.


jkriege2:
nice find!

THanks,
JAN

PR #1641, opened 2017-09-21
Add lm2937 to regul.lib

Heiko-L:
This PR adds LM2937 to regul.lib.
grafik
grafik
grafik

Datasheet: http://www.ti.com/lit/ds/symlink/lm2937.pdf


SchrodingersGat:
Great work, thanks :)

PR #1643, opened 2017-09-21
Add the Nidec Copal SH-70x0 series of rotary coded switches.

ppelleti:
[Datasheet](https://www.nidec-copal-electronics.com/e/catalog/switch/sh-7000.pdf).

This is failing KLC:

```
Checking symbol 'SW_Coded_SH-7010':
  Violating EC01 - Extra checking
    General pin number checking
    pin: C number C is not valid, should contain at least 1 number
```

However, the datasheet does not give any other pin numbering scheme.  How should I handle this?  Do I need to invent a pin numbering scheme for this part to make KLC happy?

I have footprints for these, too, but first I wanted to get the pin numbering issue resolved.

sh-7010-symbol

sh-7050-symbol


Misca1234:
> However, the datasheet does not give any other pin numbering scheme. How should I handle this? 
> Do I need to invent a pin numbering scheme for this part to make KLC happy?

Not necessary, if you state why the travis check fails the librarians will check it up and can still merge symbols that have travis errors, if it is motivated.

ppelleti:
Thanks!  If the pin numbering in the datasheet is acceptable, then I'm ready for this PR to be reviewed & merged.

I've also gone ahead and created PRs for the footprints.

jkriege2:
Hi!

since the matching footprints have already been merged ... We should also merge this. 
One comment: Please shorten the pins a bit (to 100 or 150mil) .... I think your's are 200mil at the moment.

Thanks,
JAN

jkriege2:
@SchrodingersGat Since you merge the footprints. There is a question about pin/pad names here ... @ppelleti choose the scheme from the datasheet. This is fine (IMHO) for these manufacturer-specific devices, but since such switches are available from several manufacturers with the same number/types of pins ... should we try and agree on a naming scheme for switches? I think we already discussed that somewhere ... Basically the use here cooks down to:
- C for common pin
- 1,2,4,8: pin-number = 2^bit in the encoded number
The existing swicthes are simply numbered:
- SW_Coded with 1=common
- SPDT: 2=common
- 2xSPDT: 2,5=common
- ...

Note: This numbering leads to an error in our scripts ...

Best,
JAN

poeschlr:
I think no matter how generic a device is, the manufacturer pin numbering scheme should always be used.
Otherwise the documentation (schematic) would disagree with the real device.

Yes this will likely mean that the number of symbols is higher than when we use one pin generic pin numbering and force all similar devices to use this numbering scheme no matter what the datasheet says.

ppelleti:
OK, I've shortened the pins to 100.

sh-7050-shorter-pins

jkriege2:
Thanks,
JAN

PR #1644, opened 2017-09-21
Add BQ24012 Single-Chip Li-Ion Charge Management IC

MatthewStephenson:
This is a symbol for the TI BQ24012 battery charger IC.

Datasheet here: http://www.ti.com/lit/ds/symlink/bq24012.pdf

image

Suggestions and advice welcome as always :)


Misca1234:
You should shorten the pins length to half the of the current length
Strife to have all inputs on the left side 
Strife to have all outputs on the right side
The EP pin can be invisible (would be nice if it is renamed to 1EP as well)

It is beneficial if you past the following image as well

bild

MatthewStephenson:
Hi, thanks for the feedback. I have shortened the pins and renamed EP to 1EP. I also tried to group the outputs more logically, with the three status outputs together, and the ISET separate (which is a weird one, as it's electrically an output, but its purpose is to change the set current so functionally it's an input...)

About the EP, the datasheet says:

"The exposed thermal pad must be connected to the same potential as the Vss pin on the printed circuit board. Do not use the thermal pad as the primary PAD ground input for the device."

So it's not NC, but should be tied to Vss. I tried moving it on top of the Vss pin so they would be connected, but then the checking script complained that they had different names. Should I therefore:

a) Leave it as it is, as a separate pin?
or
b) Rename it to VSS, hide it, and move it on top of VSS? (This seems like it would create confusion)

Here's the requested screenshot with the pin table.

image

Thanks,

- Matthew

Misca1234:
The stat1, stat2 and PG should be of the type "open collector"
Add the character '~' before C in CE for pin 8 (as you done for the PG one)

If you download a nightly build of kicad you will get this fancy look in the symbol editor (this is just an example)

bild

Misca1234:
I might be wrong here but should you not use foot print 
Housings_SON:VSON-10-1EP_3x3mm_Pitch0.5mm_ThermalPad
instead ?

MatthewStephenson:
Hmm, I think you're right. I've updated it.

image

The SON has the extra thermal pads on each end and better matches the datasheet image:

image

Thanks. I think it would have fitted and worked on the DFN (the pins are the right spacing and the body is the same size) but your suggestion is definitely better.

I was confused as to why the datasheet was calling it QFN, not DFN, when there were only "functional" pins along two edge, but now it makes sense.

Misca1234:
np, I am here for you

PR #1650, opened 2017-09-23
Tle2426

Misca1234:
TLE2426

This is a replacement of, seemingly abandoned push
https://github.com/KiCad/kicad-library/pull/1350

http://www.ti.com/lit/ds/symlink/tle2426.pdf

I have not added the wide TO-92 capsule, it is not clear which one it is.
I guess it is TLE2426CLPE3 but if we should add that as well I need to create a new footprint,
the current wide To-92 is not wide enough.

bild

bild

bild

bild

bild

bild


jkriege2:
Hi!

we usually don't add all temperature variants, but add an `x`as wildcard to the symbol name. For this symbol that means:
- remove the ID and MD variants and rename `TLE2426CD` to `TLE2426xD`
- same for the two other footprint variants!

Also: According to the DS, the TO-92 is available in two variants (narrow and wide stretched pins). Maybe it would be good to add a second FPFilter in this special case: `TO?92*Inline*Wide*`.

Also it seems that a pin name position offset of 20mil (instead of the default 40mil) will be suggested for new symbols in this lib ... for your symbols that will improve the look (I feel) ... so could you change that?
2017-09-23 18_30_22-part library editor -- d__kicad_kicad-library-review_library_references lib

Also I think the 3-pin variant looks better when the hor. pins move 100mil towards the top (the they look a bit more like a voltage regulator):
2017-09-23 18_33_45-part library editor -- d__kicad_kicad-library-review_library_references lib
Please change ... if you agree (I will merge it with the current look too ;-)

Maybew we should even change the appearance to a triangular symbol, as the datasheet suggests ... the NR-pin may then be at the bottom:
tle2426
What do you think on that? You could maybe even add the text 1/2 in the center of the triangle, as e.g. done here:
with12

Finally I found a typo in the gas `vitrual` ... should be `virtual` ;-)

Everything else is fine ... and I will merge, when the errors (Vitrual, temp-variants, ...) are fixed.
What do you say on may suggestions? Will implement one or both?

best,
JAN

Misca1234:
I will remove the aliases
There is no wide TO-92 that fits the data sheet, the data sheet wants a slighty wider foot print than the existing wide TO-92
Not sure why you see text offset 20, I have 40
I change the 3 pin variant to a triangular

Misca1234:
BTW
It is strange that kicad does not allow the temp variants, it will mess up the BOM list, when you generate the BOM list you can not see which variant it should be

Misca1234:
I did not change the LP variant to a triangle

Misca1234:
Wait with this one, 
my girlfriend have feed me with a lot of Laprhoipg whiskey the whole evening, 
it seems that i messed it up, I will close it and repush it

PR #631, opened 2016-07-28
Add ISSI IS31FL3216

GyrosGeier:
This is a 16 channel LED driver

PR #1652, opened 2017-09-23
Is31 fl3216

Misca1234:
This is a repush of IS31FL3216 of the, assumed abbonned, push
https://github.com/KiCad/kicad-library/pull/631

IS31FL3216 (alias IS31FL3216A)
http://www.issi.com/WW/pdf/IS31FL3216.pdf
http://www.issi.com/WW/pdf/IS31FL3216A.pdf

bild

bild


Misca1234:
Apparently kicad have "sorted and fixed" stuff on its own in the interface.lib. let me see if I can fix those

Misca1234:
The error message for IS31FL3216
"Pin CLK (1) @ (-600,200) is type BIDI : suggested INPUT"
is not correct, this is a pin that can be both input or output if you cascade several chips

I also suggest we remove  AD5282_ABC because of
No meta data 
No foot print or foot print filter
Pin missaligment
Missing pins
AD5282 already exist in the library


Shall I make a separate push that address several errors in the interface.lib ?

jkriege2:
looks good to me.

Thanks,
JAN

PS: Library reworks should be done separately, I think ... I also wanted to take a look at interface.lib in the next few days ...

Misca1234:
I will make a go on the interface.lib to fix the most obvious things

PR #1654, opened 2017-09-23
Clean up interface lib

Misca1234:
Here is a partial clean-up of the interface library.

This library have A LOT of issues, so to not overwhelming the librarians I have only corrected a few of the errors so it would be more manageable if it is handled in small portions.

This push will not fix all issues but a few.

This push consist of the following changes

XTR115U
Aligned to grid
Filled out descriptions

U2270B
Removed duplicated pins
Changed pin lenght 

TEA2014
Fixed pin lenght 
fixed frame
Fixed pin types
Should probably be moved into another library

TCA9548A_TSSOP removed, replaced with TCA9548APWR
TCA9548A_QFN removed, replaced with TCA9548AMRGER (alias TCA9548ARGER)

TCA9534
TJA1021T
TJA1021TK
TJA1029T
TJA1029TK
TJA1049T
TJA1049T/3
TJA1049TK
TJA1049TK/3
TJA1042T
TJA1042T/3
TJA1042TK/3
TJA1051T
TJA1051T/3
TJA1051T/E
TJA1051TK/3
Removed symbol name from the Description

AD5282_ABC
Removed, it already exist a AD5282

AD5290
Filled out description and metadata
Shortened the pins

AD5293
Filled out description and metadata
Shortened the pins

MC1413
Added a '* at the end of the foot print filter

MAX3226
Fixed foot print filter

bild

bild

bild

bild

bild

bild

bild


SchrodingersGat:
@Misca1234 thanks, a partial improvement is still an improvement! Thanks for your work.

Misca1234:
Just a question, 
Should I leave names like "TJA1051TK/3" as they are or replace the '/' ? with , for example, a '_' ?
The '/' is not a legal character in a name.

SchrodingersGat:
The `/` should be removed for sure. Either replace with `-` or `` (blank)

PR #1586, opened 2017-09-02
Add DRV8848 to texas library

michielbrink:
http://www.ti.com/lit/ds/symlink/drv8848.pdf

PR #1657, opened 2017-09-24
Drv8848

Misca1234:
This is replacement push of, the seemingly abandoned , push 
https://github.com/KiCad/kicad-library/pull/1586

DRV8848
http://www.ti.com/lit/ds/symlink/drv8848.pdf

bild

bild


jkriege2:
nice work!

Thanks,
JAN

PR #1658, opened 2017-09-24
Add Si7020

MatthewStephenson:
Add Silicon Labs Si7020-A20 Temperature & Humidity Sensor

Datasheet: https://www.silabs.com/documents/public/data-sheets/Si7020-A20.pdf

image

As always, suggestions and corrections welcome :)


Misca1234:
Usually the librarians want , at the end of the description line, the type of package as well
Your description would look like this 
"I2C Humidity and Temperature Sensor, DFN-6"

jkriege2:
A few remarks:
1. the package has a bottom pad, which should be floating, but is connected to GND internally (according to DS). Therefor add a pin 7 PAD, as PASSIVE-pin at the bottom of the symbol. 
2. The footprint does not match: The one you linked has a pitch of 0.95mm, instead of the 1mm from the DS ... so could you make a matching footprint in a second PR?
3. I would set the pin text offset to 20 instead of the current 40mil:
pinpadtext

Best,
JAN

Misca1234:
@MatthewStephenson 
The DFN-6 should be on the line above, on the description line

bild

> The footprint does not match: The one you linked has a pitch of 0.95mm, instead of the 1mm from 
> the DS ... so could you make a matching footprint in a second PR?
You changed to 
"Housings_DFN_QFN:DFN-6-1EP_3x3mm_Pitch1mm"

You are aware of that this foot print does not exist ?
Do you have the intention to make it ?

Misca1234:
I took the liberty to create the foot print and 3D model, see 
https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/63

jkriege2:
Nice work!

Thanks,
JAN

PR #1662, opened 2017-09-25
Added 74LS279 (Quad SR-Latch)

chrigro:
screenshot from 2017-09-25 21-52-05
screenshot from 2017-09-25 21-52-23
screenshot from 2017-09-25 21-51-54
screenshot from 2017-09-25 21-51-38
screenshot from 2017-09-25 21-51-18


Misca1234:
Why have you stacked 2,3 and 1, 12 ?

chrigro:
Indeed, I was not sure how to handle this. They are internally connected to S1 (2,3) and S3 (11,12).

Best,
Christian

Misca1234:
I have the feeling is that they should not be stacked.

The push pull output family, such as but not limited to TTL,  can not connect two or more outputs to same input so these extra inputs, 2 and 12, are used to make a more flexible construction where two control units can affect the same SR flip flop.
(however you can connect several inputs with each other, depending on technique and power specification)

I beat the SR latch 1 and 2 lower AND is a true 3 input AND and for latch 2 and 4 they are a 2 input AND so I am pretty convinced that they should not be stacked.

chrigro:
Alright. I will change it and push a new commit.

jkriege2:
please note that we start a reorganisation of the libs, which entails renaming some libs to make the overall structure more logical. Since we renamed 74xx.lib to Logic_74xx.lib, please move your parts to that new lib in order to fix the conflicts named above!

SchrodingersGat:

PR #1664, opened 2017-09-26
library:regul.lib:fix some footprint name

trabucayre:
Some symbols, like  MC79L05_SO8, have FLIST entry with error on case footprint name.
Since operating systems like Linux are case sensitive pcbnew complain about missing symbol.

This patch fix FLIST with value SOIC?8*3.9X4.9mm*Pitch1.27mm* by SOIC?8*3.9x4.9mm*Pitch1.27mm* (X upper to x lower).

Thank
Gwenhael Goavec-Merou


SchrodingersGat:
Nice work, thanks :)

PR #1666, opened 2017-09-26
Dip add suffix sm

Misca1234:
Renamed according to 
https://github.com/KiCad/Housings_DIP.pretty/pull/15


Misca1234:
I dont know wjy kicad changed in linear.lib
Just before I made the push a made a fresh clone

jkriege2:
Please revert the changes to linear.lib, before I can merge ... these also change the pin-name-relation, as far as I can see, so we really don't want them here ...

Best,
JAN

Misca1234:
Hmm, this pull have three 6 in its name 
d/kicad-library/pull/1666

Must have been the devil himself who was messing

jkriege2:
Thanks,
JAN

PR #1667, opened 2017-09-26
several fixes/modernizations to reference.lib

jkriege2:
- removed NC-pins from references (especially when 2/3-pin reference is in 8-pin package)
- used pin-stacking where it makes sense
- split REF200U into 3 units
- reworked REF3212AMDBVREP
- moved NC-pins in
- several other minor fixes (datasheets, descriptions, ...)

most of the KLC-errors left are due to pins on 50mil raster, but these are small diode-like symbols, so these should be ignorable


SchrodingersGat:
Thanks Jan these improvements look good

Misca1234:
I wonder over one replace 

$CMP TL431KTP
 D Shunt Regulator, TO-252
 K diode device shunt regulator
-F http://www.ti.com/lit/ds/symlink/tl431.pdf
+F http://www.jameco.com/Jameco/Products/ProdDS/760499.pdf

Perhaps the link to the official Texas Instrument pdf should have been keept?

evanshultz:
The Orderable Device table at the end of the TI datasheet nor the package drawings list TL431KTP, which may be why it was removed. However, the datasheet does show the package on page 3.

I do see TL431 in DPAK, but they all seem to be obsolete. So in that case, perhaps this specific part/package combination should just be removed from the KiCad library?

Issue #1663, opened 2017-09-25
NC pins in symbols


SchrodingersGat:
I think we need to readdress the NC pins rule. Currently we require that all NC pins are added to the schematic, but this makes for some really messy symbols:

https://user-images.githubusercontent.com/3242402/30396149-f8ee8b06-98c8-11e7-93a4-452812006358.png

If we drop this requirement then footprint filters may not work for some parts, (mostly generic parts as atomic ones already have their footprint associated).

When .sweet format arrives then that brings many-to-one pin association so that be good. Until then, I think we should drop the requirement that ALL NC pins must be added, and take the hit on footprint filtering (I don't think many users will find that an issue).

It makes the symbols a LOT cleaner, and it also reduces contribution complexity.

jkriege2:
Hi!

BTW: It's worse for TL431...

does the new format also bring many-to-none pin association? Basically we don't want all the NC-pins connected to each other. 

For now I would propose 2 options, from which the contributor can choose:
1. add all NC and drop the pin-count from the FPFilter (prefered if there are few and there is enough space in the symbol for NC-pins)
2. add non or only few of the NC pins and ADD the pin-count to the FPFilter (at least where that makes sense, e.g. for DIP/SOIC/...-packages) (prefered if the symbol gets significantly less messy and if the NC-pin count is comparable to or larger than the actual pin-count, or for simple devices like reference diodes)

SchrodingersGat:
I think that is sensible.

bobc:
I also came to the conclusion that it would be better to omit NC pins, particularly as they can actually get connected to a net despite the "No Connect" property.

However, I think the footprint filter must then contain the pin count. There are parts with multiple NC pins, in general how can the user know whether to select SOIC-14 or SOIC-16 without reference to datasheet? 

I've never liked leaving the pin count a wild card, we try to nail down every other parameter to avoid surprises for the user, but in this case we leave it open. If a part is available in different pin counts, then we need a separate symbol anyway? Otherwise, we can add multiple filters for each package/pin count.

SchrodingersGat:
> There are parts with multiple NC pins, in general how can the user know whether to select SOIC-14 or SOIC-16 without reference to datasheet?

To be honest, who's picking a footprint without reference to a datasheet? ;)

I think the advantages of dropping mandatory inclusion of NC pins far outweigh the downsides...

jkriege2:
[...]

PR #1669, opened 2017-09-27
Clean up interface lib part 3

Misca1234:
Here is a partial clean-up of the interface library.

This library have A LOT of issues, so to not overwhelming the librarians I have only corrected a few of the errors so it would be more manageable if it is handled in small portions.

This push will not fix all issues but a few.

This push consist of the following changes

CH376T
Changed pin 20 (VCC) from Output to Power input
Changed pin 17 (SD_CS) from output to open collector
Changed pin 13, SCS, from output to input
Changed pin 14, SCK, from output to input
Changed pin 15, SDI, from output to input
Changed pin 16, SDI, from input to output
Changed pin  5, RXD, from output to bidirectional
Changed pint 2 name from RST to RSTI

Changed foot print from 
Housings_SSOP:TSSOP-20_4.4x6.5mm_Pitch0.65mm 
to
Housings_SSOP:SSOP-20_5.3x7.2mm_Pitch0.65mm
Set FP to SSOP*5.3x7.2mm*Pitch0.65mm*

DP83848C
Aligned 4 pins to 100 grid
Set the specific foot print to Housings_QFP:LQFP-48_7x7mm_Pitch0.5mm
Set the FP to LQFP*7x7mm*Pitch0.5mm*

DP83848I
Removed, it is an alias to DP83848C

DS7820
Added metadata
Complete redraw, did not spend much time on this oldie

DS7830
Added metadata
Complete redraw, did not spend much time on this oldie

DS90C124
Filled in metadata 
aligend pins to 100 grid
shortened pins to 100 
Set FP to TQFP*7x7mm*Pitch0.5mm*
Det specific foot print to Housings_QFP:TQFP-48_7x7mm_Pitch0.5mm

All TJAXXXT/3 was renamed into TJAXXX-3

EL7242C and EL2001
Cleaned up the drawing and filled out metadata

FD1771
Added meta data 
Cleaned up symbol

FD2791 and FD2795
Can not find datasheet any place, remobed both symbols

GD65232DB, GD75232N, GD65232DW
Fixed footprint filter

bild

bild

bild

bild

bild


SchrodingersGat:
Hi @Misca1234 we have now made a decision that if there are a lot of NC pins, they can be left off the symbols - https://github.com/KiCad/kicad-library/issues/1663

The only new requirement is that the pin count must be included in any footprint filters.

e.g. `TSOP?14*` instead of `TSOP*`

So for the EL2001 as shown above, you don't need to place all those NC pins.

Will this make your life any easier for this PR? Or do you want this merged as-is?

Misca1234:
Naa, i can change it, give me an hour

Misca1234:
I have have added the pin number on those I have changed in the push,
I have not changed the FP on those already merged, this to keep down the changes in this push

jkriege2:
Hmmm ... the idea was to either add the pin number and have no NC-pins in a symbol, or have all NC-pins and no pin-number (I believe) ... @SchrodingersGat  WHat do you say? How should we interpret this?

poeschlr:
I would agree with the interpretation of @jkriege2 
Include pin number in fp filter only if the number of pins in the symbol does not agree with the number of pins needed for the footprint.

Misca1234:
The pin information in this push have been removed

SchrodingersGat:
Thansk @Misca1234 for your contributions :)

PR #1671, opened 2017-09-28
Ignore pretty folders inside the modules containter

antoniovazquezblanco:
This is a proposal.

I use KiCad master as my library. I also let KiCad download the footprints to that folder. This commits ignores the `.pretty` folders inside modules allowing me to commit small changes in the libraries without all the footprints cluttering git output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.bak
 *.bck
 internat/
+modules/*.pretty/


jkriege2:
That seems to be a quite non-standard way of doing things ... plus for the next major version, we will separate the symbols, 3D packages and footprints simply into three separate repos to make management significantly easier ... so I'm not sure whether this should be merged 

@SchrodingersGat @poeschlr @evanshultz What do you think?

evanshultz:
I'll leave it up to you all if there should be a paradigm shift, but I'm not in favor of this change. The existing plan to re-structure things is better to me.

evanshultz:
Sorry. Clicked the wrong button...

antoniovazquezblanco:
> plus for the next major version, we will separate the symbols, 3D packages and footprints simply into three separate repos to make management significantly easier

>The existing plan to re-structure things is better to me.

In my opinion this line does not conflict with that idea. When is this change planned to take place? Up until then, does this upset anyone s work flow? Does this change imply a lot of work in the future?

SchrodingersGat:
Antonio there is nothing stopping you from using this .gitignore file locally with your setup. But your directory setup is not something we should be officially accommodating

antoniovazquezblanco:
Thank you for considering it. I will change my directory setup as it seems it is not the way of doing things. 

I hope the new library structure will be adopted soon ;)

PR #1672, opened 2017-09-29
Add library symbols for SiLabs EFM8BB1 "Busy Bee" microcontrollers.

ppelleti:
[Datasheet](https://www.silabs.com/documents/public/data-sheets/efm8bb1-datasheet.pdf).

There are separate symbols for each package variant: [QFN20](https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/64), QSOP24, and SOIC16.

efm8bb1_qfn20

efm8bb1_qsop24

efm8bb1_soic16


SchrodingersGat:
Great! Thanks for your contribution

evanshultz:
Since this was merged already, here are a few notes for future symbols:
- You can stack pins that should be connected together, like QFN20 GND pins
- Our naming system is to use the full manufacturer PN to define package, and not append `SOIC16`, etc.
- Push NC pins inside the synbol so they're less likely to get accidentally connected
- I would prefer if VDD were in the middle, with ref des on the left and symbol name on the right

ppelleti:
> Our naming system is to use the full manufacturer PN to define package, and not append SOIC16, etc.

The "SOIC16" is part of the part number, although I left out some of the intervening variable information (flash size, temperature grade, and revision).  However, I think the "0" and "F" are constant, so I probably should have included them.  And maybe I should have put "x" as a placeholder for the variable information?  So that would give us:

* EFM8BB10Fxx-x-QFN20
* EFM8BB10Fxx-x-QSOP24
* EFM8BB10Fxx-x-SOIC16

efm8bb1-ordering

ppelleti:
Or should I create aliases for all of the specific part numbers?

* `EFM8BB10F8G-A-QSOP24`
* `EFM8BB10F8I-A-QSOP24`
* `EFM8BB10F8G-A-SOIC16`
* `EFM8BB10F8I-A-SOIC16`
* `EFM8BB10F8G-A-QFN20`
* `EFM8BB10F4G-A-QFN20`
* `EFM8BB10F2G-A-QFN20`
* `EFM8BB10F8I-A-QFN20`
* `EFM8BB10F4I-A-QFN20`
* `EFM8BB10F2I-A-QFN20`
* `EFM8BB10F8A-A-QFN20`
* `EFM8BB10F4A-A-QFN20`
* `EFM8BB10F2A-A-QFN20`

PR #1676, opened 2017-10-01
added 74HC7014 with DIP and SOIC packages

sandhan26:
Added KLC compliant symbols for Hex non-inverting Schmitt trigger IC 74HC7014 in DIP and SOIC packages to the existing 74xx library


Misca1234:
I guess this link 
https://assets.nexperia.com/documents/data-sheet/74HC7014.pdf
Would be the proper one

sandhan26:
yes, i used that link previously. But it only shows the SOIC14 package information. The link that i am using at present has both SOIC14 and DIP14 package information

sandhan26:
screenshot from 2017-10-01 22-16-47
screenshot from 2017-10-01 22-17-46
screenshot from 2017-10-01 22-18-13
screenshot from 2017-10-01 22-18-38

jkriege2:
Hi!

nice work, but please make symbols that are in line with the other single-gate inverters in this lib, i.e. the symbol should be split by gate+power-supply and the single-gate should look like this:
2017-10-04 19_50_21-part library editor -- d__kicad_kicad-library-review_library_74xx lib

You could simply achieve this by deriving from e.g. the existing 74HCT04 (maybe add the ST-symbol inside), or even by simply adding yours as ALIAS (if the several footprints in the existing symbols are OK).

Best,
JAN

jkriege2:
please note that we start a reorganisation of the libs, which entails renaming some libs to make the overall structure more logical. Since we renamed 74xx.lib to Logic_74xx.lib, please move your parts to that new lib in order to fix the conflicts named above!

sandhan26:
Hi you can close this pull request. I will start fresh, on Logic_74xx

SchrodingersGat:
Thanks @sandhan26

PR #1678, opened 2017-10-02
Add TMLM04103 power supply

antoniovazquezblanco:
Datasheet: https://www.tracopower.com/products/tmlm.pdf
imagen
imagen

Given the latest discussions I have not added the missing NC pins. Should I add them?

Thank you


Misca1234:
Some comments,

You could shorten the pins to the half length

It would be easy for you to add 

TMLM 04103
TMLM 04105
TMLM 04109
TMLM 04112
TMLM 04115
TMLM 04124

as aliases but different descriptions


Will you add the footprints as well ?

I can recommend to do the foot print first, then the symbol will get the right link as well because I guess the foot print will get a name like this 

ACDC_Converter_Traco-TMLM-04XXX (which means that you have to add the NC pins)
or
ACDC_Converter_Traco-TMLM-04XXX_Single

(if it should be in line with those that is already there)

antoniovazquezblanco:
imagen

Updated symbol

SchrodingersGat:
Thanks!

PR #1338, opened 2017-06-22
added MAX17572, MAX17574

ilya-epifanov:

PR #1679, opened 2017-10-02
Max17572 max17574

Misca1234:
Re push of the , presumably, abandoned push
https://github.com/KiCad/kicad-library/pull/1338

Push consist of 
MAX 17572
https://datasheets.maximintegrated.com/en/ds/MAX17572.pdf

MAX 17574
https://datasheets.maximintegrated.com/en/ds/MAX17574.pdf

bild

bild

bild

bild


Misca1234:
The travis errors should be ok

sandhan26:
Hey, just a newbie question. What are these blue texts alongside the pins. What is their importance, adn do i need to implement them in my designs? If yes, how?

SchrodingersGat:
@sandhan26 they are the 'electrical type' of the pin (input / output / power / etc). These are optionally displayed in the symbol editor but are not shown in the schematic.

sandhan26:
Hi, in my designs, i did set the electrical pin type. But i did not find any option to display it on the symbol editor.

poeschlr:
The feature mentioned by oliver is only available in nightly builds.
For stable builds (4.x) the only way to get an overview is by looking at the pin table.

Misca1234:
@sandhan26 
look here 
http://kicad-pcb.org/download/windows/
Under
Nightly Development Builds

Download and use one of those, then you get the blue lines, very helpful for the librarians
you might have to click on the new icon to the left

bild

PR #1680, opened 2017-10-02
Update SiLabs "Busy Bee" microcontrollers, based on feedback from PR.

ppelleti:
I received some feedback from @evanshultz in [my original PR for the EFM8BB1](https://github.com/KiCad/kicad-library/pull/1672) after it had already been merged.

I have made the following changes:

* Use the full manufacturer part number (and therefore, make an alias for each part number).
* Moved NC pins inside the symbol.
* Put VDD pin in the middle of the top edge.
* Put reference designator to the left of VDD, and value to the right of VDD.

I did not stack the GND pins.  Unless that is absolutely required, I would prefer not to do it, because I like being explicit about what I'm connecting.

efm8bb10f8g-a-soic16

efm8bb10f8g-a-qfn20

efm8bb10f8g-a-qsop24


SchrodingersGat:

Issue #1684, opened 2017-10-04
VSOURCE symbol of pspice.lib is too huge


efferre79:
I have just installed kicad and kicad-libraries from the github repositories compiling from source. As a tutorial I am trying to draw a schematic and simulate it with ngspice. I have loaded the pspice.lib (/usr/share/kicad/library/pspice.lib) but the symbol for the VSOURCE is definitely too huge comparing to the other symbols, here is a snapshot:

![snap](https://user-images.githubusercontent.com/4586973/31166195-32d752d2-a8ee-11e7-91ca-b864ef9739ab.png)

Looking at the video tutorial https://www.youtube.com/watch?v=2WDPzW6DGzQ, it looks regular. Is it a problem of the library or kicad?

SchrodingersGat:
It does look like it needs to go on a diet. Are you able / willing to submit a pull request to address this?

efferre79:
I have just started using kicad so I don't have confidence at all with the software and the libraries, I don't think I am able right now to fix the issue by myself :-( But I am available to test the changes in the library.

SchrodingersGat:
I have submitted a change (see above).

SchrodingersGat:
This has been fixed now, thanks for bringing it to our attention.

efferre79:
I have tested the changes, now it looks better but I think it can be made smaller. I am attaching another attachment below where I put the voltmeter symbol (from device.lib) as a reference, in my opinion the circle of VSOURCE can be made as small as the one for the voltmeter.

![snap1](https://user-images.githubusercontent.com/4586973/31234441-30690aac-a9ef-11e7-98f4-e9eb0378f589.png)

What do you think?

PR #1690, opened 2017-10-05
Improved Vsource and Isource symbols

SchrodingersGat:
Makes the Vsource and Isource symbols smaller, fixing issue https://github.com/KiCad/kicad-library/issues/1684


jkriege2:
Hi @SchrodingersGat ... those symbols seem all a bit off in style ... do you know how the PSPICE integration works? Couldn't one simply use the existing symbols? If not, we should make sure these have unique names (When placing an R from pspice.lib, you actually get the R from device.lib) and they should conform in style to the existing symbols?

best,
JAN

SchrodingersGat:
Honestly I have no idea. I don't know what these symbols are for.

evanshultz:
I'd like to take a look. I've got quite a bit of SPICE and other simulator experience. I believe the interface is a bit rough right now. If this isn't an emergency to figure out, I'll come back to it in a few weeks.

jkriege2:
In think it's not an emergency ;-) ... just saw it when reviewing Oliver's PR.
Maybe we can contact the developer, responsible for the integration to explain this?

Best,
JAN

PR #1695, opened 2017-10-05
Fix TO-220 footprints names

evanshultz:
From discussion and updates at https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/29#issuecomment-334056217, this fixes all but audio.lib. My other PR is addressing that library so it will be done separately.

I searched for "watt" in each lib to identify TO-220 footprints using the old names.

Can someone please merge this soon? I'm afraid it will get stale quickly since it touches a number of other libraries. Thanks!


evanshultz:
The KLC violations are from pin placement in regul.lib, which isn't an issue. Please merge as-is.

SchrodingersGat:
Thanks!

PR #1696, opened 2017-10-06
added 74HC014 non-inverting Schmitt Trigger to Logic_74xx.lib

sandhan26:
I  need suggestions on adding Alias. i don't understand this feature exactly.

Added symbol for 74HC7014 non inverting Schmitt trigger.
schmitt1

schmitt2


Misca1234:
You should not state the pin number in the FP filter
Your chip is using capsules
DIP-14_W7.62mm
SOIC-14_3.9x8.7mm_Pitch1.27mm

So your FP should be
DIP*W7.62mm*
SOIC*3.9x8.7mm?Pitch1.27mm*

sandhan26:
To @Misca1234 is there any specific reason or its just a convention to not do that?

I already did the same in Espressif library. So, i need to modifiy it as well?
and Bosch library too?

sandhan26:
The Footprint Filter options depend on what foot print filter patterns we supply. 
But with the current style of patterns DIP*W7.62mm* or DIP*14* or DIP*14*W7.62mm* , only the first filter seems to work i.e. filter footprint by keyword. 

screenshot from 2017-10-08 19-56-04

jkriege2:
Currently we accept one of two topions:
1. if a symbol has all pins defined that are available in the footprint (if there are NC-pins, add them as invisible NC-pins to the symbol), don't put the pin-count into the FPFilter (the pin-count-filter will filter the matching footprints for you)
2. If you have a symbol with many NC-pins or can't find a way to put them into the symbol, don't add them, but add the pin-count to teh FPFilter.

Hope that helps,
JAN

sandhan26:
@Misca1234  @jkriege2 Thanks for your suggestions and information. I have removed the pin numbers accordingly from the FP Filter.

sandhan26:
Can't we have a regex expression build into KLC to check the same?

sandhan26:
whats the difference and use cases between setting a FP filter and setting  Footprint in the field properties?

screenshot from 2017-10-10 20-20-47

evanshultz:
The default footprint will be, well, the default footprint. When assigning footprints to symbols in CvPcb it is the footprint selected.

The FPfilter allows other footprints to be listed using the footprint filter in CvPcb. This is helpful when there are variants of the default footprint, like ones with thermal vias or ones with bigger pads for handsoldering. Is it essentially a regex (note the question mark and asterisk).

I have a convenient test design with a couple parts in it so let's see. Here is the default footprint:
image

And how the filtering works in CvPcb (red boxes around the filter buttons and FPfilter list):
image

sandhan26:
To @evanshultz In this case, since there is default footprint, i think it will be sufficient with using only FP filter. 

So, i think. 74HC7014 part is pretty much complete now.

evanshultz:
@sandhan26 
Can you add an official datasheet instead of http://pdf.datasheetcatalog.com/datasheet/NXP_Semiconductors/74HC7014.pdf? https://assets.nexperia.com/documents/data-sheet/74HC7014.pdf would be good.

Also, the official Nexperia datasheet only shows the SO package. And checking Digi-Key, the DIP package is unavailable and shows up as Obsolete. It appears that only SO is still being produced. Can you please confirm that for yourself? If so, there should be a default footprint and also just a single FPfilter. (If both DIP and SO were available, you're correct that there should be no default footprint and the two FPfilter entries.)

sandhan26:
@evanshultz : a quick search through element14, digikey and mouser, revealed, that DIP package is almost gone, except this one [instance](https://www.digikey.in/products/en/integrated-circuits-ics/logic-buffers-drivers-receivers-transceivers/704?k=74hc7014&k=&pkeyword=74hc7014&pv69=80&mnonly=0&ColumnSort=0&page=1&quantity=0&ptm=0&fid=0&pageSize=25) in digikey.in. 

But, the part number shown is 74HC7014N, where N should be D for DIP packages. So, i am doubtful about it.

So, i have made SOIC as default footprint, but also kept DIP in the FP filter.

I have also updated the datasheet link to the official nxp datasheet as provided by you.

Also,corrected the SOIC package dimensions to SOIC-14_3.9x8.7mm_Pitch1.27mm.
But, i still have some doubts on considering the correct dimensions from the datasheet. It will be great if you can provide some insight on this. Refer page 8, datasheet (https://assets.nexperia.com/documents/data-sheet/74HC7014.pdf)

evanshultz:
@sandhan26 
The footprint you chose is correct. This will fit our 3.9x8.7mm footprint.

FYI, please always include the datasheet in the original PR. That will make it easier to review.

I took a closer look at your submission and have a few requests before merging:
1. The description needs to be updated to fix a typo and remove the DIP package:
`D Hex non-inverting buffer with precision Schmitt Trigger inputs, SOIC-14`
1. Also update the keywords so they're not an exact copy of the description. Something like:
`K Hex Schmitt buffer`
1. Remove the DIP package from the FPfilter and change the question mark to an ansterisk so it will be only (sorry for the incorrect guidance above):
`SOIC*3.9x8.7mm?Pitch1.27mm*`

After that I will merge!

evanshultz:
@sandhan26 
Just one thing: Can you please update the keywords as I mentioned in #2 above? It should not be a sentence or description, just a space-separated list of identifying words.

sandhan26:
@evanshultz i just thought the keyword 'non-inverting' will be necessary. Don't you think?

evanshultz:
It is not necessary. The word "buffer" implies non-inversion, as a buffer's output will be identical to the input but isolated (electrically, optically, etc.). But to avoid discussion, I'll just merge.

sandhan26:
@evanshultz : Didn't know that. Thanks

PR #1697, opened 2017-10-06
Split sensors library

SchrodingersGat:
This PR splits the `sensors.lib` library into multiple `Sensor_xxx` functional libraries.

Captures all current symbols in the sensors library.


SchrodingersGat:
It appears that splitting these libraries out piece-meal will be too messy. I will close this PR and continue with the original plan of improving the transfer scripts such that the entire move can be made in bulk for v5 release.

PR #1698, opened 2017-10-06
Split memory libraries

SchrodingersGat:
Split contents of memory.lib into

* Memory_RAM
* Memory_NVRAM
* Memory_Flash
* Memory_EEPROM

With some very poor quality and obsolete symbols left in memory.lib

Unallocated Symbols:
memory : MS628128
memory : SIM4X32
memory : SIMM
memory : TMS4C1050N
memory : UPD42280V


SchrodingersGat:
Note: I have not attempted to fix any of the outstanding KLC issues in these libs which is why they still exist

SchrodingersGat:
It appears that splitting these libraries out piece-meal will be too messy. I will close this PR and continue with the original plan of improving the transfer scripts such that the entire move can be made in bulk for v5 release.

PR #1456, opened 2017-07-27
Da dc04 11

Misca1234:
Push for DA04-11 and DC04-11

kicad-library push branch is 
https://github.com/KiCad/kicad-library/pull/1456

Displays_7-Segment.pretty push branch is 
https://github.com/KiCad/Displays_7-Segment.pretty/pull/10

packages3D push branch is 
https://github.com/KiCad/packages3D/pull/96

packages3D-source push branch is 
https://github.com/KiCad/packages3D-source/pull/30

DA04-11CGKWA
http://www.kingbright.com/attachments/file/psearch/000/00/00/DA04-11CGKWA(Ver.6A).pdf

DA04-11EWA
http://www.kingbrightusa.com/images/catalog/SPEC/DA04-11EWA.pdf

DA04-11GWA 
http://www.kingbrightusa.com/images/catalog/SPEC/DA04-11GWA.pdf

DA04-11SRWA
http://www.kingbrightusa.com/images/catalog/SPEC/DA04-11SRWA.pdf

DA04-11SEKWA
http://www.kingbright.com/attachments/file/psearch/000/00/00/DA04-11SEKWA(Ver.9A).pdf

DA04-11SURKWA
http://www.kingbright.com/attachments/file/psearch/000/00/00/DA04-11SURKWA(Ver.10A).pdf

DA04-11SYKWA 
http://www.kingbright.com/attachments/file/psearch/000/00/00/DA04-11SYKWA(Ver.6A).pdf

DA04-11YWA
http://www.kingbrightusa.com/images/catalog/SPEC/DA04-11YWA.pdf

DC04-11CGKWA
http://www.kingbright.com/attachments/file/psearch/000/00/00/DC04-11CGKWA(Ver.7A).pdf

DC04-11GWA 
http://www.kingbrightusa.com/images/catalog/SPEC/DC04-11GWA.pdf

DC04-11SEKWA
http://www.kingbright.com/attachments/file/psearch/000/00/00/DC04-11SEKWA(Ver.7A).pdf

DC04-11SURKWA 
http://www.kingbright.com/attachments/file/psearch/000/00/00/DC04-11SURKWA(Ver.6A).pdf

DC04-11SYKWA 
http://www.kingbright.com/attachments/file/psearch/000/00/00/DC04-11SYKWA(Ver.6A).pdf

bild

PR #1700, opened 2017-10-06
Da dc04 11 2

Misca1234:
This is a re push of 
https://github.com/KiCad/kicad-library/pull/1456

It have happened to much in display.dcm and display.lib since this push creation.
It is easier and safer to make a re push so this push will be closed and re push


SchrodingersGat:
Thanks @Misca1234

PR #1702, opened 2017-10-07
Added Bosch sensors library with BME680 and BME280 sensors

sandhan26:
1. The SDI and SDO pins of both the IC's have been set to bi-directional as per the datasheet. 
     But, KLC doesn't seem to approve it.

2. I  need suggestions on adding Alias. i don't understand this feature exactly.

3. Both have LGA-8 package type, but 680 has 3mmx3mm and 280 has 2.5mmx2.5mm. Both of which  seems to be unavailable.

#### 3 in 1 digital humidity, temperature & pressure sensor.

bosch-bme280_1

bosch-bme280_2

------------------------------------------

#### 4 in 1 digital humidity, temperature, pressure & gas sensor
bosch-bme680_1

bosch-bme680_2


Misca1234:
> The SDI and SDO pins of both the IC's have been set to bi-directional as per the datasheet.
> But, KLC doesn't seem to approve it.

it can be ok, the KLC test script is there to catch the most common and obvious errors,
The librarians might still approve a push with travis errors, as long as the errors is in fact positive false.

> I need suggestions on adding Alias. i don't understand this feature exactly.

Aliases are used for a component that have variations, such as electrical specifications, temperature or other differences but have only one footprint(s).
So if you create a symbol and gives it a number of aliases and a foot print
you can edit each of those aliases in the symbol editor and give each specific alias their own description but all aliases use same symbol and in the long run same foot print.

> Both have LGA-8 package type, but 680 has 3mmx3mm and 280 has 2.5mmx2.5mm. Both of which seems to be unavailable.
Yes, see below



I am not sure you got the foot print correct
Your filter is LGA?8
and the only one that match in the libraries is

bild

which have the size of 3x5 but according to the datasheet
https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME280_DS001-11.pdf
the capsule is 2.5x2.5

I can recommend that if you are going to add a component which foot print does not exist is it to create the foot print (and of course the 3D models ^^) first or make them at same time and refer to them crosswise.

sandhan26:
What naming convention do i adopt for the new footprints

LGA-8_2.5x2.5mm_Pitch0.65mm

or 

BOSCH_LGA-8_2.5x2.5mm_Pitch0.65mm ?

sandhan26:
Also apart from * in footprint filters, is there any other simple like ?
You used it in your comment as LGA?8

Difference between the two?

Misca1234:
The '?' will only wildcard one character
The '*' will wildcard from 0 to infinite number of characters different from the character next to the right of the '*'

So 

"Schaffner__1EP_3x2"

Will match for "Schaffner?_1EP_3x"
Will not match for "Schaffner?1EP_3x"

But will match for "Schaffner*1EP_3x"

Misca1234:
If you want to push the foot print in the same directory as the 
LGA-8_3x5mm_Pitch1.25mm

you fork and update the 
https://github.com/KiCad/Housings_LGA.pretty

sandhan26:
I am on it. Thanks for the suggestions.

sandhan26:
corrected the FP Filter. Removed pin number. A 2D footprint coming soon.

SchrodingersGat:
Hi @sandhan26 it looks like the Logic symbol has accidentally been added here - can you remove this so we can merge just the Bosch sensors?

sandhan26:
@SchrodingersGat Its really easy to mess things up. Thanks

sandhan26:
added footprint to [kicad.Housings_LGA repository](https://github.com/KiCad/Housings_LGA.pretty/pull/10) 

screenshot from 2017-10-13 00-39-44

sandhan26:
@SchrodingersGat  how do i resolve this two conflicts?

SchrodingersGat:
I am not sure why you are getting conflicts on those files if you have not touched them. Try merging kicad/master into your new branch and you should be able to fix the conflicts there.

sandhan26:
to @SchrodingersGat as per your advice, i have been able to resolve the conflicts. Kindly check, if i haven created any further mess. 

Its too much problem solving for a git newbie!!!

Misca1234:
If it get to messy I usually create a new branch and copy the component into the new branch and make a brand new push.
Then I close and delete the faulty one.

If you look here, you can see what your push contain for changes (click-able link)

bild

sandhan26:
To @Misca1234 thanks for the info. At present, only the two files related to Bosch library have changed, which is a good news.

Also, yesterday i have noticed 2-3 minutes delay after pushing my changes. Initially, it was driving me nuts. Do these kind of delays frequently happen in github?

sandhan26:
@SchrodingersGat since there are no conflicts, you may merge this branch, if you see fit.

PR #1704, opened 2017-10-07
created WS2813

Gansemarsch:
Created WS2813, an improved version of the WS2812b

Datasheet: https://www.element14.com/community/servlet/JiveServlet/previewBody/82167-102-1-346252/WS2813%20LED%20Datasheet.pdf

KLC Newbie Question: The footprint I filtered for is named "LED_WS2812-PLCC6.kicad_mod", should the name of that footprint (in LEDs.pretty) be changed/made generic?

image


Misca1234:
You should change the datasheet link to 
http://www.normandled.com/upload/201605/WS2813%20LED%20Datasheet.pdf

Why do you have different pin lenght between 2,4,6 and 1,5 ?

Normally the product id, that is WS2813, should not be in the description line and key word line in the properties

Gansemarsch:
_You should change the datasheet_
Fixed, thanks for the better link.

_Why do you have different pin lenght between 2,4,6 and 1,5 ?_
I had done this because the 4.1 says that pins must be on a 100 mil grid but nothing about pin lengths being the same. Fixed(?) in latest commit.

_Normally the product id, that is WS2813, should not be in the description line and key word line in the properties_
Fixed

_Pin 1_
Fixed

_Pin 2_
Fixed

_According to page 3 in the data sheet right top most image you can see that the soldering pad require a thermal vias._
I'm sorry but I just don't see the thermal pad that you are talking about. Is it the red circled area in the image below that you mean, because it doesn't show the same hatching for the hole in the soldermask as pads 1-6.

image

poeschlr:
@Misca1234, @Gansemarsch I would not include Thermal vias in the footprint filter. If we do that there will always only be one footprint shown as a result of the filter. (Defeats the point of having it.) But of course the footprint field should be filled out such that it points to this footprint.

So if the footprint is named: LED_Normandled_WS2813-06-1EP_5.0x5.0mm-Pitch1.6mm_ThermalVias
The filter would be: `LED?Normandled?WS2813?06?1EP?5.0x5.0mm*Pitch1.6mm*` (Also notice i change some ? wildcards with *)

Misca1234:
Perhaps you are right that the component don't have a thermal vias but it is some details of the component itself they show, but it sure do look like one (the red circle in the image).
However they got their pin table and pin image different so who knows what they mean.

Gansemarsch:
Updated filter to `LED?Normandled?WS2813?06?5.0x5.0mm*Pitch1.6mm*` because we don't think there is a thermal pad.

I also made a footprint named `LED_Normandled_WS2813-06_5.0x5.0mm_Pitch1.6mm` in LEDs.pretty. See: https://github.com/KiCad/LEDs.pretty/pull/47

antoniovazquezblanco:
I believe there is no thermal pad:

50-1000pcs-font-b-ws2813-b-font-6pins-5050-smd-black-white-version-font-b-ws2813

Also interested in this component ;)

Gansemarsch:
The footprint for the WS2813 has been merged: https://github.com/KiCad/LEDs.pretty/pull/47

Is there anything else preventing this from being merged?

SchrodingersGat:
Yes one final thing. In addition to setting the footprint filter, please assign the merged footprint to the footprint field.

SchrodingersGat:
@Gansemarsch if you can contribute a fix for this in the future, please do so.

PR #1709, opened 2017-10-08
Lt3080

Misca1234:
This is a repush of the , seimgnly abbonned puh of LT3080
https://github.com/KiCad/kicad-library/pull/1331

LT3080
http://cds.linear.com/docs/en/datasheet/3080fc.pdf

Note, LT3080xQ "wants" a DD-PAK, I used TO-263 which is very close
Is this ok or should i create the foot print according to 
http://cds.linear.com/docs/en/packaging/Q%20DD5%2005-08-1461%20Rev%20F.pdf

bild

bild

bild

bild

bild


SchrodingersGat:
DDPAK / D2PAK / TO-263 are nominally the same. The TO-263-5 in the library should be fine. THanks for the contribution :+1:

PR #1713, opened 2017-10-09
Vigortronix 10 w v2

Gasman2014:
Add 3d models for 10W Vigortronix PSU
screen shot 2017-05-29 at 18 43 12
[10WattAC-DCConverters.pdf](https://github.com/KiCad/kicad-library/files/1369956/10WattAC-DCConverters.pdf)


SchrodingersGat:
I think something has gone wrong with this PR - did you make the branch from a recent kicad-library commit? There are a lot of changes here that are unrelated to your model files.

Gasman2014:
I have refreshed the baseline and made a new branch and a new pull request.Sorry I'm not making a lot of progress on this and if the new PR doesn't work I will abandon it as I can't spare any more time on this (there has to be a better way than this - I am happy making library components, footprints and models and would love to share them but the whole Git thing is just too much to get my head around. I think the problem is related to the fact that I have the my Modules in a directory outside of the Git library as I download those using the git tool within Kicad to manage them. 

Would it not be possible to have a workflow where exported libraries/footprints and models could be submitted without using git. I think a lot of potential contributors are put off by the complexity.

evanshultz:
Git is the only method currently supported by KiCad.

You can use the GitHub web interface, or the GitHub application, instead of command-line Git. Perhaps that is simpler? I believe you'll find lots of info on the web, or please ask if you'd like tips on that avenue.

PR #1716, opened 2017-10-11
Change Vbat_sense line of MCP72871 to input instead of bidirectional.

reportingsjr:
The datasheet claims it is an I/O, but the block diagram shows it as
input only. ERC complains if you use it as intended (bidi can't
be tied to a Power Out pin) which this PR fixes.


SchrodingersGat:
Thanks Jon

PR #1718, opened 2017-10-11
Add Analog Devices ADP7142 and ADP7182 LDO regulators to regul library

aurel32:
These two LDO regulators are often used together to regulate both positive and negative supplies.

They pass the KLC checks, except 4.5, Pin orientation: Wherever possible, pins should be arranged by function. I believe that it is traditionally not done that way for voltage regulators, and it is not done for other chips in the same library.


SchrodingersGat:
thanks @aurel32 :) great work

PR #1720, opened 2017-10-12
Fix decription for Q_Photo_NPN_EC.

ppelleti:
`Q_Photo_NPN_EC` had the description "Phototransistor NPN, 2-pin (C=1, E=2)".  I think it should be the other way around.


jkriege2:
thanks for spotting!

JAN

PR #1724, opened 2017-10-14
Add USB_B_Micro and USB_B_Mini schematic symbols.

ppelleti:
Strangely, there are already some footprints for micro USB connectors in the [Connectors_USB library](https://github.com/KiCad/Connectors_USB.pretty) but I couldn't find any schematic symbols for a micro USB connector.  (USB micro B and mini B connectors have an additional pin, the `ID` pin, which is not present in the traditional USB B connector, so the existing `USB_B` symbol can't be used.)

So, I have added a symbol `USB_B_Micro`, based on the existing `USB_B` symbol.  I've also added `USB_B_Mini` as an alias for `USB_B_Micro`, because the micro and mini connectors have the same pinout.

The pinout is [here](http://pinouts.ru/PortableDevices/micro_usb_pinout.shtml).

usb_b_micro


SchrodingersGat:
Thanks!

PR #1728, opened 2017-10-15
Add RFM69HW radio module

MatthewStephenson:
Add RFM69HW radio module.

Datasheet: http://www.hoperf.com/upload/rf/RFM69HW-V1.3.pdf

image

Note: The automated checks fail on the NC pin, which the checker wants to be invisible and of type NC. However, the datasheet says it should be connected to GND (despite being called "NC", go figure), so I have left it visible so the user can connect it to ground. I would stack it with the other GND pins, but KLC says they have to be the same name, and they're not...

Footprint PR: https://github.com/KiCad/RF_Modules.pretty/pull/26

As always, comments and suggestions welcome.


SchrodingersGat:
You are correct in saying that this should be an exception to the KLC error. Thanks for your contribution :)

PR #1731, opened 2017-10-16
MCU_ST_STM32: Correct descriptions of STM32F446Zxx

piotr-wiszowaty:
I'd change the rest of the descriptions in this manner (without checking MCU parameter correctness, as in this case). Or even further, e.g.:
Core: ARM Cortex-M4, Package: LQFP144, Flash: 256 kB, RAM: 128 kB, Frequency: 180 MHz, Voltage: 1.7..3.6V, I/O-pins: 114
Rationale: "KB" - Kelvin times something vs.  "kB" - kilobyte, "Ram" - a male sheep vs. "RAM" - Random Access Memory.
Is there any policy on this?

### DIFF
--- a/library/MCU_ST_STM32.dcm
+++ b/library/MCU_ST_STM32.dcm
@@ -2839,19 +2839,19 @@ F http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datas
 $ENDCMP
 #
 $CMP STM32F446ZCTx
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
We can't have sheep running around everywhere. Thanks for the fix. 

The policy is "abbreviations should be capitalised as per their context". So you have done the right thing here.

PR #1735, opened 2017-10-17
added DirectFET scripted 3D models

Shackmeister:
directfet_3d


easyw:
Hi Frank,
it seems the Step file has 2 parts instead of a single one (fusion doesn't get fine)
Normally it is enough to have a small overlap between the 2 obj to fuse
df-2p

Shackmeister:
@easyw since you made such a big deal about a broken step file it should now be fixed ;)
thanks for the check :)

Issue #1627, opened 2017-09-17
KLC: Connector schematic symbol conventions


matthijskooijman:
I was looking at some connector symbols today and noticed that there is quite some difference in styles. Looking at the KLC, I couldn't really figure out what the recommendation is for them.

In particular, I noticed this symbol (atmel:AVR-JTAG-10):

![image](https://user-images.githubusercontent.com/194491/30521963-00d43ae8-9bc9-11e7-9ade-2d4fd07f4c60.png)

It violates some KLC constraints (origin is not in the middle, pin origin is on a 50mil grid, not 100). I'm not sure what KLC says about the visual style of connectors.

This particular symbol also has pin labels on the outside instead of on the inside (as is customary for filled blocks for ICs), probably because the visual style resembles the common (but not only) "shrouded header" footprint used for this connection. Because the labels are centered on the pins, any wires that you would connect go right through the labels, which doesn't really look good.

As for the visual style, it seems that there's quite some variation. The generic `CONN_*` symbols use a basic block with pins:

![image](https://user-images.githubusercontent.com/194491/30552519-8281b0f4-9c9e-11e7-9af0-03a5ce2aebf9.png)

But more specific connectors use more graphical representations, with or without fill. Here's a fairly random selection from the `conn` library:

![image](https://user-images.githubusercontent.com/194491/30551847-3ad5dcdc-9c9c-11e7-9c66-347ae9fbef19.png)

So, the question is: what policies do we (want to) have for connector symbols? I assume they should adhere to the regular policies (e.g. pin grid alignment) wherever possible (though a particular visual style might conflict with that), but are there more specific ideas for the visual style?

As for the AVR-JTAG connector, which I originally wanted to improve, I guess it would be good to keep the "shrouded header" visual style, but realign the pins and move the labels slightly out of the way to be more useful? Any other specific suggestions for this part?

bobc:
I don't like the overly graphic connector symbols, IMO the schematic should show logical connections, I don't need a picture of the connector in the schematic. Aesthetics are highly subjective, so these symbols tend to change every few months when someone decides they don't like the look of the current symbols. 

The AVR-JTAG symbol is particularly horrible. A simple box with labels on inside would be much better.

Unless there are labelled pins like USB, I tend to use the generic CONN_xxyy symbols.

poeschlr:
You should update your libs before you make an issue. The dilxx connectors do no longer exist and the screw terminals now use the same pin spacing as other connectors.

[...]

PR #1736, opened 2017-10-17
AVR-ISP-6: Replace symbol

matthijskooijman:
This PR changes the AVR-ISP-6 connector from:
selection_004
to:
selection_006

This should more closely follow the KLC, supplemented by the discussion
at #1627. I've just done AVR-ISP-6 now, if this is accepted, I'll also
do AVR-ISP-10 and AVR-JTAG-10.

As suggested in #1627, I've used a logical layout (as opposed to mimicing the
physical 2x3 pin layout). An extra advantage is that the layout of the
AVR-ISP-10 connector will be identical (it only adds more GND and NC pins), so
the symbols are interchangeable.

I can imagine (as I suggested in #1627 as well) that it can be useful to have
an alternative version of the symbol that does follow physical layout,
especially for the AVR-ISP-10 symbol, which has multiple GND pins and one NC
pin that you might want to wire differently. You can always use a generic 2x3
connector symbol for that, of course, but that does not have the pins labeled
with their names (but I guess it's enough of a corner case to just accept this
minor downside).

The new part does not completely pass checklib.py, remarks below.

        $ checklib.py -p AVR-ISP-6 -v -v atmel.lib
        Checking symbol 'AVR-ISP-6':
            Component AVR-ISP-6 has metadata errors:
             - Missing DOCUMENTATION FILE NAME field on 'Properties > Description' tab
            Minuses and underscores in footprint filter 'IDC-Header*2x03*' should be escaped with '?' or '*'.
            Minuses and underscores in footprint filter 'Pin_Header*2x03*' should be escaped with '?' or '*'.
          Violating Rule 4.6 - Pin types
            Pin electrical type should match pin function
            Power pins should be of type POWER INPUT or POWER OUTPUT
             - Pin VCC (2) @ (-100,500) is of type PASSIVE
             - Pin GND (6) @ (-100,-400) is of type PASSIVE

    - Missing DOCUMENTATION FILE NAME field on 'Properties > Description' tab

I could not find any official documentation on this particular pinout from
Atmel (only some images as part of documentation on specific programmers), so I
left this blank. KLC says "Documentation field contains URL to part
documentation if applicable, or is left blank", so this error seems too strict?

    Minuses and underscores in footprint filter 'IDC-Header*2x03*' should be escaped with '?' or '*'.
    Minuses and underscores in footprint filter 'Pin_Header*2x03*' should be escaped with '?' or '*'.

I'm not sure what is meant here. The KLC does not mention this rule at
all, and the examples aren't clear either.. I can imagine that all `_`
that separate different parts of the name be replaced with `*` so you
could potentially insert more qualifiers later and still match (but that
does not explain why you would wait to replace them with `?`). Also,
replacing a the `-` or `_` with `*` or `?` in these cases seems silly,
since they are just contained in a single part of the name (e.g. "IDC"
and "Header" aren't really separate parts AFAICS).

    Power pins should be of type POWER INPUT or POWER OUTPUT
     - Pin VCC (2) @ (-100,500) is of type PASSIVE
     - Pin GND (6) @ (-100,-400) is of type PASSIVE

This part contains pins named VCC and GND, but this is just a connector, so
they can be either INPUT or OUTPUT, depending on how the connector is used. I
think this warrants an exception to this rule?


jkriege2:
Hi!

about the FPFilters, simply use `IDC?Header*2x03*` and `Pin?Header*2x03*` ... this way the FPFilter is still valid, even if someone should rename `'Pin_Header` to `'Pin-Header`.

ABout the symbol itself, could you add small squares that we commonly use for other connectors, e.g. 
2017-10-17 20_53_03-library browser d__kicad_kicad-library_library_connector lib
2017-10-17 20_52_43-library browser d__kicad_kicad-library_library_connector lib

matthijskooijman:
> about the FPFilters, simply use IDC?Header*2x03* and Pin?Header*2x03* ... this way the FPFilter is still valid, even if someone should rename 'Pin_Header to 'Pin-Header.

Ah, so that's the idea. Sounds like there would be better ways to solve this (e.g. let KiCad match _ as - and vice versa), but if this is the current policy, I've updated the symbol to match.

> ABout the symbol itself, could you add small squares that we commonly use for other connectors, e.g

Oh, I was planning to, but apparently forgot. I now added them, using 10x30mil boxes (seems there is some variation, the generic connectors use 10x50mil, the USB3 connector you linked uses 20x40mil, the USB1/2 connectors use 10x30mil). I used 30mil long boxes, since 40 mil comes uncomfortably close to the text.

jkriege2:
nice work!

Thanks,
JAN

matthijskooijman:
Thanks for merging! I'll try to do AVR-ISP-10 and and AVR-JTAG in the same style soon.

jkriege2:
Great!

JAN

matthijskooijman:
@jkriege2, I just submitted #1810 for the other AVR connectors.

PR #1737, opened 2017-10-17
add max31856

ptooley:
Add footprint for the Maxim 31856 thermocouple reader, see https://datasheets.maximintegrated.com/en/ds/MAX31856.pdf following the style of the 3182x parts already in sensors.lib
max31856


Misca1234:
Change description from 
Precision Thermocouple to Digital Converter with Linearization
To
Precision Thermocouple to Digital Converter with Linearization, TSOP-14

I guess the Footprint filter should be
TSSOP*4.4x5mm?Pitch0.65mm*
but I am a little uncertain

ptooley:
okay, that is done. I assumed you meant TSSOP-14 in the description and I have retained filtering to 14 pin TSSOPs

jkriege2:
Sorry for the long wait ... very nice work. Could you please do some final corrections:
1. DRDY and FAULT ar OUTPUTS as stated in the datasheet
2. move the end of the NC pin onto the outline, so it's less likely to erroneously connect to it
3. set the pin label offset to 20mil to improve visual appearance

Thanks,
JAN

2017-11-06 21_14_32-part library editor -- d__kicad_kicad-library-review_library_sensors lib

ptooley:
@jkriege2 hopefully that should be everything now.

jkriege2:
nice work!
Thanks,
JAN

PR #1738, opened 2017-10-17
MCU_ST_STM32: Add STM32F722/732 MCU in LQFP64 package

aurel32:
This patch adds STM32F722/732 MCU in LQFP64 package, it passes the KLC checks. Note however that I didn't generate it with the stm32-kicad-lib, as I believe listing all the pin alternatives doesn't make sense for big MCUs. It creates a symbol that does not even fit on an A4 page and adds very limited value. Instead I used the same pin names as on page 34 of the datasheet.


SchrodingersGat:
@aurel32 thanks for the contribution - the other STM parts are a mess, with the extended pin names.

PR #1742, opened 2017-10-18
ESP-12E

piotr-wiszowaty:
The documentation file is not strictly for this module, but this is the closest I could find on the
manufacturers website. The modules have compatible pinouts most likely.


Misca1234:
From where do you know the meaning of pin 9 - 14 ?

If the data sheet of ESP-12E differs to much from ESP-12F it might not be a good idea to have it as a reference, have you contacted ai-thinker and asked them for a url to the ESP-12E ?

Here is a link 
`http://www.kloppenborg.net/images/blog/esp8266/esp8266-esp12e-specs.pdf`
Which probably is your device,

One idea could be that you download this pdf
create directory `kicad-library/template/ai_thinker_ESP-12E` 
And put the url there, and describe where the url is in the datasheet field

But that must a librarian answer if it is a good idea.

piotr-wiszowaty:
On 10/19/2017 06:53 PM, Stefan wrote:
>
> From where do you know the meaning of pin 9 - 14 ?
>
 From the document you mentioned below - esp8266-esp12e-specs.pdf.

> If the data sheet of ESP-12E differs to much from ESP-12F it might not 
> be a good idea to have it as a reference, have you contacted 
> ai-thinker and asked them for a url to the ESP-12E ?
>
Now I have.

> Here is a link
> |http://www.kloppenborg.net/images/blog/esp8266/esp8266-esp12e-specs.pdf|
> Which probably is your device,
>
It is, but the location of this PDF is somewhat "unofficial" - that's 
why I used the one for ESP-12F.

> One idea could be that you download this pdf
> create directory |kicad-library/template/ai_thinker_ESP-12E|
> And put the url there, and describe where the url is in the datasheet 
> field
>
That I don't understand - why put an URL in templates directory and 
point to it in the symbols field?

Misca1234:
I mean save the pdf file in the directory  `kicad-library/template/ai_thinker_ESP-12E`
(just because the pdf is hosted by a private person)
and the point to this directory on the datasheet line

SchrodingersGat:
If you are going to store the documentation locally, then I would do two things:

1. Store it under a new directory library/docs
2. Ensure that KiCad can open the documentation file when the path is specified as library/docs

(e.g. right-click on the symbol in the schematic and select "open documentation". If KiCad doesn't know how to access the path information then we cannot go with this route until such a feature is available)

piotr-wiszowaty:
Doesn't work - KiCad cannot find the file (also tried prefixing with a 
path variable).

The manufacturer said that:
1. ESP-12F is compatible with ESP-12E,
2. Pins 9-14 are for internal use only,
3. One should not use ESP-12E anyway, because it's not certified.

SchrodingersGat:
In that case we should leave the documentation link as it is. Do you have a footprint for this device?

piotr-wiszowaty:
Yes, PR #27 in RF_Modules.pretty.

SchrodingersGat:
Ok then please update the footprint to match the form `RF_Modules:ESP-12E`

piotr-wiszowaty:
Updated.

SchrodingersGat:
Thanks for your contribution

PR #1745, opened 2017-10-19
Zilog cleanup

Misca1234:
Got an hour over and cleaned up the Zilog.lib, it contains two hopeless outdated symbols
But the lib do not have any KLC errors anymore

Z80CPU
Removed double inversion

Z8530
Removed double inversion
Added datasheet URL


SchrodingersGat:
Great work :)

PR #1749, opened 2017-10-21
Add MEC 5G series of tactile pushbutton switches

MatthewStephenson:
These are nice (but expensive) PCB-mount tactile switches with optional LED illumination, single or bi-colour.

Datasheet: http://www.apem.com/int/index.php?controller=attachment&id_attachment=488

image

I wanted to use the generic symbols for switch and illuminated switch, but the pin numbering is different, so I opted to make manufacturer-specific symbols - especially as there wasn't a bi-colour illuminated switch symbol. So I've added 3 symbols: Non-illuminated, 1 LED, and 2 LEDs.

Non-illuminated:

image

One LED, based on the existing illuminated switch symbol:

image

Two LEDs (bi-colour):

image

SMD footprint PR: https://github.com/KiCad/Buttons_Switches_SMD.pretty/pull/27
THT footprint PR: https://github.com/KiCad/Buttons_Switches_THT.pretty/pull/26

As always, comments and suggestions welcome :)


SchrodingersGat:
@MatthewStephenson symbols look great. Please address the issues raised for the footprints and then we can merge :)

MatthewStephenson:
@SchrodingersGat Symbol issues fixed now, I think. Let me know if anything else needs changing.

SchrodingersGat:
Now that the footprints are merged you will also need to slightly adjust the name of the footprint association, so that the symbols map to the correct footprints

MatthewStephenson:
Actually I think they're ok - the current footprint filter for all three symbols is "SW*MEC*5G*" which still matches both SMD and THT footprints after the name change.

SchrodingersGat:
Ah yes my mistake. Thanks for your work :)

PR #1752, opened 2017-10-21
Add Johansen Chip Antenna Symbol (+ Antennas Library)

MatthewStephenson:
I'm not at all sure how to do this...

I want to add these ceramic chip antennas to the library:

https://www.johansontechnology.com/antennas

image

Example Datasheet: https://www.johansontechnology.com/datasheets/antennas/0868AT43A0020.pdf

I couldn't find a library they would fit nicely into - the closest was rfcom.lib but that's all RF modules. So I've made a new Antenna.lib and created a symbol in it (which will fit all of the antennas listed).

image

Is a new library the right thing to do? If it is, I think there will need to be a new Antennas.pretty footprint repository, and I've no idea how (or whether) I can make that.

Help?


Misca1234:
Interesting, then the foot print library perhaps need two foot prints, one with the recommended antenna in copper layer and one without it.

Misca1234:
By the way, are you going to make foot print and 3D models as well ?

MatthewStephenson:
I like the idea of two footprints, one including the recommended PCB trace

I'm working on the footprint now, but as I say, I think it will require a new repository which I don't know how (or if) I can create.

I haven't got the tools to do the 3D models, so feel free to pick that up if you like.

jkriege2:
We already have 4 generic antenna symbols (in device.lib where we collect all/most generic symbols) ... Would one of these fit your antenna (maybe with a modification? You box with the sticking out antenna seems a bit non-standard?

2017-10-22 12_52_51-library browser d__kicad_kicad-library_library_device lib
2017-10-22 12_52_43-library browser d__kicad_kicad-library_library_device lib

2017-10-22 12_52_24-library browser d__kicad_kicad-library_library_device lib

MatthewStephenson:
Hmm. Well, it has two terminals, so it would have to be one of the second two. It's not really a dipole as only one terminal is connected. And it's not really shielded either.

I think of the three, the best fit is Antenna_Shield, which I could modify to Antenna_Chip, and change the second terminal's name to something like PCB_Trace?

Where should the footprint(s) go?

MatthewStephenson:
Updated as suggested, modifying the Antenna_Shield in device.lib to add Antenna_Chip:

image

There don't seem to be any footprint filters for the antennas in device.lib, and I could only find one antenna footprint in a manufacturer-specific library. Should the footprint go in RF_Modules.pretty? In Oddities?

jkriege2:
Are the two pins connected? If not, maybe find another solution than the horizontal wire in your symbol???

The footprint should go to https://github.com/KiCad/Antennas.pretty

JAN

MatthewStephenson:
A fair point. Measuring with a DC multimeter, yes, the pins appear connected. However, after a bit of research, these antennas work as an approximation to a ceramic resonator, so I've drawn a resonator symbol between the two pins.

The CI check failures are consistent with the other antenna symbols in device.lib

Thanks for your advice.

jkriege2:
nice work!

Thanks,
JAN

PR #1753, opened 2017-10-21
Added Arduino Pro Micro Template

heitorPB:
Hello,

I need to make several boards using the Arduino Pro Micro from Sparkfun, so I decided to share a template with the community. I hope It is good enough for the other users, but if there is something wrong or something to improve, please say so :)

Heitor.


SchrodingersGat:

PR #1755, opened 2017-10-22
Added FFC/FPC symbols

jhol:
This patch adds symbols for single and dual-row style FFC
connectors.

All of the symbols have a pair of flange pads F1 and F2. For FFC
connector types that do not have flanges, the Conn_01xXX symbols
are sufficient.

The list of pin-counts was compiled from a simple survey of
products listed by Mouser.


Misca1234:
I guess you are adding a serie of a connectors

Can you post an url of the datasheet ?

I am pretty sure the name FFC_Dual_Row_XX is not sufficient.

The foot print and the 3D models for connectors are very suited to be created with (python) scripts.
I can show you how to create them with python scripts if you want.

poeschlr:
The datasheet given in the symbol seems to be the one for DIN connectors.

Have a look at the generic connector conn_??x??_* symbols. make sure your connectors follow the same visual style.

---

Can you explain what the pins F1 and F2 are used for?

- If they are for mechanical stability -> pins are not needed on the schematic. (Have a look at molex picoblade connector footprints to see how it is done.) Would mean the normal connector symbols are already sufficient for this type of connector.
- If they are for connecting cable shielding -> maybe give them a better name that shows this.
- If they are normal connector general purpose pins -> why would they need a special naming?

---

Do you plan on adding the footprints as well? (If you already have an open pull request please link it.)

jhol:
Hi, thanks for your comments. The context of this is that I'd like to add support for the Molex 502598-xx93 series to the library.

http://www.molex.com/molex/products/datasheet.jsp?part=active/5025985193_FFC_FPC_CONNECTORS.xml

I have created footprints and 3d models for the entire series robotically with python scripts, and I planned to submit MRs for these later once the schematic symbols had been included.

I thought about adding the schematic symbols to a new "Molex" symbol library, but it turns out that the schematic symbols for FFC connectors are quite generic. Many of these connectors have multiple suppliers which have the same connections, but different footprints.

Based on @poeschlr 's comment, it sounds like I was wrong to include the flange pins - these are for mechnical stability. If these are removed, the single-row FFC symbols are redundant - I will remove them from my MR.

The dual row FFC symbols are worth having I think, because they have an odd number of pins e.g. I want to add a symbol for the 51-pin FFC connector. For this it's helpful to have a symbol 26-pins on the left side, and 25-pins on the right.

One question... if I proceed with the dual-row symbols, should they be named `FFC_DUAL_ROW_*`? Or should we have a more generic title to cover other dual-row odd-pin-numbered connectors, that are not for FFC? If so, what should that title be?

The symbols were generated programatically, so it's no problem for me to make visual changes to match `conn_??x??_*` symbols.

jhol:
I updated the branch with a few changes...
- Removed all the single row FFC symbols. They are redundant when the `conn_??x??_*` symbols are available.
- Removed flange pads.
- Updates style to match `conn_??x??_* `
- Added 3D models for the Molex 502250 series.
- The PR for the footprints is here: https://github.com/KiCad/Connectors_Molex.pretty/pull/56

poeschlr:
Did you create the 3d models your self or are they downloaded somewhere?
If you created them your self: Did you create it in a program capable of exporting step? Example Freecad.

A generic name for these symbols would be nice but to be honest i don't have a good idea what they should be called. 
Maybe something like 
- `Conn_Dual_Row_Odd_Pin_Count_[num pins]` (Quite long)
- `Conn_Dual_Row_[num pins]` (I think this would be understandable.)
- `Conn_FFC_Dual_Row_[num pins]` (If we want to use FFC in there.)
@SchrodingersGat, @jkriege2 or @evanshultz any ideas from your side?

As description i would suggest something similar to the `Conn_02x*_Odd_Even` symbols.
`Generic connector, double row, [num pins] pins (row 2 has one pin less than row 1), odd/even pin numbering scheme (row 1 odd numbers, row 2 even numbers)`

For the footprint filter i would use: `Conn*:*_2x*` 
Filters all dual row connectors with the new naming convention. 
The pin numbers are filtered by a separate filter option. So don't include it in the filter.

If we want them to work in kicad v4 we might also want to add `*FCC*_2x*` as a second filter. (would only filter your connectors.)

jhol:
@poeschlr 

* I created the 3d shapes using a parameterized openscad script, and a bash script to export the STL files of 3 different layers of each connector variant. I then combined the STLs manually in wings, and exported as VRML.
* Given that the symbols were generated by script also, it's no problem to update the name - if we can get a consensus on that.
* Once the name is agreed, I'll to update the footprint list as you suggest.

Thanks for your review.

poeschlr:
> I created the 3d shapes using a parameterized openscad script, and a bash script to export the STL files of 3 different layers of each connector variant. I then combined the STLs manually in wings, and exported as VRML.

The currently suggested workflow is to either use FreeCad + [Kicad-Stepup](https://sourceforge.net/projects/kicadstepup/) or the [CadQuery scripts](https://github.com/easyw/kicad-3d-models-in-freecad/tree/master/cadquery/FCAD_script_generator) by @easyw (Both workflows allow for the creation of step and wrl models. The new library requires both of them for MCAD collaboration.)

I think there should be a way to get openscad to export step. (It is also possible to run openscad scripts from within freecad. Then freecad should be usable to export both step and wrl with the help of the kicad-stepup tools workbench i linked above.)

If these scripts are standalone than we should also consider adding them to the packages3d-source repo.

jhol:
@poeschlr I made the 3D models with this workflow: http://happyrobotlabs.com/posts/tutorials/tutorial-3d-kicad-parts-using-openscad-and-wings3d/

This is rather convoluted, but I found it worked pretty well. I only wish there was a way to combine the STL layers on the command line - then I could do the OpenSCAD to WRL generation in a bash script.

jhol:
@poeschlr I just updated the symbols with the changes you suggested, and I took a decision on the naming.

The only outstanding issues is that I'm getting a few Travis warnings about the footprint list...
```
Checking symbol 'Conn_Dual_Row_17':

    Minuses and underscores in footprint filter '*FCC*_2Rows*' should be escaped with '?' or '*'.
```

I could make the change, but wouldn't it be a problem to have `*2Rows`? That would match `_12Rows` for example.

poeschlr:
The wings workflow is outdated. The new lib needs step files for mechanical integration. This can not be done using stl/wrl and wings! (We sadly can only accept such contributions to this legacy repo not to the new shiny kicad-packages3d repo that will be used for kicad v5)

If you could share your script (simply post it as and attachment to a comment) i could try it in freecad and post a short tutorial on how to do get step and wrl from within freecad. (Freecad can merge multiple bodies as well. And with kicad stepup it does produce rather nice models in both step and wrl format.)

---

About the footprint filter: Yes please include the underline. We can ignore the travis error in this case.

jhol:
@poeschlr That's annoying with regard to the OpenSCAD thing. If you want to cherry-pick the symbols from the pull request please go ahead. Otherwise, I can resubmit without the 3D models later.

With the 3D models, I'm happy using FreeCAD to make the files - though if you could do it super-quick that would help. I can submit my OpenSCAD model if that would help.

I just watched this video, which shows how to import a Molex STEP file: https://www.youtube.com/watch?v=O6vr8QFnYGw&feature=youtu.be . What about the licensing of this Molex file? I presumed it would be proprietary, which is why I made my own simplified model from the schematic drawing.

I suppose the best way would be to recreate my OpenSCAD model in FreeCAD python script. I want to be able to generate the different variants of the connector programatically.

jhol:
@poeschlr just re-pushed the branch with the symbols for the Molex 502250. Are these good enough to merge in?

Also, do you have any thoughts about the above comments?

I'm happy to rebuild the models in FreeCAD. You linked a video that showed how to import them from Molex STEP files - but would this not be a copyright violation?

I'm happy to create my own simple 3D models, but I do need a method of scripting the Part Designer. Doing it manually for the whole series seems laborious and easy to mess up. Is there any information about how to do that?

easyw:
@jhol 
for scripting 3D models you can have a look at:
https://github.com/easyw/kicad-3d-models-in-freecad

poeschlr:
We are already in the process of moving stuff over to the new v5 repo. Because the moving of Connector symbols already started we can not accept any further pull requests for this lib in the legacy repo.
Please re submit this as soon as https://github.com/KiCad/kicad-symbols/pull/26 is merged.

Your symbols should then go into Connector_Generic
As a better naming for them i would suggest `Conn_2Rows-nnPins`, Where nn is the number of pins (two digits with leading zero) 
This would fit the footprint naming for such components.

SchrodingersGat:

PR #1757, opened 2017-10-22
Microchip: SY89312V

calebns:
3.3/5V, ECL/PECL 4GHz x2 clock divider

image

Documentation:
http://ww1.microchip.com/downloads/en/DeviceDoc/sy89312v.pdf


Misca1234:
Hi

Some common comments

Consider to use "fill foreground colour" for the box and that the line of th box is non zero, for example 0.254 mm thick ?

You could use 2.54 mm length for the pins to avoid to have so long

Change the description from
3.3/5V, 4GHz ECL/PECL x2 divider
to
3.3/5V, 4GHz ECL/PECL x2 divider**, QFN-8**

calebns:
Fixed.

jkriege2:
Hi!

nice work ... could you please reduce the pin-length to 100mil and set the pin text offset to 20mil:
2017-11-06 21_08_34-part library editor -- d__kicad_kicad-library-review_library_microchip lib

Thanks,
JAN

SchrodingersGat:
@calebns can you make the suggested changes? :)

calebns:
Sorry, completely forgot about this. Should be fixed now. Thanks!

jkriege2:
Nice work!

Thanks,
JAN

PR #1760, opened 2017-10-22
Fix naming for Schurter FUP 3d models.

poeschlr:
I had a typo in the company name (Schruter instead of Schurter)
I now use the new naming scheme.

This footprint is valid for two different parts. One part is valid for a 5x20mm fuse the other for the 6.3x32mm fuse. This makes the naming and description quite hard to understand. (Should we duplicate it for each fuse type?)

Footprint renaming:
https://github.com/KiCad/Fuse_Holders_and_Fuses.pretty/pull/19

Other 3d part renaming pull requests:
packages3d repo: https://github.com/KiCad/kicad-packages3D/pull/149
packages3d-source repo: https://github.com/KiCad/kicad-packages3D-source/pull/59


SchrodingersGat:
I'll merge this one but we should not duplicate 3D models to this repository any more, as we need to start the transition :)

PR #1764, opened 2017-10-23
Copy connector shape from USB_OTG to USB_B_Micro.

ppelleti:
I recently [added the USB_B_Micro symbol](https://github.com/KiCad/kicad-library/pull/1724) (and an alias, USB_B_Mini).  I hadn't noticed that there was already a similar USB_OTG symbol.

However, the two need to remain separate, because USB_OTG has VBUS and GND marked as power inputs, while USB_B_Micro has them as power outputs.

This commit makes USB_B_Micro identical to USB_OTG, except for the direction of the power pins.  Specifically, it copies the connector shape from USB_OTG to USB_B_Micro.  (I had been too lazy to draw the connector shape when I created USB_B_Micro.)

usb_b_micro


SchrodingersGat:
Thanks @ppelleti

PR #1770, opened 2017-10-24
A version of the ATmega328P-PU symbol with Arduino pin names.

ppelleti:
I'm not entirely sure this even belongs in the KiCad library, so please let me know if it doesn't.

This is a symbol for the [ATmega328P-PU](http://www.atmel.com/images/Atmel-8271-8-bit-AVR-Microcontroller-ATmega48A-48PA-88A-88PA-168A-168PA-328-328P_datasheet_Complete.pdf) microcontroller, but instead of using the manufacturer's pin names, it uses the Arduino pin names.  This makes it easier when designing a circuit whose software is meant to be programmed with Arduino.

The Arduino pinout is [here](https://www.arduino.cc/en/Hacking/PinMapping168), although most of the exact pin names came from [this sticker](https://cdn-shop.adafruit.com/970x728/554-04.jpg).

atmega328p-pu_arduinopinout


SchrodingersGat:

PR #1775, opened 2017-10-28
Added 3D models for dip switches

Misca1234:
Push of 3D models for dip switches
The models is made with python scripts

3D models in kicad-library
https://github.com/KiCad/kicad-library/pull/1775

Modifications of foot prints
https://github.com/KiCad/Buttons_Switches_SMD.pretty/pull/28

3D models in 3D-packages
https://github.com/KiCad/kicad-packages3D/pull/151

3D python script
https://github.com/easyw/kicad-3d-models-in-freecad/pull/116

The following SMB dip switches have 3D models

SW_DIP_x1_W5.08mm_Slide_Copal_CHS-A
SW_DIP_x1_W7.62mm_Slide_Copal_CHS-B
SW_DIP_x1_W8.61mm_Slide_LowProfile
_SPST_EVQPE1
_SPST_REED_CT05-XXXX-G1
_SPST_REED_CT05-XXXX-J1
_SPST_REED_CT10-XXXX-G1
_SPST_REED_CT10-XXXX-G2
_SPST_REED_CT10-XXXX-G4
SW_DIP_x2_W5.08mm_Slide_Copal_CHS-A
SW_DIP_x2_W6.15mm_Slide_Omron_A6H
SW_DIP_x2_W7.62mm_Slide_Copal_CHS-B
SW_DIP_x2_W8.61mm_Slide_LowProfile
SW_DIP_x3_W8.61mm_Slide_LowProfile
SW_DIP_x4_W5.08mm_Slide_Copal_CHS-A
SW_DIP_x4_W6.15mm_Slide_Omron_A6H
SW_DIP_x4_W7.62mm_Slide_Copal_CHS-B
SW_DIP_x4_W8.61mm_Slide_LowProfile
SW_DIP_x5_W8.61mm_Slide_LowProfile
SW_DIP_x6_W5.08mm_Slide_Copal_CHS-A
SW_DIP_x6_W6.15mm_Slide_Omron_A6H
SW_DIP_x6_W7.62mm_Slide_Copal_CHS-B
SW_DIP_x6_W8.61mm_Slide_LowProfile
SW_DIP_x7_W8.61mm_Slide_LowProfile
SW_DIP_x8_W5.08mm_Slide_Copal_CHS-A
SW_DIP_x8_W5.9mm_Slide_Copal_CVS
SW_DIP_x8_W6.15mm_Slide_Omron_A6H
SW_DIP_x8_W7.62mm_Slide_Copal_CHS-B
SW_DIP_x8_W8.61mm_Slide_LowProfile
SW_DIP_x9_W8.61mm_Slide_LowProfile
SW_DIP_x10_W5.08mm_Slide_Copal_CHS-A
SW_DIP_x10_W6.15mm_Slide_Omron_A6H
SW_DIP_x10_W7.62mm_Slide_Copal_CHS-B
SW_DIP_x10_W8.61mm_Slide_LowProfile
SW_DIP_x11_W8.61mm_Slide_LowProfile
SW_DIP_x12_W8.61mm_Slide_LowProfile

Example of models

bild

bild

bild

bild

bild


poeschlr:
Your pull request adds SW_DIP_x8_W5.9mm_Slide_Copal_CVS but none of the other switches of this series. (This one model does not really look like it is the correct model for this footprint. It is missing 4 connection points at the corners. It does look more like Cobal CHS.)

The x1, x10, x11 and x12 switch versions have much thinner leads than the x2x,3,x4,...,x9 switch versions. I think there might be a mistake in your scripts because all variants are effected. (In the screenshot all switches are from the Cobal CHS variant)
screenshot from 2017-10-28 21-34-34

Misca1234:
I deliberately avoided to make those with corner pads, so if some of them are included they should not have been included but I can try to find the data sheets for them as well.

The width of the leads is "manually" defined in the script, do you want the leads as wide as the pads on all 3D models ?

poeschlr:
I only noticed that the x1 version looks different to the x2 version for all switch series.
I don't know what the correct width is. If all of them would have the same width i would not have noticed any problem. (I did not check against a datasheet. For 3d models i only make a coarse optical check if it could be correct.)

---

In your script it might be a good idea to have some parameters that are in a data structure dedicated per series. (would reduce the chances of something like this happening.)

Example a second dict called series_param holding another namedtuple. 
This namedtuple would hold information such as padsw, pitch, ... (everything that is the same for all switches of the same series.)
In the make_params function you then simply add one parameter called series_params instead of it expecting the parameters them selves.
In it you would then simply write `padsw=series_params.padsw`

And instead of writing: (This is symbolic. representing line 184 of your current script)
`"SW_DIP_x1_W5.08mm_Slide_Copal_CHS-A"    : make_params(L, W, H, pitch, paddist, PH, npins, padsw, pinshape, modelName, rotation, type, corner),`
you would then write:
`"SW_DIP_x1_W5.08mm_Slide_Copal_CHS-A"    : make_params(L, W, H, npins, modelName, rotation, type, corner, series_params["Slide_Cobal_CHS-A"]),`

Maybe rotation, type and corner could also be part of the series parameters. (I can't tell it from the name alone.)

Doing it this way makes it a lot easier to maintain such a script. (Just assume you later find you want to change the pin width of a specific series of switches. Currently you need to go through all the definitions and hope you found all instances of this parameter.)

---

Another tip:
If you look at my molex scripts you will find a directory called cq_models in it. All the scripts in there can be directly loaded in the freecad cadquerry workbench. (makes development a lot easier.)
The generation via script is then done in a separate script.
(I also separated the parameters but i would advice against this. There is a bug in the workbench that makes it harder that way.)

To find out what you did i changed your script such that it can run in the workbench. 
So if you want to split the python part from the cadquery part half the work is already done.
This would only require you do delete everything except the main function plus the 3d model generator from your current file. (You also still need the imports and the stuff about licensing, ...)

In the make 3d model function you would then call the generate_part function instead of calling all 3 functions them selves.
You also would need to add 
sys.path.append("cq_models") # assuming you add the generator in the subdirectory called cq_models.
from dip_switches import *

[dip_switches.py.txt](https://github.com/KiCad/kicad-library/files/1426701/dip_switches.py.txt)

Misca1234:
The script was originally from DIP 3D models THT variant if i do not recall wrong and all the stuff in end of the main loop of replacing colours etc. I had no clue why and how so I sticked to the structure.

But let me see if I can change the script in line with your comments

Misca1234:
I am not that fond of changing the structure atm, both this script and the script comming for the ACDC_DCDC converters look the same.

I will look into your molex script for the next batch of 3D models (after the ACDC_DCDC) and then I can modify this and the ACDC_DCDC when i get it working according to your molex script.

If it is ok with you?

I am changing the with of the pads/leds for this script now so they don't look so "thin"

Misca1234:
I noticed in the 
https://www.nidec-copal-electronics.com/e/catalog/switch/cvs.pdf

that the one with "pads in the corners" are metal shielded, these requires different 3D models than these simpler which should be taken into account at an update of the script in the future

Misca1234:
The pad sizes have been changed to be equal or slightly smaller than the foot print pad size

sandhan26:
@Misca1234 i need help with designing some 3d models for some kicad components i have in mind. For the time being, i want to do it manually in FreeCAD, without the cadquery scripts.
I am familiar with using the draft tool and Part design workbenches in FreeCAD.
I need help in the sense, that when designing a 3d model in FreeCAD, which will be used in Kicad, what extra needs to be done.

Misca1234:
@sandhan26 

It is fairly easy, what you need is the stepup macro to kicad, you use it like this

1. Download and install stepup macro, http://sourceforge.net/projects/kicadstepup/
2. Design the component as it is in real life, do not scale it, just use real life cordinates
3. Use any colours you want to distinguish different parts
4. When done, concatenate all the components with this macro button

bild

5. Select the concatenated object and click on this button

bild

6. You can answer the first questions with OK
7. Then you get a number of questions of matching your chosen colours against the official KiCad's official colours
8. The macro will then create a .step and .wrl file with right scaling, object colours and other paramters.

NOTE:
Remember to save to the .Fcstd file before you do do step 4

poeschlr:
@Misca1234 thanks for these models. They look great.

And thanks for this in dept explanation. We might really want to add this stuff to the webside as part of the library contribution guide. (I think you asked in another pull request about this but i can't find it any more.)

The webside is made using hugo v0.3 and the pages are mostly written in a markup language similar to the one used on github. As i have helped @SchrodingersGat with the new KLC stuff i know a bit about how this stuff works so you can always ask if you need help.

Misca1234:
Thx, yes we discussed it in another thread and I volunteered to write some html pages, let me check out hugo v3

We need essential these guides with info about how to fork, push and make updates
Tips of how to do the push (include link to data sheet and screen shot in the push)

1. Creating a symbol 
2. Create a foot print
3. Create a 3D model from scratch, with info about stepup macro 
4. Create a 3D model from python scripts scratch, this should not explain python nor the existing scripts functions, just the benefits of using scripting for example connectors.

Even if 4 is somewhat little information, the link can be provided when someone ask and and they can get a starting point to work from.

sandhan26:
@Misca1234 thanks for the in depth explanation. I have few confusions, which i would like to clear.

This the 3d model for Bosch BME280/680  LGA-8_2.5x2.5mm package.

* How do i add pads in this? Will the pads be simply very thin body objects?
* Is there any way to make the top hole more darker? 
screenshot from 2017-11-03 07-28-02

screenshot from 2017-11-03 07-25-52

Misca1234:
@sandhan26 

> How do i add pads in this? Will the pads be simply very thin body objects?
Yes

> Is there any way to make the top hole more darker?
I usually put a hollow cylinder inside the hole with a different colour, let say blue
then, when the stepup macro asks for colour transformantion you choose to conver the blue to dark grey or something.

> Is there any color convention for kicad 3d models?
Not what I am aware of (I guess someone would argue "keep it as natural as possible")
Also, there is not that many colours to play around with either.

In the end it is up to us to make an appealing object.
If you want to pimp up your model with false colours, just go for it.

easyw:
https://cld.pt/dl/download/64e39e99-c5b6-451b-accd-9e25331ceafe/KiCad_3D-Viewer_component-materials-reference-list_MarioLuzeiro.pdf?download=true
This is a doc for material properties (it is in the README file of packages3d)

PR #1777, opened 2017-10-29
Add MKL26Z128VFT4 with Teensy-LC pin naming

MatthewStephenson:
I'm not sure that this symbol will be wanted in the library, but I've made it for my own use so I figured I'd offer it up, just in case.

This is a modified version of the NXP Kinetis MKL26Z128VFT4 MCU, with the pins renamed and reordered to reflect their use on the [Teensy-LC](https://www.pjrc.com/teensy/teensyLC.html) arduino-alternative. Its intended use is for making a custom board using the Teensy-LC's software, libraries, bootloader, etc. This symbol makes it less error-prone to connect the correct pins.

I've kept the Kinetis's pin names, but prefixed them with "Txx" for "Teensy", e.g. T11/PTC6 is pin 11 on the Teensy, and PTC6 on the Kinetis.

image

It is a bit of an oddity, so I won't be offended if it's not wanted in the library!


poeschlr:
The MCU_NXP libs have already been transferred over to the new repo for the v5 release.
If you are still interested please re submit your symbol(s) over at the new repo https://github.com/KiCad/kicad-symbols.

SchrodingersGat:

PR #1778, opened 2017-10-29
Add TDA1308

adamheinrich:
Add schematic library for the class-AB stereo headphone driver by NXP (https://www.nxp.com/docs/en/data-sheet/TDA1308.pdf):

tda1308-mod

Pin 8 in the library (`V+`) has a different name than pin 8 in the datasheet (`VDD`). This is to make the symbol look similar to others. Otherwise, I would have to shift pin 4 or 8 horizontally (because their labels would not fit).


jkriege2:
nice work!
Thanks,
JAN

PR #1779, opened 2017-10-30
Add symbol for D1 mini by WEMOS

Rantanen:
## D1 mini by WEMOS

https://wiki.wemos.cc/products:d1:d1_mini

This was a bit more straight forward compared to the previous module I PR'd, the Maple Mini. Mainly because I had better understanding of what would be preferable. On the other hand, I kind of breezed through the symbol this time, so it's more likely there are silly mistakes that I didn't catch. :)

The major KLC difference is the positioning of the pins. As most pins are bidirectional, there's no clear input/output sides. Probably more importantly the only clear "input" pin is positioned on the right sides in slight violation of KLC. I felt this was justifiable as it's part of the TX/RX pair of pins and they are usually wired together.

The one thing I had doubts was the footprint filter (not pictured). It's specified as: `D1*mini*`, which I adapted from the previous ones - however in this case that might be a bit generic having two very short and rather generic identifiers in it

While I could have named this "WEMOS D1 mini", I preferred just "D1 mini" thanks to the amount of clones and copies in circulation.

Another thing worth noting is the orientation of the module. While it only affects the pin numbering in the symbol, it's more significant in the eventual footprint. The pins on this symbol are numbered counter clockwise (as tradition) but in such a way that the USB connector is at the bottom of the module - this differs from modules such as the Maple Mini where the "pin 1 top left" orientation has the USB connector at the top.

The "USB connector at the bottom" orientation is justified by the silk screen texts, as seen [here](https://wiki.wemos.cc/_media/products:d1:d1_mini_v2.3.0_1_16x9.jpg), which are oriented with the connector at the bottom.

symbol
fields
props


SchrodingersGat:

PR #1780, opened 2017-10-30
pspice: Add MOSFET symbols

orsonmmz:


SchrodingersGat:
Hey Orson can you address the issues raised by the KLC / Travis scripts?

orsonmmz:
It should be better now. There are no footprint or documentation fields assigned, because these are generic MOSFET symbols. The actual type of transistor is selected by associating a Spice model.

SchrodingersGat:
Thanks Orson

PR #1783, opened 2017-10-30
Add TPS22929D power distribution switch

aurel32:
KLC complains about "Rule 4.5 - Pin orientation", however I believe that a power distribution chip should be like a voltage regulator with power input on the left and power output on the right.


SchrodingersGat:
Hi @aurel32 thanks for the contribution - unfortunately there has been a more recent touch of that file, can you please fix the conflicts?

aurel32:
Hi @SchrodingersGat. I have just updated my branch to fix the conflict.

SchrodingersGat:
Thanks!

PR #1789, opened 2017-11-02
Move opto gate drivers to driver_mosfet

El3ctr0nic:
Some tweaks and unification on symbols move.


El3ctr0nic:
opto_move

Ok some Jenskins fail so i must repair it :) Be patient :)

El3ctr0nic:
I have some issue with new KLC. Power pins should be placed top/bottom of the symbol.
It make some complications in dual isolated gate drivers. Connection will be totally unlogical (personal opinion). I think best way is setup power pins as OC and OE to pass Jenkins.
Dual driver example below. Making splitted symbol isn't good idea too. In drivers logic and physic connection is very close (3-phase bridges).

acpl314

 
I need some opinion to continue my work :)
Regards Marcin

El3ctr0nic:
I resized LTV temporary to pass Jenskins (i think it should be splitted symbol).
I made smode cheat to in drivers. Shared output transistors and uses underscore before pin names to pass.
Links for doc updated to Broadcom site.
Pls feedback :) 
Regards Marcin

SchrodingersGat:
@El3ctr0nic thanks for the contribution, sorry that it has not had a response yet.

Firstly, please note that the KLC is not always 100% right and there are exceptions. Please do not "cheat" by renaming the pins. In this case I think that the way that you have drawn these symbols is correct, and the warnings generated by KLC can be ignored.

Also, note that warnings are (generally speaking) associated with rules that say **should** instead of **must** (that's the idea, anyway).

So, please revert the pin names to their original form :)

SchrodingersGat:

PR #1790, opened 2017-11-02
Add X cap discharge ICs

evanshultz:
Datasheets:
- https://www.nxp.com/docs/en/data-sheet/TEA1708T.pdf
- http://www.monolithicpower.com/DesktopModules/DocumentManage/API/Document/getDocument?id=142

I'm using the smaller pin name style discussed at https://github.com/KiCad/kicad-library/pull/1611#issuecomment-331274382. In this case, there are lines inside the part body so the pin text can't easily fit inside the part body. This may need discussion, and certainly an update to KLC.

Image:
image


jkriege2:
nice work,
Thanks,
JAN

evanshultz:
@jkriege2 Thanks!

@SchrodingersGat Can you update KLC to cover these pin name sizes and placements? I don't want things to get out of hand before we've controlled what is acceptable.

PR #1797, opened 2017-11-04
Added HX712 ADC

tk5ep:
Added the AVIAC HX712 24 Bit single power 
[Datasheet](https://www.scribd.com/doc/94396103/Hx712-English-Datasheet)


SchrodingersGat:
Hi @tk5ep the pins are very long on these, can you please reduce pin length from `200mil` to `100mil` :)

tk5ep:
Done.. HX711 and HX712 pin length are now 0.1.

SchrodingersGat:
Thanks! The other issue I am only now noticing is that you have put the footprint data into the footprint filter!

The footprint information should go here:

selection_021

And the footprint filter should be set like:

selection_022

Thanks!

SchrodingersGat:
Sorry - still some Travis errors - https://travis-ci.org/KiCad/kicad-library/builds/305911862?utm_source=github_status&utm_medium=notification

tk5ep:
Hi,
Hard work... 
Before i, correct.. What's wrong in the HX711 footprint name and illegal character detected ?
Is it the dot in 1.27mm ?
BTW i don't know if it is necessary to put a footprint name in the description.. There are different footprint possible and as recommended in KLC, it should be left blank.

SchrodingersGat:
Issue is that you have wildcard * character in the footprint field. This must point to a single explicit footprint. The filtering is only for the footprint filter fields!

SchrodingersGat:

PR #1803, opened 2017-11-06
Added the Analog Deivces ADP230xARDZ

johnfwhitmore:
I've re-added this component, after previous fail, but looking at this now and this is another failed pull request. I'll make the request so that I have somewhere to have a conversation about the problems I'm seeing. Firstly I can't attach a screen shot to this pull request - "Something went really wrong, and we can't process that file." It's a PNG file? I can't see any mention of what is expected of a screen shot?

In addition to that problem I've looked at the diff file. I've added a new component based on an existing component. I'd expect all the changes to the dc-dc.lib file to be my additions but kicad seems to have edited numerous parts of this file. There are lots of changes according to diff. I've no idea what is going on but I didn't expect this many changes to the lib file. Should I be editing the library by editing the actual text file, in a text editor, to make sure I only add the component I'm working on and not use the CAD package? There seem to be a lot of lines removed from the lib file. I've removed nothing!

So you're going to come back and say that I've made changes to numerous symbols and I'm just going to have to forget this process. It does not appear to work.


johnfwhitmore:
screenshot from 2017-11-06 10-17-07

johnfwhitmore:
OK One problem is resolved as I managed to attach the png file.

The other problem is with numerous changes to lib file

lavadew:
Looks like a big mess.. but it's not you fault. 
I also had something like this.
What you can do:
Take the text file of the dc-dc lib form the master branch and copy+paste this in to your local dc-dc files, so you have a working library again. Then take the section of your ADP230(...) symbol from the text file here and append it to your local dc-dc lib files..
Then commit again and see if it builds..
I hope my way is understandable.

johnfwhitmore:
If I understand you correctly I should NOT be using kicad to create new components but rather be using emacs, or a similar text editor to create the new components in text files.

johnfwhitmore:
In addition the Continuous Integration is not failing because of the component I've added to the file. I've tested the component I added before I issued a pull request:
kicad-library-utils/schlib$ ./checklib.py -c ADP230xARDZ ../../kicad-library/library/dc-dc.lib 
Checking symbol 'ADP230xARDZ' - No errors

I have not corrected pre-existing errors in the dc-dc.lib file

johnfwhitmore:
So if I'm going down the text file editing route can I simply rebase my fork to the upstream master with:
$ git rebase upstream/master
$ git push -f origin master
And edit the necessary text files before committing again. I will have to delete this pull request as there are already two commits in it which were the result of graphically adding a component.

poeschlr:
It is not the preferred way to use a text editor. But sometimes it might be required. (especially if you need to merge stuff. Or if you do not use a local git client under windows. Then you need to correct the line endings from CRLF to LF before uploading your files via the github web interface.)

Just make sure that you use a relatively new kicad version. (kicad v4.0.6 or newer) Older versions introduce unwanted whitespace characters. @SchrodingersGat correct me if i am wrong here. (What version of kicad is new enougth such that it does not introduce white space chars.)

Also make sure you open the lib in kicad before your commit it. (This ensures a valid lib. If kicad throws a error message then something went wrong.)

johnfwhitmore:
Application: kicad
Version: (2017-11-03 revision 08a9dc0)-master, debug build
Libraries:
    wxWidgets 3.0.2
    libcurl/7.47.0 GnuTLS/3.4.10 zlib/1.2.8 libidn/1.32 librtmp/2.3
Platform: Linux 4.4.0-97-generic x86_64, 64 bit, Little endian, wxGTK
Build Info:
    wxWidgets: 3.0.2 (wchar_t,wx containers,compatible with 2.8) GTK+ 2.24
    Boost: 1.58.0
    Curl: 7.47.0
    Compiler: GCC 5.4.0 with C++ ABI 1009

Build settings:
    USE_WX_GRAPHICS_CONTEXT=OFF
    USE_WX_OVERLAY=OFF
    KICAD_SCRIPTING=OFF
    KICAD_SCRIPTING_MODULES=OFF
    KICAD_SCRIPTING_WXPYTHON=OFF
    KICAD_SCRIPTING_ACTION_MENU=OFF
    BUILD_GITHUB_PLUGIN=ON
    KICAD_USE_OCE=OFF
    KICAD_SPICE=OFF

johnfwhitmore:
Running on Ubuntu so line ending should not be a problem. The problem is that when I graphically add a new component to the dc-dc.lib file other components are changed by that edit. It appears that the only way to cleanly add a new component is with emacs.

poeschlr:
I would also suggest you do not use your personal master as source for a pull request. create a separate branch for each pull request.

below i assume upstream points to the official repo and origin to your personal fork.
```
git fetch upstream
git checkout upstream/master
git checkout -b descriptive_branch_name
git push --set-upstream origin descriptive_branch_name
```

do your stuff

```
git add -u
git commit -m "descriptive commit message"
git push
```

This way you always start in sync with the current HEAD.
You also only need to merge if github says something about not possible to merge.

PR #1804, opened 2017-11-06
Adjust symbol names to the new KLC

lavadew:
Some symbols with an "/" in their names get their names adjusted to the new KLC.

PR #1805, opened 2017-11-06
Adjust symbol names to the new KLC and adjust the aliases

lavadew:
Removed "/" from some names and adjusted the aliases of the changed symbols as well.


jkriege2:
thanks for this fix. Could you do the same replacements in adc-dac.dcm, otherwise the description for these symbols is not found!

Thanks,
JAN

jkriege2:
... ahhh just saw https://github.com/KiCad/kicad-library/pull/1804 ... could you please combine these two into one PR, then it's easier to check the result!

Thanks,
JAN

lavadew:
PR can be closed, created a new one #1813

PR #1806, opened 2017-11-06
Added the Analog Devices ADP230xARDZ

johnfwhitmore:
Third time's a charm:
Added the Analog Devices ADP230xARDZ using emacs this time, hopefully it's cleaner then using kicad graphically.

screenshot from 2017-11-06 10-17-07


jkriege2:
Hi!

looks good up to two things:
1. You don't need the "bastract base" ADP230xARDZ ... please remove it from the library
2. The VIN pin can be on the left ... this style is adopted for all voltage regulators and switching regulators in this lib ... so we'll ignore that KLC-violation/warning, see e.g. 
2017-11-06 19_02_21-choose component 559 items loaded
2017-11-06 19_01_40-choose component 559 items loaded

Thanks,
JAN

PR #1808, opened 2017-11-07
Added Microchip MCP1825S LDP Regulator

johnfwhitmore:
I'm initiating a pull request to have the Continuous Integration test run on what I've added. There are better error message from it then kicad-libraries-utils/schlib/checklib.py, which is giving me an error "Violating Rule 4.8 - Field text size" but the klc [http://kicad-pcb.org/libraries/klc/] doesn't actually list a rule 4.8?

Unfortunately there are 3, 4 and 5 pin packages for this device. I've only added the 4 pin SOT-223 for the moment, which has two Gnd connections whereas the TO-220 and DDPAK only have one Gnd. I'll add those as a 3 pin component.

mcp1825s


SchrodingersGat:
Sounds like your local copy of the kicad-library-utils scripts is out of date. There were some very big changes made recently.

johnfwhitmore:
Oops not sure when I updated them. Let me pull them... yes you're right :-( that's a list of error messages that matches the KLC

jkriege2:
- Note: These need to be moved to regul.lib as they are LDOs and not switching regulators.
- Also: You can use the SOT223-footprint, which has the large-pin ad pin2, then you only need pins 1,2,3
- finally please make them the same size as the existing LDOs (from the image I suspect the pins are 150mil long .... 100mil is enough!

THanks,
JAN

johnfwhitmore:
More oops, thanks for all that. I'll delete this PR when I move all to LDO. Now I have the latest utils I should be able to correct errors before I re-submit.

PR #1809, opened 2017-11-07
Added the Microchip MCP1825S 500mA LDO Regulator

johnfwhitmore:
This more or less a direct copy of the regul/MCP1826S LDO Regulator. One change has been made to clear a KLC error on the Pins.

The MCP1826 is a 100mA part whereas this MCP1825 is a 500mA Part. 

There is one outstanding error "Violating S3.1" because the part is not centered on the origin?

mcp1825s


jkriege2:
nice work!

Thanks,
JAN

johnfwhitmore:
Cheers Jan, that one was too easy, once you'd pointed out regul.lib, and I'd found the sister part.

PR #1810, opened 2017-11-07
AVR-ISP-10 and AVR-JTAG-10: Replace symbols

matthijskooijman:
This redesigns the AVR-ISP-10 and AVR-JTAG-10 symbols to be compliant with the current KLC. It basically uses the same layout as previously used for AVR-ISP-6 (see #1736).

Just like with AVR-ISP-6, there's a few checklib warnings that seem acceptable to me (I cannot find any official docs about these pinouts to link to, and since the power pins can be either input or output, depending on how the connector is used, I used "passive" for all pins).

selection_002
selection_003

For AVR-JTAG-10, I'm not entirely sure if the VREF pin should be seen as a power pin and be at the top, or if it should be at the side?


jkriege2:
Please move the NC pin as indicated below to make it harder to connect to it by accident
2017-11-07 22_08_36-part library editor -- d__kicad_kicad-library-review_library_atmel lib

jkriege2:
PS: Would you also redo the 2 remaining symbols
- `AVR_PUDI-6'
- `AVR_TPI-6'
That would be great!

Thanks,
JAN

matthijskooijman:
I just pushed an update to this PR. I:
 - Included AVR-TPI-6
 - Included AVR-PUDI-6 (and renamed it to AVR-PDI-6, since I couldn't find any references online to a "PUDI" connector or connection, the interface seems be be called "PDI").
 - Changed the refdes of all connectors (including AVR-ISP-6 from the previous PR) to J? as per KLC.
 - Changed some pins to have a bar above the pin name to mark them as active-low on all connectors (including AVR-ISP-6 from the previous PR) to J? as per KLC.
 - Moved the NC pins inside the box, as requested.

As far as I'm concerned, this PR is ready for merging.

poeschlr:
If these are connectors then they should be moved into a connector lib.
The connector symbols have already been moved over to the new repo. I would suggest you add them to the Connector_Specialized lib in the [kicad-symbols](https://github.com/KiCad/kicad-symbols) repo.

The MCU repos should only hold the micro controllers.

matthijskooijman:
Ah, didn't realize the symbols also moved to a new repo. I'll create a new PR there, and move the symbols to the other lib at the same time.

poeschlr:
I think the connector lib already has some jtag connectors. So some of these might be duplicates of symbols already in there.

matthijskooijman:
@poeschlr, I moved these changes over to https://github.com/KiCad/kicad-symbols/pull/39. Thanks for pointing out the other JTAG symbols. They are not duplicates (different pin numbers between AVR and ARM), but I did match the pin layout, so the AVR version can be interchanged with the 10-pin ARM version more easily.

I'm closing this PR, as it is superseded by the new one.

PR #1812, opened 2017-11-07
Add Microsemi PD70224

evanshultz:
Datasheet: https://www.microsemi.com/document-portal/doc_download/131677-pd70224-data-sheet

Footprint PR is at https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/74.

Image:
image


evanshultz:
The dot near pin 8 was placed slightly off, so I fixed it.

evanshultz:
I don't find 10mil pin name offset to be objectionable here. 20mil being the typical value is OK to me, but flagging other values seems odd. I could see not allowing more than 20mil, but I think <=20mil has purpose in the library and is already used.

evanshultz:
Can this be merged?

evanshultz:
Can this be merged?

SchrodingersGat:
Yep

evanshultz:
@SchrodingersGat Thanks! Would you take a look at my comment above? KLC update?

SchrodingersGat:
I believe that's appropriate, I will remove the minimum offset requirement. It will probably still be a warning

PR #1815, opened 2017-11-08
update LTC2309

lorem-ipsum:
stack GND and VDD pins allowing for same schematic for both housing versions.
Minor changes to adhere to KLC.


SchrodingersGat:
Nice work, thanks :)

PR #1817, opened 2017-11-09
Add Allegro current sensors

evanshultz:
Allegro parts:
-Wildcarded temperature character in MPNs (though this is sometimes tied to the part's ampacity)
-Used complete MPNs that include the package type, breaking up the symbols by package and adding footprint information
-Aligned pin names with datasheets
-Stacked pins where possible (and then reduced body size, if possible)
-Selected the new footprint repo for all custom packages
-Add all missing sensors from:
--http://www.allegromicro.com/en/Products/Current-Sensor-ICs/Zero-To-Fifty-Amp-Integrated-Conductor-Sensor-ICs.aspx
--http://www.allegromicro.com/en/Products/Current-Sensor-ICs/Fifty-To-Two-Hundred-Amp-Integrated-Conductor-Sensor-ICs.aspx
--http://www.allegromicro.com/en/Products/Current-Sensor-ICs/SIP-Package-Zero-To-Thousand-Amp-Sensor-ICs.aspx

AKM parts:
-Added all 4 symbols for AKM current sensors that match existing footprints

Footprint info:
-All existing Allegro and AKM footprints were merged at https://github.com/KiCad/kicad-footprints/pull/12.
-Missing Allegro footprints are at https://github.com/KiCad/kicad-footprints/pull/30

KLC violations:
-Many pins named VOUT aren't power output (perhaps the check scripts should only mention this for regulator libraries?)
-It may be best to rename the CQ-7* AKM footprints to fix FPFilter issues; suggestions?
-Some NC pins can be connected to GND so they're visible and passive (ACS722xMA, ACS780, etc.)

Sampling of new symbols:
image


SchrodingersGat:
Nice work, thanks Evan

PR #1818, opened 2017-11-10
Updated sym_lib_table file

SchrodingersGat:
This PR fixes the sym-lib-table and updates it to match the current state of the symbol libraries.

This is something that we now have to track, as the symbol library table feature has now been integrated into KiCad nightlies


evanshultz:
Assuming correctness...

PR #1819, opened 2017-11-10
Added Tja1052i Isolated CAN Transreciever

matyro:
Hi,
tja1052

I added TJA1052 isolated can transreciever.
You can find the datasheet under:
https://www.nxp.com/docs/en/data-sheet/TJA1052I.pdf

Cheers
Dominik


matyro:
Following errors of my symbol:

> Violating S4.3
> 
>     Rules for pin stacking
> 
>     NC Pin NC (4) @ (0,600)is stacked on other pins
> 
>     NC Pin NC (6) @ (0,600)is stacked on other pins
> 
>     Pinstack @ (0,600) must have exactly one (1) visible pin
> 
>      - Pin NC (4) is INVISIBLE
> 
>      - Pin NC (6) is INVISIBLE


The rest are errors in "old" symbols

SchrodingersGat:
Hi @matyro the NC pins **must not be stacked** - this will cause them to be connected together. 

If (and only if) the NC pins are **never** to be connected to anything (as per the datasheet) then they can be deleted from the symbol entirely.

matyro:
I made them visible and moved them to the µC site based on their physical location.

tja1052

SchrodingersGat:
1. RS pin - please remove this pin as there is no pin 18 on this package (or in the datasheet)
1. Ground pin stacking

Pin stacking requires some very specific rules to be followed. In particular:

- Only one pin in the stack can be visible
- All the invisible pins in the stack must be set to 'passive'

Reference: http://kicad-pcb.org/libraries/klc/S4.3/

matyro:
I moved some pins arround. 
RS should be STB. It is isolated from the left site of the circuit so i put it on the right side even though it is basicly an input.


tja1052

matyro:
Travis says that NC pins should be invisibe. When this is a fixed rule I will change it.

At least in my case I would still atach them to Gnd1 for additional mechanical strength.

SchrodingersGat:
In that case I think that the NC pins should be placed on the lower left, and the TXD / RXD should line up with CAN H / CAN L

matyro:
tja1052

SchrodingersGat:
Thanks @matyro for the changes!

PR #1823, opened 2017-11-12
MCU_Texas_MSP430: update metadata

aimylios:
I started to clean up the MCU_Texas_MSP430 library based on the new KLC3. This PR includes updates of the description, keywords and documentation file for all parts.
Next step would be to rework the LIB file, but that does not seem to be an easy task. A lot of violations to the KLC need to be resolved, and some of the required footprints (e.g., [TVSOP](http://www.ti.com/lit/an/scba009e/scba009e.pdf)) are not available in the .pretty repositories. Any hints how to do this efficiently are welcome. Otherwise I might just update the few parts I am personally interested in.

### DIFF
--- a/library/MCU_Texas_MSP430.dcm
+++ b/library/MCU_Texas_MSP430.dcm
@@ -1,2143 +1,2143 @@
 EESchema-DOCLIB  Version 2.0
 #
 $CMP MSP430AFE221IPW
-D 24pin TSSOP, 4KB Flash Memory, 256B RAM
+D 4kB Flash, 256B RAM, TSSOP-24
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@aimylios thanks for this contribution, a fine effort :)

I appreciate that looking at reworking an entire library can be a daunting task! We would accept any partial changes that you are willing to contribute.

I would suggest picking a particular violating and addressing it for all of the symbols which violate it, then moving to the next! Symbols that are missing footprints should have the appropriate footprints submitted to the relevant footprint repository!

Please let us know if you need any particular help or guidance, and thanks again for this great fix :)

PR #1828, opened 2017-11-14
Test PR for 7400 series symbols in IEC flavor

mrnuke:


SchrodingersGat:
A few thoughts:

1. Library naming convention

Please have a look at how Logic libs are currently named. Yours should be `Logic_74xx_IEC`

2. KLC

Your submissions are failing KLC - see output [here](https://travis-ci.org/KiCad/kicad-library/builds/301781433?utm_source=github_status&utm_medium=notification)

You can browse the KLC [here](http://kicad-pcb.org/libraries/klc/)

Also please note that these symbols should be background-filled as is the convention with existing library parts

3. Hidden Power Pins

You mentioned hidden power pins in your GitHub repo - we do not allow hidden power pins for symbols as part of the canonical library. (At least this is the case now, there are still some legacy symbols in the library that do have them).

mrnuke:
Hi Oliver,

I've been looking over the KLC, and it does not appear to support that these symbols should be background-filled. Specifically:

> Simple components (e.g. discrete components or those with a distinctive shape) should not be filled with background color

I think you're referring to

> Black-box symbols (e.g. ICs with hidden functionality) should be filled with background color

However, these symbols are _explicitly_ NOT blackbox symbols. I'll take a look at the other points you mentioned.

If you have a look at the actual symbols, you'll notice that there are no hidden pins.

jkriege2:
Please note that @bobc put in a lot of work to script these libs, see his PR for the updated symbols: https://github.com/KiCad/kicad-library/pull/1578 ... It would be great to simply use his scripts, so this work does not have to be done by hand and style-changes are easy to implement.

Just for future discussions, here are two screenshots of your proposed symbol-style (74xx573) and the existing symbols (74LS573):
2017-11-14 06_53_46-part library editor -- d__kicad_kicad_iec_symbols-master_74xx_iec lib
2017-11-14 06_54_08-part library editor -- d__kicad_kicad-library-review_library_logic_74xx lib

mrnuke:
Hi @SchrodingersGat. I;m trying to fix

>     - Field UNKNOWN at posx -100 posy 250 size 60
>     - Field UNKNOWN at posx -100 posy 250 size 60

But I don't see any such fields in the library editor. I'm not sure if it's somehow related to

>    Footprint field must be INVISIBLE
>    Datasheet field must be INVISIBLE

But again, no way to fix this in the library editor (Kicad 4.0.6). Please advise

evanshultz:
@mrnuke 
Let's look at `74xx164`. Near the top of the file there are 4 lines of the file look like this:
```
F0 "U" -150 400 60 H V C CNN
F1 "74xx164" 50 -1000 60 H V C CNN
F2 "" -100 250 60 H V C CNN
F3 "" -100 250 60 H V C CNN
```

The `60` is the text size and it should be `50` instead. The `V` is visible while `I` is invisible.

I prefer to edit text files for this sort of thing so I can make quick changes to the entire file, but this is also accessible in the GUI by:
image

There are a number of other errors and warnings, so keep asking questions! Thanks for being willing to contribute to KiCad!

mrnuke:
A few issues with the KLC itself:

> Symbol not centered on origin

It makes more sense to center the symbol on the interface between the control block and logic block. Otherwise the pins are pushed 1.27mm off-grid, and we need to make the symbol larger in order to be able to align such pins. I think this requirement will cause headaches for specific symbols.

> Pin ~ (7) @ (-100,-1000) length (50mils) is below 100mils

I had to make the power pins 1.27 mm long in order to align them to the grid. if I make them 3.81mm instead, they would look quite bad in relation to the other pins.

Not sure if there is a better solution than the one I provided. Given the 1.27mm text size for pin numbers, I also had to hide the pin numbers. I could make the pins longer, though I wouldn't be too happy with the appearance of the symbol were I to do so.

74xx164

evanshultz:
You're on the right track. 100mil pin length is best, and the symbol doesn't have to be centered. We will stick with KLC where we can, but this is a case where an exception is allowed. You can offset the symbol body a bit if everything else is better. Thanks for posting the image and describing the issue so that we can discuss with you before merging!

mrnuke:
Another interesting point is do we want pin numbers? I would say yes, and all schematics (produced by others) that I have show pin numbers on IEC symbols. In order to make them fit with the 1.27mm text size, I do have to make the pins larger.

Which one do you think?

Pin numbers with original length pins:
74xx164_crap_pin_numbers

Pin numbers with longer pins:
74xx164_pin_numbers

mrnuke:
For comparison with prior art, here's the symbol for an MC14503B in comparison to a resistor and opamp symbol from an old HP schematic.

hp_size_comparison

jkriege2:
1. shouldn't power-pins have a label so you know what "that pin at the top" does?
2. please have a look at the scripting-efforts from @bobc (i.e. his repo there: https://github.com/bobc/kicad-symgen and a tiny bit of explanation there: https://github.com/bobc/kicad-symgen/issues/1). I think he also started on IEC-symbols! Maybe you two can work together to get these libs scripted as are the other 74xx/40xx-libs.
3. The centering-rule for symbols is maybe (IMHO) the weakest rule we have in KLC ... Read it more like "center as best as it makes sense", see the explanation below the rule http://kicad-pcb.org/libraries/klc/S3.1/

Best,
JAN

mrnuke:
Hi @jkriege2.

1. Readability over ID10T proofing. The main point of IEC symbols is to make schematics more readable by describing the function of the component. In cases where a pin label interferes with the symbol, I will leave out the label, as is the case with the 74HC164.
2. Scripts are great for automating repetitive tasks -- e.g. connectors with varying numbers of pins --, but I don't see how this is easily apliable to IEC symbols, which are in many ways unique. I prefer to design a symbol visually before scripting it.
3. That's great! It makes sense in this case to center the component at the interface between the logic and control blocks.

jkriege2:
Hi @mrnuke 
1. the problem is that the symbol describes the function of all pins explicitly except these pins at the top/bottom. One solution would be to separate out the power-supply pins into a second unit in the symbol (as e.g. done for the new logic-gate symbols from @bobc ... and as will be done in future for e.g. OPAMPS etz.)
2. yes, but scripting will also make it way easier to manage the libs, once they are part of this repo. E.g. if we have changes in the KLC, it is very tedious to do all changes by hand (I did that e.g. for regul.lib or dc-dc.lib ... and that took me days of hand-work!). Also these symbols are "merely variants" of the differentl styled symbols already there ... so it would be great to manage them at one place, i.e. if someone adds a new logic device, he/she will only have to add it to the scripts and regenerate the libs ... and a new symbol with correct/mathing names, metadata, etz. will appear everywhere ... Maybe talk to @bobc on how he encodes IECs so we don't do double-work and can maybe profit from both your efforts and have good managability!

Best,
JAN

bobc:
Aesthetics is a tricky one, everyone seems to have their own ideas, unfortunately we end up with libraries that show the quirks of their author and look inconsistent with the other libraries. I think the KLC establishes a good compromise between consistency and freedom of expression.

I didn't find any guidance on how power pins should be handled on IEC symbols, I guess it is assumed they are a separate unit. To keep consistency I added power pins consistent with our other logic symbols, extending the height of the box if necessary. With the script I could easily generate a separate power unit instead.

Note that there are some components such as level converters that have two different VCC, in which case they really need to be labelled VCCA, VCCB or whatever.

As an example of a scripted symbols, the "script" looks like this :

```
#
# 4021
#
COMP 4021 U
FPLIST
DIP?16*
DESC 8-stage Shift Register
KEYW CMOS shft register
DOC http://www.intersil.com/content/dam/Intersil/documents/cd40/cd4014bms-21bms.pdf
#
UNIT WIDTH 500
ELEM CONTROL
GROUP "&ge;1" "C" ""
9 PL O L
SPC L
10 CP O L "&st;"
END-GROUP

ELEM
11 DS O L
7 P0 I L
ELEM
6 P1 I L
ELEM
5 P2 I L
ELEM
4 P3 I L
ELEM
13 P4 I L
ELEM
14 P5 I L
2 Q5 O R
ELEM
15 P6 I L
12 Q6 O R
ELEM
1 P7 I L
3 Q7 O R

SPC L

UNIT PWR
8 VSS PI B
16 VDD PI T
END
```

which generates 

image

The script has all the data to generate the symbols, "ELEM CONTROL" creates the control block. Pins can be grouped with GROUP. For special characters, an HTML style scheme is used. "&st;"  outputs the Schmitt trigger (hysteresis) symbol "&#x238E;", "\&ge;" is greater or equal sign, "&ge;". 

It is also easy to specify default box width, pin length but also override these for specific symbols.

The nice thing is that the symbol generator (symgen) ensures that most of the KLC rules are followed, and it is very easy to change. e.g. to make the power unit separate, just add the "SEPARATE" keyword i.e.
> UNIT PWR SEP

mrnuke:
Hi @jkriege2 

1. It's almost universally common in kicad libraries to have electrons coming in at the bottom and leaving at the top. In lack of further evidence, the conversation about labeling power pins is moot.
2. Saying how to do the work rather than what is micromanagement. I'm more than happy to work with @bobc to script the symbols once we reach some form of agreement on how they should look. AS far as changing/updating the symbols until such agreement, I will do so visually.

@bobc, do you have any symbols scripted with inverted inputs or other graphic features? I'm curious if those overlap with the pin numbers, as I've exemplified earlier.

bobc:
>  once we reach some form of agreement on how they should look

That could be difficult, if you are unwilling to discuss it. Generally the KiCad house style is the preferred style for submissions.  If you present a _fait accomplit_, I guess the library maintainers will take a view whether the look and style fits the KiCad house style or not.

There is no magic with regard to overlapping of pin numbers, you need to make the pin length long enough to accommodate the graphics. Normally 150-200 mil is fine.

jkriege2:
Hi @mrnuke !
1. May be, but it is also universally common to label (power-)pins ... so that argument does not really help. Also it is very common in other ECAD-systems (and a prefered way in the newer KiCAD lib symbols) to have power-pins as separate symbols. Currently KLC dictates power-pins on the symbols for single-device ICs, but I find that a bit inconsistent, so I proposed a change for that there: https://github.com/KiCad/kicad-library/issues/1839 ... let's see what the other's have to say on that, so we come up with better KLC-rules for these cases (and have them fixed!)
2. *Suggesting* how to do it in this case is trying to make the libs as manageable as possible (with the limited number of maintainers that we are) + proposing a way to have less work for you with more and better standardized output. You are free to do this work any way you like, but we were very happy to get the scripted symbols from @bobc and it was generally agreed that that is a good approach ... so use that info as you like. So see this as a hint to tools that are proven and available in the KiCAD ecosystem.

Best,
JAN

mrnuke:
There valid cases where it makes sense to use separate power pins, and valid cases where it makes sense to have integrated power pins. It's preferable to have both these options in the library, and leave it to the circuit designer to choose one or another, depending on what makes the schematic more readable. I focus on the latter because that's the harder problem to solve.

Re. pin labeling, IEC pins are either labeled by function, or not at all, yet they are not labeled by name. And since 7400 spans several technologies, It makes no sense to label VDD a pin that may just as well be VCC. And I'm not going to make IEC-look-alike-but-not-really-IEC symbols because I have to stretch it out of proportion to fit a GND/VEE label, like we see on the CD4021 example.

Now, to the stuff that actually matters.
* 100mil pins, no number, 200 mil pins, with number?
* 200mil pins all around, or only on symbols where 100 mil pins look cut-off with pin numbers?
* Stretch the symbol by 100 mil and use 150 mil pins on the sides?

jkriege2:
@mrnuke 

For the power pins: There are several devices that are e.g. available from different manufacturers, using different names for the pins in their datasheets ... Still we only have one symbol for all these variants in the lib and the idea is to find a name that everybody UNDERSTANDs. For the power-pins I think it's crucial to label them, as it is not at all clear from having a single pin, what it does, especially if there are (as suggested already) different types of power-pins e.g. for level-translators!
Also: For me the schematic gets less readable, when pins are not labeled + it makes it much harder to use these symbols, as I first have to look up (and remeber) what the unlabeled pin does!

About the sretching out: That's why I suggested the separate power-units...

About pin-length: Current KLC-rule on that: http://kicad-pcb.org/libraries/klc/S4.1/ ... Otherwise: As you wish: I WOULD make all pins on one side of the symbol have the same length, so all pins end/start at the same y-coordinate, but you can simply start from 100mil and increase in 50mil steps until text fits and you're satisfied ...

mrnuke:
Here are two possible KLC-compliant symbols incomparison to a datasheet IEC symbol (from an NXP datasheet).
Which one is the one we want?


iec-164-kicad

mrnuke:
As far as power pins go, it's obvious from a well designed schematic what the function is.
 
iec--kicad-sschem

If you're willing to accept compliant symbols with unlabeled integrated power pins, I'm willing to accept an alternate and compliant symbol with labeled unintegrated power pins. Everyone wins.

mrnuke:
To give you an idea, excessive labels on an IEC symbol clutter it, and make the schematic unsuitable for publication.

iec--kicad-sschem1

poeschlr:
I would include the pin numbers in the symbol. This helps when searching for searching a bug in the system. 
Then you can simply follow a signal with the schematics alone.
If you don't include pin names you need to also have the datasheet next to you and translate between datasheet and schematic before knowing which pin is responsible for the signal you are interested in.

jkriege2:
@poeschlr Did you mean to include pin **names** in your first line?

mrnuke:
ping

SchrodingersGat:
@mrnuke sorry but we have begun the transition process to a new consolidated symbols library at https://github.com/kicad/kicad-symbols 

Please re-open this PR at the new repository. Sorry for the inconvenience. There will still need to be some discussion but the new library will be the place to do it :)

PR #1829, opened 2017-11-14
Adjusted names of the adc-dac symbols to the new KLC

lavadew:


SchrodingersGat:
Fantastic work, thanks :)

PR #1832, opened 2017-11-15
Adjusted the PIC12 symbol names to the new KLC

lavadew:
Adjusted the PIC12 symbol names to the new KLC and created an alias for every LF variant.


lavadew:
saw some errors @ travis, will tackle them tomorrow.

SchrodingersGat:
Looks like you are missing some documentation for some of the part aliases

lavadew:
still not happy with the travis results, i'm on it.

SchrodingersGat:
Thanks @lavadew :)

Just a heads up - if you make further commits, we do not get notified unless you make a comment here. So please bump the thread when it's ready for review!

Great work!

PR #1834, opened 2017-11-16
Added TPS61090, TPS61091, and TPS61092 boost converters.

reportingsjr:
A couple of things:
I didn't stack the power pins since I really dislike pins not being shown on schematic symbols. Let me know if I should change this, but meh. Sounds like we will be getting the ability to have multiple numbers on one pin soon though!

I don't know why kicad decided to reorder half of the lines in lib file, but I can try to manually undo that for this PR.

Lastly, I just recompiled kicad and the library version was recently updated. Is that an issue since these library files are technically the equivalent of nightlies? :)

All three symbols I added are the same graphically (one is an adjustable output, one is 3v out, and one is 5v out) so here is a photo of what they all look like:
tps61090

Thanks!!!


reportingsjr:
Just an FYI, I see the KLC violations. I'll go ahead and fix them.

SchrodingersGat:
The line reordering comes about when people submit PRs from different KiCad versions. It could be that the most recent commit to this file was from stable.

If you fix the outstanding KLC violations I'm happy to merge :)

reportingsjr:
Ok, I went through and fixed the KLC errors. Let me know if I need to change anything else!

SchrodingersGat:
Thanks :)

Issue #1838, opened 2017-11-17
dc-dc.lib not working with kicad 4.0.7


aurel32:
Commit d89bebae7594f3ecf19027510e1e569798de691e made many format changes to the dc-dc library and this causes the description, the keywords and the documentation fields to be empty for all devices in this library. It seems that just downgrading the version in the header from 2.4 to 2.3 fixes the issues, but I don't know if there other side effects.

jkriege2:
I posted a fix: https://github.com/KiCad/kicad-library/pull/1840

aurel32:
Thanks, I confirm the fix works.

PR #1840, opened 2017-11-17
Fix https://github.com/KiCad/kicad-library/issues/1838

jkriege2:
- fixes https://github.com/KiCad/kicad-library/issues/1838
- removes one whiteline that lead to an error


evanshultz:
Is someone using a post-4.0.7 build of KiCad?

SchrodingersGat:
At least until v5 is out we need to ensure that the files submitted are compatible with stable builds. Could be added to the scripts.

PR #1845, opened 2017-11-19
fix two minor bugs in PIC12-lib

jkriege2:
- removed non-existent device from DCM (that device really is not produced!)
- fixed one description (package type)


evanshultz:
I see "ISM" instead of "xSM" which includes temperature, but that would need more fixes throughout. Thanks Jan!

jkriege2:
... yes and all package-variants are just ALIASes to the same symbol ... so that needs fixing too!

PR #1847, opened 2017-11-21
CD405

SchrodingersGat:
Add symbol for CD4051 8:1 analog mux/demux

image


jkriege2:
looks good ... but shouldn't that be part of 40xx-libs (also 4052 and 4053) ot 74xx-libs (e.g. for 74HCT4051 etal., see https://www.westfloridacomponents.com/mm5/graphics/F07/74HCT4051D.pdf)? I was thinking about that problem, when doing the rework of the switches ...

Best,
JAN

PR #1206, opened 2017-04-30
Add VNH2SP30

Marsfan:
Added a high current motor driver that I created for a personal project. 

3 Travis Failures: 
1. Footprint filter (not sure what this is)
2. Pin suggestion (CS means current sense in this case, not clock select, so it should be an output)
3. Unknown text field of bad size (Cannot fix this without finding it, and it eludes me)

PR #1849, opened 2017-11-21
Add VNH2SP30

Marsfan:
A while back I made this part, and just got around to cleaning up the schematic. But because of weirdess, that branch got royally screwed up. Here is a new PR without those issues.

FYI, the previous PR was #1206


Marsfan:
Missed the grid thing. Fixing it now

Marsfan:
Ok. Everything should be lined up now. The only failure left is that the systems is incorrectly assuming the use of a pin.

PR #1854, opened 2017-11-23
Add internal schematics to some ADI parts

evanshultz:
At https://github.com/KiCad/kicad-library/pull/1493, we discussed simplified internal schematics.
- I did this for the big DDS parts but forget to merge it.
- For the smaller parts with 3 numbers, the symbols are too small and I don't see why they should grow.
- For the isolation parts, I'd like to wait for dashed lines which I believe will come in KiCad v5

Datasheets for changed parts:
- http://www.analog.com/media/en/technical-documentation/data-sheets/AD9831.pdf
- http://www.analog.com/media/en/technical-documentation/data-sheets/AD9852.pdf
- http://www.analog.com/media/en/technical-documentation/data-sheets/AD9910.pdf

Image:
image

@jkriege2 You asked for this before, so you can review and merge?


jkriege2:
Just as a side-note @SchrodingersGat: The Travis output here is interesting ... it seems that the check-scripts do not work, when the dcm is missing?

Maybe you can have a look at the scripts, I will review this PR later ;-)

jkriege2:
looks good to me!

Thanks,
JAN

PR #1856, opened 2017-11-23
Hdc1080

DerekGn:
Added HDC1080 humidity sensor

image

Updated


SchrodingersGat:
Looks like you have found a weakness in the Travis scripts - where a new library file is added! I'll look into that. Thanks for the contribution

DerekGn:
Yes I noticed that and thought it was odd.

PR #1859, opened 2017-11-27
Add MAX3485 as an alias

antoniovazquezblanco:
Description: True RS-485/RS-422, 10Mbps, Slew-Rate Limited, with low-power shutdown, with receiver/driver enable, 32 receiver drive capacitity, DIP-8 and SOIC-8

Keywords: True RS-485/RS-422, 10Mbps, Slew-Rate Limited, with low-power shutdown, with receiver/driver enable, 32 receiver drive capacitity, DIP-8 and SOIC-8

Datasheet: https://datasheets.maximintegrated.com/en/ds/MAX3483-MAX3491.pdf


poeschlr:
The interface lib has already been transferred over to the new repo. Please re submit this over at the new repo https://github.com/KiCad/kicad-symbols

PR #1865, opened 2017-12-02
Adds Microchip HV513 (SOIC & QFN) symbols

RobertBaruch:
Note that the library editor moved some parts around in the lib and dcm files.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/20005846A.pdf

hv513wg

hv513k7


RobertBaruch:
After corrections:

hv513wg

hv513k7

PR #1868, opened 2017-12-05
fixed mislabled pin function

DerekGn:
Fixed mislabled pin function on SAML21E devices


poeschlr:
I am not convinced it is even a good idea to include all these alternative functions in the schematic symbol. This makes the symbols unnecessarily large and quite hard to read.

I would suggest to have the pin names short, with only enough information to identify the pin.

DerekGn:
Most of the parts in this library follow the same pattern if you check out ATXMEGA8E5-AU ping 20 label is PR0/XTAL2/TOSC2/CLKOUT/EVOUT/RTCOUT/AC1OUT

Shortening the labels would add to the current inconsitancies in this library. The SAML21 parts where added with fully qualified labels to be consistant to a degree with the other symbols.

Since they where added before the current KLC validation they require a rewrite, not sure i have time to do this now so I will cancel this PR.

poeschlr:
@DerekGn wrote:
> Most of the parts in this library follow the same pattern if you check out ATXMEGA8E5-AU ping 20 label is PR0/XTAL2/TOSC2/CLKOUT/EVOUT/RTCOUT/AC1OUT

Not really a good reason to keep something like it is.

> Shortening the labels would add to the current inconsitancies in this library. The SAML21 parts where added with fully qualified labels to be consistant to a degree with the other symbols.

Only if we stop after fixing a few symbols. Look at the NXP symbol libs on how it might look like. (I think there the shortening of the pin names is done. If there are still some pins with very long pin names then these have been overlooked.)

DerekGn:
> Not really a good reason to keep something like it is.

It was reasonable at the time the parts where added to the library. As I said I dont have time at the moment to redo the parts

PR #1869, opened 2017-12-06
add 4P4C connector (aka RJ9 or RJ10 or RJ22)

lorem-ipsum:
Add the (not really standardised) 4P4C connector. It is often called RJ9, RJ10 or RJ22, but neither is correct as it not technically a registered jack.
The drawing is slightly asymmetric and I find it ugly, but there is no way to fix it as the drawing has to be on the 50 mils grid (or I am too stupid to find the override).
There is no specification or data sheet for this connector I could find.

4p4c


lorem-ipsum:
I've just seen the comments regarding the library move in the previous PR. I will create a new PR on the new repo.

PR #1875, opened 2017-12-18
Added fourth pin to MCP1825S

johnfwhitmore:
I originally contributed the MCP1825S Component by copying the MCP1826S regulator. There is however a fourth pin which I've added. If this is Acceptable I can make the change to MCP1826S.

screenshot from 2017-12-18 20-59-07


SchrodingersGat:
Thanks :)

SchrodingersGat:
Thanks :)

johnfwhitmore:
Deadly! That was quick, thanks for that. I'll do the same to the MCP1826S

jkriege2:
Hmmm ... When reworking the regul.lib, I cleaned it up a bit: The MCP1825S explicitly has the footprint filter `SOT?223*TabPin2*`, which points to a SOT-223-footprint, where the tab (you pin ) is actually pin2, so this "fix" is unnecessary and breaks the function(!!!), as the footprint only has pins 1,2,3. Also with the third pin, this symbol does no longer fit the other footprints in the filter-list (TO-220 only has 3 pins!!!). So please revert this. Or split out the 3 package variants into single symbols and set the explicit and correct footprint for each!

Also: The two GND-pins should be stacked for a nicer style!

Best,
JAN

SchrodingersGat:
Ah Jan you are correct, good catch. I did not spot that you had set the footprint to one that already took this "missing" pin into account. I'll revert this change.

johnfwhitmore:
Oops, sorry I'll add the various parts :-(

PR #1883, opened 2017-12-29
Added bluepill module

kajusK:
Added a cheap stm32 development board - http://wiki.stm32duino.com/index.php?title=Blue_Pill


SchrodingersGat:
Hi @kajusK we are transitioning to the new library repository in preparation for the v5 KiCad release. Please resubmit this PR to https://github.com/kicad/kicad-symbols 👍

kajusK:
Hi,
could you give me a hint to which file I should add the module in the new repository? I can't see modules there and quick search for Arduino and similar from the old repo found nothing.

poeschlr:
As the modules lib is not yet ticked in the todo file, i would guess that the transition of this lib has not yet been finished.

It does look like this process has not even started yet.
This means make a new lib called modules to the new repo.

PR #1885, opened 2018-01-03
Add Microchip MCS3142 Keeloq enocder

swbr:
The Microchip MCS3142 is used in many car key fobs to detect pressed buttons, encode an encrypted data package, and send it via e.g. 433MHz. Two crystals, the buttons and a battery are the only necessary external components for this TSSOP20 IC.
See http://ww1.microchip.com/downloads/en/DeviceDoc/40001747A.pdf

The device has pins DATA_IN/OUT and CONTROL_IN/OUT, which are not described in the datasheet, except that the OUTs should be connected to the INs. 
So, I have grouped those four pins together on the left for a more clear schematic.

mcs3142


SchrodingersGat:
Hi @swbr we are transitioning to the new library repository in preparation for the v5 KiCad release. Please resubmit this PR to https://github.com/kicad/kicad-symbols 👍

